PR #60, opened 2015-01-21
nxdomain on mesos support

eyberg:


resolver/resolver.go
```diff
@@ -126,6 +126,27 @@ func (res *Resolver) formatA(dom string, target string) (*dns.A, error) {
...
+			Rrtype: dns.TypeSOA,
+			Class:  dns.ClassINET,
+			Ttl:    ttl,
+		},
+		Ns:      res.Config.Mname,
+		Mbox:    res.Config.Email,
+		Serial:  uint32(time.Now().Unix()),
+		Refresh: ttl,
+		Retry:   600,
+		Expire:  86400,
```
tarnfeld: It might be nice to allow the user to configure these sort of parameters.

PR #71, opened 2015-01-24
Frameworkhostfix

eyberg:


kozyraki:
@eyberg  Does this branch pass go test ./... for you? 
I am having some problems with it. Can you check?

PR #77, opened 2015-02-10
version

eyberg:


lloesche:
- Can we put the version into a standalone file like version.go?
- The commit message is a dot.

PR #80, opened 2015-02-19
add 'Testing Mesos-DNS' section

air:
So users don't have to go digging for the `dig` format, like I had to today in a demo.

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -54,6 +54,53 @@ Note that the `hostname` field refers to the hostname used by the slave when it 
 curl http://master_hostname:5050/master/state.json | python -mjson.tool
 ```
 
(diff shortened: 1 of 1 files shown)


kozyraki:
@air (cool username btw)

You can get dig and naming instructions from http://mesosphere.github.io/mesos-dns/docs/naming.html 

In general, it's better if testing instructions are not inline with setup.Ie, explain all the setup (including slave setup) and then explain some basic testing. Ie, move this to the end of the file. If testing is linked to a specific example, instructions should to into doc that explains the example (e.g., http://mesosphere.github.io/mesos-dns/docs/tutorial-gce.html)

tsenart:
Duplicate of #168

PR #81, opened 2015-02-19
multiple port support + A/SRV entries for master

eyberg:


records/generator.go
```diff
@@ -134,12 +134,21 @@ func (rg *RecordGenerator) loadWrap(ip string, port string) (StateJSON, error) {
...
+// yankPorts takes an array of port ranges
+// FIXME - does not use ranges
+func yankPorts(ports string) []string {
 	rhs := strings.Split(ports, "[")[1]
 	lhs := strings.Split(rhs, "]")[0]
-	return strings.Split(lhs, "-")[0]
+
+	yports := []string{}
+
+	mports := strings.Split(lhs, ", ")
```
tarnfeld: Would it not be better to split on `,` and then lstrip/rstrip any extra whitespace?

kozyraki:
@eyberg Ian my apologies for this. By mistake I worked on this branch when implementing another feature (insert SRV and A records for masters). I hope you don't mind if we keep them together.

PR #85, opened 2015-02-24
Make names and queries case-insensitive

### LINKED ISSUE #84: Query handling should be case-insensitive
The DNS protocol is supposed to be case-insensitive, but it appears that mesos-dns is not.  e.g. a query for mesos-dns.mesos. will work with the included sample config, but Mesos-Dns.mesos. will not. Makes it extra fidgety when frameworks register with mixed-case names...

benley:
This could be done more elegantly, but at least it seems to work for now.  Still needs tests.

Fixes #84

[Review on Reviewable](https://reviewable.io/reviews/mesosphere/mesos-dns/85)

### DIFF
--- a/factories/fake.json
+++ b/factories/fake.json
@@ -92,7 +92,7 @@
             "failover_timeout": 1200,
             "hostname": "localhost",
             "id": "20140703-014514-3041283216-5050-5348-0001",
-            "name": "chronos with a space -2.0.1",
+            "name": "chronos with a space AND MIXED CASE-2.0.1",
             "offers": [],
             "registered_time": 1414913537.10742,
             "reregistered_time": 1414913537.10744,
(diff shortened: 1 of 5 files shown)


kozyraki:
@benley Great catch Ben. I am surprised none of us noticed this before. Let me test your branch a little first and then will move forward. Thanks

benley:
Thanks. I'll try to add some unit tests this afternoon if you want to use this patch.

kozyraki:
@benley Great, ping me when the unit tests are in and I will do my side quickly after that. Thanks a lot.

benley:
OK, added some test coverage. Just fyi, this is literally the first time I've edited Go code so look out for noob mistakes :)

kozyraki:
@benley Thanks. I will test it now.

kozyraki:
@benley made a new branch with a few corrections. Will make a pull request from that. 
Thanks

PR #89, opened 2015-02-27
added mesos-dns nameserver feature

geraldstanje:
added mesos-dns nameserver feature. 
the implemented test does not succeed and needs further debugging. 
m.Answer in resolver.go is empty.


geraldstanje:
please review with shepherd @BenWhitehead

geraldstanje:
@BenWhitehead a quick reminder to add a test-case for the bug we found: r.Question[0].Name

PR #92, opened 2015-03-03
Support multiple listeners

### LINKED ISSUE #87: Be able to bind mesos-dns to multiple interfaces
Right now the "listener" option only takes one IP. It would be good if I could bind mesos-dns to several IPs so that it is available from e.g. the office via the servers public interface as well as on the internal network via it's internal interface. But I don't want it to bind to the VM network or the Docker interface so I can't bind it to 0.0.0.0.

discordianfish:
This changes the listener parameter to listeners list parameter.
mesos-dns will try to bind each address and also use them for the
generated records like the SOA and mesos-dns records.

This closes #87.

### DIFF
--- a/config.json.sample
+++ b/config.json.sample
@@ -6,6 +6,6 @@
   "port": 8053,
   "resolvers": ["8.8.8.8"],
   "timeout": 5,
-  "listener": "0.0.0.0",
+  "listeners": [ "0.0.0.0" ],
   "email": "root.mesos-dns.mesos"
 }
--- a/docs/docs/configuration-parameters.md
+++ b/docs/docs/configuration-parameters.md
@@ -17,7 +17,7 @@ The configuration file should include the following fields:
   "port": 53,
   "resolvers": ["169.254.169.254"],
   "timeout": 5, 
-  "listener": "10.101.160.16",
+  "listeners": [ "10.101.160.16" ],
   "email": "root.mesos-dns.mesos"
 }
 ```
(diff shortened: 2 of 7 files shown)


kozyraki:
@discordianfish Thanks a lot, this is very useful!
I may not get to test it this week but I will try to get to it as soon as possible. I think we need a couple more tests in resolver_test.go to be on the safe side too.

discordianfish:
@kozyraki The problem is, testing the multiple listeners requires the test to bind additional IPs to some interface for which there is no platform independent way.

BenWhitehead:
In your tests you might be able to bind to multiple local interfaces `127.0.0.1,127.0.1.1,127.0.2.1` for example and then verify each address responds to requests.

discordianfish:
@BenWhitehead No, you need to configure the IPs to the interface first, at least on linux and osx. On linux you can also set 'net.ipv4.ip_nonlocal_bind=1', but anyway it's a platform specific thing.
So for proper tests, this means either to limit those tests to run only on a some platforms or implement platform specific methods to configure the IPs on all supported platform.

BenWhitehead:
fair enough

tsenart:
@discordianfish: I'm really sorry this got lost on the review queue. We're stepping up our game here! Would you be so kind  as to re-apply your changes to the current master? Thanks.

discordianfish:
Given how simple the change but outdate it is, it probably makes more sense to redo it instead of fixing the merge conflicts.
@lloesche Is this still something we want? If so, I'll do a new PR.

PR #99, opened 2015-03-09
exit if panic or error in listener goroutine

eyberg:


kozyraki:
@eyberg Ian can you verify that this exits in both cases: if the listener has an error or if it returns (somehow) without an error? I want to protect against both cases...

Issue #107, opened 2015-03-20
Support multiple ports in SVR records


richtera:
Right now an instance will just have _tcp and _udp records for all their ports.
There is no provision for _tcp0 _tcp1 and such (maybe do it here) or add .0 .1 to the end of the service name. The only way I found was to guess that the lower port is the first port.

richtera:
Maybe here

```go
                                       for s := 0; s < len(sports); s++ {
                                                var srvhost string = tname + "." + fname + "." + domain + ":" + sports[s]

                                                // either change this to output
                                                // tname + "_" + index
                                                // or register additional SRV records so it doesn't break anything else

                                                tcp := "_" + tname + "._tcp." + tail
                                                udp := "_" + tname + "._udp." + tail

                                                rg.insertRR(tcp, srvhost, "SRV")
                                                rg.insertRR(udp, srvhost, "SRV")
                                        }

```

kozyraki:
@richtera with the 0.22 mesos release, you will be able to name ports. We will update Mesos-DNS to use these names to generate proper SRV records. Can you wait till then? It should be very soon.

richtera:
I’ll use my branch till then, but it’s only a proof of concept. What’s “very soon” :)?
Thanks
Andy

richtera:
Question though. This mesos-dns is a good start, but there is no easy way to map those ports during startup of another instance. I created a lookup module in python where you can do
```javascript
{
  ...
  "cmd": "<entrypoint> `python -m lookup ../dependency --index 0`"
  ...
}
```
And it would convert the command to
```javascript
<entrypoint> <machine>:<port>
or
<entrypoint> <machine1>:<port1>,<machine2>:<port2>
```
But it all seems a little cumbersome. Will there be better mapping of dependent ports in the next mesos version?

kozyraki:
@richtera Hi Andreas. I will share the plans for Mesos-DNS 0.2 soon but here is a preview of the two features that you may care about:
- the ability to flexibly parse the service discovery info for tasks coming with Mesos 0.22. This will allow you to pass names for ports and all sorts of labels for your tasks
- a flexible naming scheme for tasks (e.g. the ability to generate names like mytask.prod.eastcoast.mesos)
- an http interface for SRV/port information (since few people use SRV requests as is). This will get you more or less what you want in the last message (e.g., if you query _http._tcp.mytask.something you will get the info for the http port of your task -- assuming you named it that way). 

[...]

PR #108, opened 2015-03-20
Fix multiple ports to have _<portIndex> in the name. Proposal for #107

richtera:
Add _<portindex> to the "name" part of the instance name.


tsenart:
@richtera: Apologies for the (very) late reply. This change set seems now out-dated. If you fell this is still relevant would you please provide the rationale and motivation behind the proposal? Thanks!

richtera:
It’s ok, my project is currently suspended, but I am going to hopefully pick it up again and then I’ll see how things will get put together.
Thanks
Andy

tsenart:
@richtera: Don't hesitate to start discussion again when you're ready. :-)

PR #109, opened 2015-03-21
ZKdetect interface cleanup; address concurrency issues

jdef:
Fixed some problems that cropped up in `make testrace` with `GOMAXPROCS=24`. Haven't done much run-time testing of these changes.


kozyraki:
Thanks a lot. I ll test and merge tomorrow.

PR #125, opened 2015-04-12
Panic handler

jdef:
- centralize panic handling for the resolver, move calls to `os.Exit` into main
- fix problems reported by `go vet`
- ran `go fmt`
- refactored config reload: can also be triggered now by detected mesos leadership changes

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,7 +60,6 @@ func main() {
 		newLeader, zkErr = resolver.LaunchZK(zkInitialDetectionTimeout)
 	}
 
-	defer util.HandleCrash()
 	handleServerErr := func(name string, err error) {
 		if err != nil {
 			logging.Error.Fatalf("%s failed: %v", name, err)
--- a/main.go
+++ b/main.go
@@ -38,46 +38,46 @@ func main() {
 	config := records.SetConfig(*cjson)
 	resolver := resolver.New(version, config)
 
+	var dnsErr, httpErr, zkErr <-chan error
 	// launch DNS server
 	if config.DnsOn {
-		go func() {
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -33,17 +40,31 @@ func main() {
...
+			}
+		}()
 	}
 
 	// launch HTTP server
 	if config.HttpOn {
-		go resolver.LaunchHTTP()
+		go func() {
+			if err := resolver.LaunchHTTP(); err != nil {
+				logging.Error.Fatalf("HTTP server failed: %v", err)
```
kozyraki: Wouldn't it be cleaner if the for launch{DNS, HTTP, ZK} happened within these functions?

jdef: Thanks for the suggestion. I cleaned it up

resolver/resolver.go
```diff
@@ -85,30 +81,34 @@ func (res *Resolver) Serve(net string) {
...
-func (res *Resolver) LaunchZK() {
-	dr, err := res.ZKdetect()
-	if err != nil {
-		logging.Error.Println(err.Error())
-		os.Exit(1)
-	}
+// launches Zookeeper detector, returns immediately
+func (res *Resolver) LaunchZK() <-chan error {
+	errCh := make(chan error, 1)
+	go func() {
```
kozyraki: @jdef There is one issue now that we lunch ZKdetect as a goroutine. The first reload() of state happens before ZKdetect gets the initial master and one has to wait for a reload cycle for Mesos-DNS to work. This is annoying for testing (not a major issue after that though). 

Is this go routine really needed?

PR #127, opened 2015-04-16
sanitized task names should make for good SRV service labels

jdef:
change generation of SRV records like this:
```
 _apiserver.core.km._tcp.marathon.mesos.: apiserver.core.km-s0.marathon.mesos.:31000
```

to
```
 _apiserver-core-km._tcp.marathon.mesos.: apiserver.core.km-s0.marathon.mesos.:31000
```


jdef:
This impacts A record generation too:
- `apiserver.core.km-s0.marathon.mesos` --> `apiserver-core-km-s0.marathon.mesos`

Issue #114, opened 2015-03-26
How can I expose a Marathon endpoint for my Marathon cluster


doctapp:
Would be really useful to know how to have a known Marathon endpoint for a set of Marathon masters. This way, I could have a single highly-available endpoint for launching Marathon apps.
Thanks

kozyraki:
@doctapp let me investigate this a little and get back to you. 
There are a number of issues here:
- how was Marathon launched. If it was launched through Mesos (runs on slaves etc) then Mesos-DNS knows about it. Otherwise, Mesos-DNS is clueless. 
- what is the endpoint you care about and how do you want it advertised. We can stick anything in a TXT record and we are also adding an http interface. But if you want a redirection, it has implications. 
Share your thoughts on these issues.

doctapp:
Thanks for the info. Marathon is started as a local service on the mesos masters. Since it's a framework, I don't understand how mesos it can be launched otherwise. Does moving this process to slaves the recommended way?

The endpoint I care about is marathon's REST API since I want to have a highly available way of starting and stopping apps. I would like this advertised as a single endpoint backed by multiple marathon masters and I don't want my apps to know about all marathon masters for this purpose.

Does this info help you understand my need better?

kensipe:
@doctapp to clarify... marathon running on the slaves isn't sufficient for it to register with mesos-dns.   What was meant, is that marathon is launched as a task by mesos on a slave.   Although it may seem odd... it would be possible for instance to launch marathon (m2) with a "master" marathon (m1).     The m2 marathon would be registered in mesos-dns because it is launched by mesos onto a slave.   

@kozyraki it would be nice to have frameworks A records regardless...   This is a good example.   It would be great if `marathon.mesos.` returned the IPs of marathons regardless of how they were run...

doctapp:
@kensipe thanks for the clarification, that's a good hack. Would definitely be great if there was a reliable way of getting the marathon endpoints.

tsenart:
Fixed by https://github.com/mesosphere/mesos-dns/pull/209

PR #133, opened 2015-04-20
add dns records for framework webui_urls

stuart-warren:
I'd like to discover where my frameworks are running

In mesos 0.21.0 mesos allows a ``webui_url`` parameter for frameworks to
advertise themselves to mesos users.
This change adds extra dns records (A + SRV) for each framework entry in ``/master/state.json``, pulling the host/port from the ``webui_url`` field:

e.g.
```
{
"name":"marathon",
"tasks":[],
"webui_url":"http:\/\/192.168.187.10:8080"
}

[A]	webui.marathon.mesos.: 192.168.187.10
[SRV]	_webui._tcp.marathon.mesos.: webui.marathon.mesos.:8080
[SRV]	_webui._udp.marathon.mesos.: webui.marathon.mesos.:8080
```

http://localhost:8123/v1/services/_webui._tcp.marathon.mesos.

May be related to #114


records/generator.go
```diff
@@ -224,6 +226,25 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
...
+			if fhost != "" && fport != "" {
+				// if fhost is an ip add an A record
+				if net.ParseIP(fhost) != nil {
+					arec := "webui" + "." + tail
+					rg.insertRR(arec, fhost, "A")
+					fhost = arec
+				} else {
+					fhost += "."
+				}
+				tcp := "_webui" + "._tcp." + tail
```
jdef: why _webui as the service? since the web ui URL can actually specify a protocol, why not attempt to pull the protocol from the parsed URL? if none, then does it make sense to default to _http?

kozyraki:
I agree with @jdef, it seems more reasonable to create _http SRV records (which means parse the protocol and use that -- or http by default). 

An alternative is to assume that these are always HTTP protocols and go with records named _marathon._tcp.mesos. 

Great idea overall.

jdef:
Advise caution using _framework as the service name since some framework
names might look like "framework.zone1.east". Also, I don't know how
comfortable I am with always assuming HTTP. For example some frameworks
might expose an https endpoint.

PR #140, opened 2015-04-26
startup fails if only zk and not masters is specified in config

jdef:


kozyraki:
Replaced by somewhat simplified 2fix branch.

PR #141, opened 2015-04-27
Eliminate potential race when zk not field not used.

kozyraki:


kozyraki:
@jdef this is your jdef:fix-zk-only branch with some code eliminated and a couple of minor fixes. 
I've tested it with and without the ZK field. If you don't see any issues, we can merge it.

jdef:
we can close this, yes?

PR #146, opened 2015-05-02
Fix 144

jdef:
change summary:
- make unit tests a little more reliable
- add some unit testing for zk detection
- rebase to latest samual/go-zookeeper
- rebase to latest miekg/dns
- rebase to latest mesos/mesos-go
  - resolves #144 (needs more real-world integration testing)
- reduce initial zk leader detection timeout

Issue #144, opened 2015-04-30
Not detecting mesos creating entries in zookeeper


cmaloney:
On multi-master clusters I launch, fairly frequently mesos-dns connects to the zk before mesos does, fails to watch because `zk: node does not exist`, then never sees when finally mesos attaches and does a election, making the changes not show up until mesos-dns is restarted because it didn't detect a leader.

mesos-dns log:
```
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal systemd[1]: Starting Mesos DNS...
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal systemd[1]: Started Mesos DNS.
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal mesos-dns[1629]: VERY VERBOSE: 2015/04/30 04:51:54 config.go:148: Mesos-DNS configuration:
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal mesos-dns[1629]: VERY VERBOSE: 2015/04/30 04:51:54 config.go:153:    - Zookeeper:  zk://127.0.0.1:2181/mesos
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal mesos-dns[1629]: VERY VERBOSE: 2015/04/30 04:51:54 config.go:155:    - RefreshSeconds:  30
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal mesos-dns[1629]: VERY VERBOSE: 2015/04/30 04:51:54 config.go:156:    - Domain: mesos
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal mesos-dns[1629]: VERY VERBOSE: 2015/04/30 04:51:54 config.go:157:    - Listener: 0.0.0.0
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal mesos-dns[1629]: VERY VERBOSE: 2015/04/30 04:51:54 config.go:158:    - Port:  53
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal mesos-dns[1629]: VERY VERBOSE: 2015/04/30 04:51:54 config.go:159:    - DnsOn:  true
Apr 30 04:51:54 ip-10-0-0-13.us-west-2.compute.internal mesos-dns[1629]: VERY VERBOSE: 2015/04/30 04:51:54 config.go:160:    - TTL:  60
[...]

PR #147, opened 2015-05-04
WIP: Add DiscoveryInfo and custom domain name support

sttts:
- add domain name `Templates` configuration flag
- create non-canonical records according the domain patterns, interpolated as text/templates
- implement container IP proposal of https://github.com/mesosphere/mesos-dns/pull/197#issuecomment-128594330 (unfinished, see TODO below)

Reasons to implement our own template engine:
- text/template has no error handling for missing values of variables (until 5 months ago, but not in Go 1.3 or 1.4 yet)
- the template syntax is an interface to the user. If text/template causes performance problems (it seems to be pretty central in the record generation and could be very well be the bottleneck), we would not be able to optimize without the burden to fork the complete text/template library. With our own implementation we have full control.
- text/template would not allow easy checking for validity of a template
- text/template implements much more logic than we need, it uses reflection heavily. Overall it feels to be pretty heavy for our small domain name templates.

TODO:
- [x] re-add tests
- [x] drop records with missing context variable values
- [x] fallback from di-name -> name
- [x] document patterns
- [ ] add default values for variables in patterns (maybe follow-up PR)
- [ ] do benchmarks
- [ ] add (d) and (e) of container IP proposal at https://github.com/mesosphere/mesos-dns/pull/197#issuecomment-128594330 (maybe follow-up PR)
- [ ] add support for DiscoveryInfo *visibility* (maybe follow-up PR)
- [ ] use `domainFrag` and friends
- [ ] handle spec returning empty values
- [ ] use better hash function, e.g. `zbase32` (maybe follow-up PR)

### DIFF
--- a/docs/docs/configuration-parameters.md
+++ b/docs/docs/configuration-parameters.md
@@ -29,7 +29,8 @@ The configuration file should include the following fields:
   "SOARetry":   600,
   "SOAExpire":  86400,
   "SOAMinttl": 60,
-  "IPSources": ["mesos", "host"]
+  "IPSources": ["mesos", "host"],
+  "templates": ["{name}.{framework}"]
 }
 ```
 
@@ -74,3 +75,5 @@ It is sufficient to specify just one of the `zk` or `masters` field. If both are
 `recurseon` controls if the DNS replies for names in the Mesos domain will indicate that recursion is available. The default value is `true`. 
 
 `enforceRFC952` will enforce an older, more strict set of rules for DNS labels. For details, see the [RFC-952](https://tools.ietf.org/html/rfc952). The default value is `false`.
+
+`templates` is a list of domain name templates used to create records. The default value is `["{name}.{framework}"]`, compare the [naming documentation](naming.md).
--- a/docs/docs/naming.md
+++ b/docs/docs/naming.md
@@ -64,7 +64,7 @@ For example, a query of the A records for `search.marathon.slave.mesos` would yi
 - `MesosContainerizer.NetworkSettings.IPAddress`.
 
 In general support for these will not be available before Mesos 0.24.
- 
+
 ## SRV Records
 
 An SRV record associates a service name to a hostname and an IP port.
@@ -85,7 +85,7 @@ $ dig _search._tcp.marathon.mesos SRV
 
 ;; ANSWER SECTION:
 _search._tcp.marathon.mesos.	60 IN SRV 0 0 31302 10.254.132.41.
-``` 
+```
 
 Mesos-DNS supports the use of a task's DiscoveryInfo for SRV record generation.
 If no DiscoveryInfo is available then Mesos-DNS will fall back to those "ports" resources allocated for the task.
@@ -113,6 +113,53 @@ Also note that there is inherent delay between the election of a new master and 
 Mesos-DNS generates A records for itself that list all the IP addresses that Mesos-DNS is listening to. The name for Mesos-DNS can be selected using the `SOARname` [configuration parameter](configuration-parameters.html). The default name is `ns1.mesos`. 
 
 In addition to A and SRV records for Mesos tasks, Mesos-DNS supports requests for SOA and NS records for the Mesos domain. DNS requests for records of other types in the Mesos domain will return `NXDOMAIN`. Mesos-DNS does not support PTR records needed fo reserve lookups. 
+
+## Customization of Records
+
+Mesos-DNS allows customization of the list of *record templates* that are used 
+to generate the A record names for each task. Compare the `Template` 
+configuration setting in the [configuration parameter documentation](configuration-parameters.md).
+
+By default only one *record template* is used:
+
+```
+{name}.{framework}
+```
+
+This gives the A records described above, e.g. `nginx.marathon.mesos.`.
+
+Additional templates can be defined using a number of *template variables*. These exist for every task:
+
+- `framework`: the framework name, e.g. `marathon`
+- `slave-id-short`: the number in the slave ID after the last dash `-`, e.g. `0` for a slave ID `20140803-125133-3041283216-5050-2410-0`
+- `slave-id`: the complete slave ID, e.g. `20140803-125133-3041283216-5050-2410-0`
+- `task-id`: the complate task ID, e.g. `chronos.49b91a9a-3dda-11e4-a088-c20493233aa5`
+- `task-id-hash`: a hash value of the task ID, e.g. 76830, sufficiently unique for a realistic number of tasks on a cluster
+- `name`: the name of a task, provided by the framework in the *DiscoveryInfo*, falling back to the name in the *TaskInfo*, e.g. `nginx`
+
+In addition there are a number of optional *template variables* which may be provided by the framework to Mesos in the *DiscoveryInfo* field of a task:
+
+- `version`: the version of a task, provided by the framework in the *DiscoveryInfo*, e.g. `1.0`
+- `location`: the location of a task, provider by the framework in the *DiscoveryInfo*, e.g. `europe`
+- `environment`: the environment of a task, provider by the framework in the *DiscoveryInfo*, e.g. `prod`
+- `label:<key>`: the value for the given label key, provided by the framework in the *DiscoveryInfo*, e.g. label key `canary` with label value `lanzarote`
+
+A *record template* is a string with the condition that after substitution of all *variable references* with a constant string `a` the resulting string is a valid domain name.
+
+A *variable references* is a string matching the regular expression `{\s*<template variable>\s*}`
+where `<template variable>` is any of the *template variable* identifiers from 
+above.
+
+Mesos-DNS will interpolate the given list of *record templates* for each running
+task in the cluster. If a *referenced variable is not defined* or the value is
+the empty string, the record *will not be created*.
+
+Examples:
+- `{name}.{framework}` gives e.g. `nginx.marathon.mesos`.
+- `{name}.{location}.{framework}` gives e.g. `nginx.europe.marathon.mesos`, or no record at all for those tasks that don't define the location.
+- `{version}.{name}.{framework}` gives e.g. `1.0.nginx.marathon.mesos`, or no record at all for those tasks that don't define the version.
+- `{label:canary}.{name}` gives e.g. `lanzarote.nginx.mesos`.
+- `{label:unknown}.{name}` will give no record at all, if no label `unknown` is provided by the framework.
 
 ## Notes
 
--- a/records/generator.go
+++ b/records/generator.go
@@ -16,6 +16,7 @@ import (
 	"github.com/mesosphere/mesos-dns/logging"
 	"github.com/mesosphere/mesos-dns/records/labels"
 	"github.com/mesosphere/mesos-dns/records/state"
+	"github.com/mesosphere/mesos-dns/records/tmpl"
 )
 
 // Map host/service name to DNS answer
@@ -52,7 +53,7 @@ func (rg *RecordGenerator) ParseState(leader string, c Config) error {
 		hostSpec = labels.RFC952
 	}
 
-	return rg.InsertState(sj, c.Domain, c.SOARname, c.Listener, c.Masters, c.IPSources, hostSpec)
+	return rg.InsertState(sj, c.Domain, c.SOARname, c.Listener, c.Masters, c.IPSources, c.Templates, hostSpec)
 }
 
 // Tries each master and looks for the leader
@@ -156,17 +157,6 @@ func (rg *RecordGenerator) loadWrap(ip string, port string) (state.State, error)
 	return sj, err
 }
 
-// BUG: The probability of hashing collisions is too high with only 17 bits.
-// NOTE: Using a numerical base as high as valid characters in DNS names would
-// reduce the resulting length without risking more collisions.
-func hashString(s string) string {
-	h := fnv.New32a()
-	_, _ = h.Write([]byte(s))
-	sum := h.Sum32()
-	lower, upper := uint16(sum), uint16(sum>>16)
-	return strconv.FormatUint(uint64(lower+upper), 10)
-}
-
 // attempt to translate the hostname into an IPv4 address. logs an error if IP
 // lookup fails. if an IP address cannot be found, returns the same hostname
 // that was given. upon success returns the IP address as a string.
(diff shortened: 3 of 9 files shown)


records/generator.go
```diff
@@ -21,12 +23,15 @@ import (
...
 
+var defaultDomainPatterns = []string{"{{.tname}}"}
+
 // Mesos-DNS state
 // Refactor when discovery id is available
 type RecordGenerator struct {
 	As     rrs
 	SRVs   rrs
 	Slaves map[string]string
+	DomainPatterns []string
```
jdef: the current design passes Config items to each func as needed, why is this special? If RecordGenerator is going to house Config items, then perhaps it should declare Config as an anonymous field:
```go
type RecordGenerator struct {
  Config
  As rrs
  SRVs rrs
  //...
}
```

this allows:

1. simple Config assignment: `rg.Config = someOtherConfig`
1. simple access to embedded Config field's from RecordGenerator obj: `for _, pattern in rg.DomainPatterns {`
1. a more consistent initialization of default config values (`defaultDomainPatterns` probably belongs in Config with some of the other defaults)

records/generator.go
```diff
@@ -232,24 +258,40 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
 				continue
 			}
 
-			tname := labels.AsDNS952(task.Name)
-			sid := slaveIdTail(task.SlaveId)
-			tag := hashString(task.Id)
-			tail := fname + "." + domain + "."
+			fields := map[string]string{
+				"sid": slaveIdTail(task.SlaveId),
```
jdef: DRY: consider const values instead of repeating `"sid"`, et al. in several places

resolver/resolver_test.go
```diff
@@ -242,7 +252,26 @@ func TestHandler(t *testing.T) {
...
+	res, err := fakeDNSwithPatterns(8053, &[]string{"test-{{.tname}}"})
+	if err != nil {
+		t.Error(err)
+	}
 
+	dns.HandleFunc("mesos.", res.HandleMesos)
+	go res.Serve("udp")
+	go res.Serve("tcp")
+
+	// wait for startup ? lame
```
jdef: there's a new way to wait for DNS server startup (`safeStartDNSResolver`) coming down the line in #144.
should probably do the same here once that PR has been merged.

records/generator.go
sttts: I have moved the pattern list out of the RecordGenerator again. Now it's a parameter to `InsertState` like all other settings coming from the config.

records/generator.go
sttts: This is not the final word yet. Next step is to identify the values we want in the patterns, now with the limited knowledge of the tasks, later from the `DiscoveryInfo` fields.

resolver/resolver_test.go
sttts: Very good. Arbitrary sleeps to synchronize code hurt the eyes. What's the status of #146? Looks like your changes and mine will conflict, so we have to adapt this or #146, whichever is merged earlier.

jdef: I think I saw #146 merged today

records/generator.go
```diff
@@ -21,6 +23,8 @@ import (
 // Will likely become map[string][]discoveryinfo
 type rrs map[string][]string
 
+var DefaultDomainPatterns = []string{"{{.TaskName}}"}
```
jdef: I tend to dislike public, mutable, global defaults. Can we find another way?

sttts: I also dislike it. Is there any better pattern than moving it into a function?

records/generator.go
```diff
@@ -232,24 +258,41 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
...
 			}
 
-			tname := labels.AsDNS952(task.Name)
-			sid := slaveIdTail(task.SlaveId)
-			tag := hashString(task.Id)
-			tail := fname + "." + domain + "."
+			// context used to resolve DomainPattern variables for the current task
+			context := PatternContext{
+				SlaveId: slaveIdTail(task.SlaveId),
+				TaskId: hashString(task.Id),
```
jdef: can we incorporate additional information into this hash? (and maybe modify the hash to actually use (at least) 32 bits). for example hashString(task.Id+task.SlaveId+task.frameworkId+task.executorId). and if we do that, do we really need the tail(slaveId) anymore?

records/generator.go
```diff
@@ -232,24 +258,41 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
...
+
 			// SRV records
-			if task.Resources.Ports != "" {
+			if task.Resources.Ports != "" && trec != "" {
 				ports := yankPorts(task.Resources.Ports)
 				for _, port := range ports {
 					var srvhost string = trec + ":" + port
-					tcp := "_" + tname + "._tcp." + tail
-					udp := "_" + tname + "._udp." + tail
+					tcp := "_" + context.TaskName + "._tcp." + tail
```
jdef: i'd actually like for us to shift our thinking re: using dynamic _service types. using standardized service names (even if they're our standard) can provide us some flexibility later on w/ respect to discovery. for example instead of the above:
```
_{taskName}._{proto}.{tail}
```
what about this:
```
_ep._{proto}.{taskName}.{tail}
```
and once we have service discovery, we can get create additional, named endpoint records:
```
_named._{proto}.{portName}.{taskName}.{tail}
--or--
_named._sub._ep._{proto}.{portName}.{taskName}.{tail}
```
(the `_ep` represents a task "endpoint")

or perhaps even this should be configurable via templates. thoughts?

records/generator.go
jdef: I'd be tempted to make it package-private and then write a public func that returns a clone. I'd still use the package-private reference within the package for ease of use and readability. And maybe a comment for the var that mentions that it's intended to be read-only.

sttts: Made cloning method out of it.

records/generator.go
sttts: The `task.Id` is unique for the framework, the framework Id is unique for the cluster, compare https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L64

Right now we have the framework name in the domain, hence can leave out the framework Id from the hash and still get unique hashes.

I completely agree with 32 bits.

sttts: Moreover, the reduction to 16 bits in `hashString` is not what is proposed for that hashing algorithm at http://www.isthe.com/chongo/tech/comp/fnv/index.html#xor-fold.

sttts: Compare #157

records/generator.go
sttts: The question is whether we expect the users to do a 2-phase service discovery, i.e. querying the task names first with a template driven domain name schema and then the ports based on the result of the first query.

If not, we really should add the template driven, dynamic approach you sketched above.

I tend to prefer the second.

sttts:
Rebased to build-fix branch. Examples are gone for now.

sttts:
Based on #216. Only c1ffcb7 and 8765c8 matter here.

PS: what is github doing with the commit order. The commit view looks completely mixed up to me. Everything is fine in git-log, with the same sha1s.

docs/docs/configuration-parameters.md
```diff
@@ -73,3 +74,5 @@ It is sufficient to specify just one of the `zk` or `masters` field. If both are
 `recurseon` controls if the DNS replies for names in the Mesos domain will indicate that recursion is available. The default value is `true`. 
 
 `enforceRFC952` will enforce an older, more strict set of rules for DNS labels. For details, see the [RFC-952](https://tools.ietf.org/html/rfc952). The default value is `false`.
+
+`templates` is a list of domain templates used to create records. The default value is `["{name}.{framework}"]`, compare the [naming documentation](naming.md).
```
tsenart: *domain name templates*

records/tmpl/templates.go
```diff
@@ -0,0 +1,199 @@
...
+// Compile compiles a Template to a fast Compiled template.
+func (t Template) Compile(spec labels.Func) (*Compiled, error) {
+	tokens := []token{}
+
+	if string(t) == "" {
+		return nil, fmt.Errorf("invalid empty template")
+	}
+
+	// split template into tokens: strings, separators and {variables}
+	varRE, err := regexp.Compile(`{[\s\w-:]*}`)
```
jdef: my gut says that a regexp is overkill for the lexer. it's more code to hand-build a lexer but it will probably be more performant and you'll be able to provide better error messages to users when a template has problems.

sttts: This code is not performance criticial.

sttts: The execution of the templates is important code path performance-wise.

records/tmpl/templates.go
```diff
@@ -0,0 +1,199 @@
...
+func (c *Compiled) Execute(ctx Context) (string, error) {
+	labels := make([]string, 0, len(c.tokens))
+	for _, t := range c.tokens {
+		label, err := t.interpolate(ctx)
+		if err != nil {
+			return "", err
+		}
+		labels = append(labels, label)
+	}
+	labels = append(labels)
```
jdef: why is this line needed?

docs/docs/naming.md
```diff
@@ -56,6 +56,56 @@ Mesos-DNS generates A records for itself that list all the IP addresses that Mes
...
+```
+{name}.{framework}
+```
+
+This gives the A records described above, e.g. `nginx.marathon.mesos.`.
+
+Additional templates can be defined using a number of *template variables*:
+
+- `framework`: the framework name, e.g. `marathon`
+- `slave-id-short`: the number in the slave ID after the last dash `-`, e.g. `0` for a slave ID `20140803-125133-3041283216-5050-2410-0`
```
jdef: what's the rationale behind the `-short` identifier? I don't think it's very robust.

docs/docs/naming.md
```diff
@@ -56,6 +56,56 @@ Mesos-DNS generates A records for itself that list all the IP addresses that Mes
...
+- `slave-id`: the complete slave ID, e.g. `20140803-125133-3041283216-5050-2410-0`
+- `task-id`: the complate task ID, e.g. `chronos.49b91a9a-3dda-11e4-a088-c20493233aa5`
+- `task-id-hash`: a hash value of the task ID, e.g. 76830, sufficiently unique for a realistic number of tasks on a cluster
+- `name`: the name of a task, provided by the framework either in the *TaskInfo* or the *DiscoveryInfo*, e.g. `nginx`
+- `version`: the version of a task, provided by the framework in the *DiscoveryInfo*, e.g. `1.0`
+- `location`: the location of a task, provider by the framework in the *DiscoveryInfo*, e.g. `europe`
+- `environment`: the environment of a task, provider by the framework in the *DiscoveryInfo*, e.g. `prod`
+- `label:<key>`: the value for the given label key, provided by the framework in the *DiscoveryInfo*, e.g. label key `canary` with label value `lanzarote`
+
+The first 6 variables should exist for every task. The following variables only
```
jdef: I think that all of this would be represented better in a table format vs. forcing me to count line by line through the above items.

records/tmpl/templates.go
sttts: It's not. Good spot.

docs/docs/naming.md
```diff
@@ -56,6 +56,56 @@ Mesos-DNS generates A records for itself that list all the IP addresses that Mes
...
+This gives the A records described above, e.g. `nginx.marathon.mesos.`.
+
+Additional templates can be defined using a number of *template variables*:
+
+- `framework`: the framework name, e.g. `marathon`
+- `slave-id-short`: the number in the slave ID after the last dash `-`, e.g. `0` for a slave ID `20140803-125133-3041283216-5050-2410-0`
+- `slave-id`: the complete slave ID, e.g. `20140803-125133-3041283216-5050-2410-0`
+- `task-id`: the complate task ID, e.g. `chronos.49b91a9a-3dda-11e4-a088-c20493233aa5`
+- `task-id-hash`: a hash value of the task ID, e.g. 76830, sufficiently unique for a realistic number of tasks on a cluster
+- `name`: the name of a task, provided by the framework either in the *TaskInfo* or the *DiscoveryInfo*, e.g. `nginx`
```
jdef: assuming that the name in the DI trumps that of the task? if so, should clarify here

docs/docs/naming.md
```diff
@@ -56,6 +56,56 @@ Mesos-DNS generates A records for itself that list all the IP addresses that Mes
...
+Mesos-DNS will interpolate the given list of *record templates* for each running
+task in the cluster. If a *referenced variable is not defined* or the value is
+the empty string, the record *will not be created*.
+
+Examples:
+- `{name}.{framework}` gives e.g. `nginx.marathon.mesos`.
+- `{name}.{location}.{framework}` gives e.g. `nginx.europe.marathon.mesos`, or no record at all for those tasks that don't define the location.
+- `{version}.{name}.{framework}` gives e.g. `1.0.nginx.marathon.mesos`, or no record at all for those tasks that don't define the version.
+- `{label:canary}.{name}` gives e.g. `lanzarote.nginx.mesos`.
+- `{label:unknown}.{name}` will give no record at all, if no label `unknown` is provided by the framework.
```
jdef: what about labels with empty values?

records/generator.go
```diff
@@ -344,42 +355,131 @@ func (rg *RecordGenerator) listenerRecord(listener string, ns string) {
...
+	compiledTmpl := compileNonCanonicalTemplates(templates, spec)
+	canonicalTmpl := compileEssentialTemplates("{name}-{task-id-hash}-{slave-id-short}.{framework}", spec)
+	tcpRFC2782Tmpl := compileEssentialTemplates("_{name}._tcp.{framework}", spec)
+	udpRFC2782Tmpl := compileEssentialTemplates("_{name}._udp.{framework}", spec)
+
+	// records to create
+	aRecordsTmpls := []recordTemplate{
+		{append(compiledTmpl, canonicalTmpl), domain, "task-ip"},
+		{append(compiledTmpl, canonicalTmpl), "slave." + domain, "slave-ip"},
+		{append(compiledTmpl, canonicalTmpl), "container." + domain, "container-ip"},
```
tsenart: This doesn't look right. 3 appends of the same thing?

records/generator.go
```diff
@@ -494,3 +588,29 @@ func getProto(pair string) (string, string, error) {
 	}
 	return h[0], h[1], nil
 }
+
+func specEachLabel(s string, spec labels.Func) string {
```
tsenart: `labels.DomainFrag`?

records/generator.go
```diff
@@ -494,3 +588,29 @@ func getProto(pair string) (string, string, error) {
...
+// return the slave number from a Mesos slave id
+func slaveIDTail(slaveID string) string {
+	fields := strings.Split(slaveID, "-")
+	return strings.ToLower(fields[len(fields)-1])
+}
+
+// BUG: The probability of hashing collisions is too high with only 17 bits.
+// NOTE: Using a numerical base as high as valid characters in DNS names would
+// reduce the resulting length without risking more collisions.
+func hashString(s string) string {
```
jdef: you're probably better off not dropping bits and just base32-encoding the hash.

docs/docs/naming.md
sttts: See above: empty values => no record

records/generator.go
sttts: not elegant, but correct.

records/generator.go
sttts: It's pretty long then. Using a hash is fine, but the collision probability should be low for the cluster sizes we target. Compare generalized birthday paradox.

docs/docs/naming.md
sttts: It's still the old canonical task domain name. I agree it's not very clever.

records/generator.go
jdef: i'd actually prefer zbase32 (http://philzimmermann.com/docs/human-oriented-base-32-encoding.txt)

sttts:
Closing. Follow-up PR #230 based on a local branch.

PR #110, opened 2015-03-21
WIP: Plugin support for mesos-dns

jdef:
Looking for some feedback here. Ultimately I'd like to write up a kubernetes-mesos plugin that exposes kubernetes services to mesos-dns, but I need some scaffolding in place first.

TODOs:
- [x] add optional HTTP handler/filter registration for plugins
- [ ] support custom flag sets for plugins, rely less on config.json file for settings

Some plugin ideas:
- examine registration records and push updates to an external DNS service (e.g. Route53)
- "generic" HTTP endpoint that accepts POSTs, allows external processes to add/remove records
  - ~~POST `/services/{service-name}?ip={ip-address}&port={port}&ttl={ttl}&priority={priority}&weight={weight}`~~
  - records would be added to `services.mesos` namespace
- emulate a SkyDNS/etcd endpoint for PUT's and DELETE's
    - `/skydns/{reverse}/{domain}/{name}`, for example:
    - `/skydns/mesos/kubernetes/services/{k8s-namespace}/{k8s-service}`
    - PUT accepts JSON [blobs][2]
    - goal: compatibility with properly configured [kube2sky][1]
- support PUSH-based updates: custom HTTP plugins that accept JSON posts, specific to frameworks
  - POST `/plugins/marathon/services/{...}`
    - plugin adds records to `services.marathon.mesos` namespace
    - secondary, out of band process, could process Marathon `/events` stream and periodically POST to this endpoint
    - alternatively, plugin could expose an endpoint for receiving a Marathon `/events` stream

[1]: https://github.com/GoogleCloudPlatform/kubernetes/blob/release-0.13/cluster/addons/dns/kube2sky/kube2sky.go
[2]: https://github.com/skynetservices/skydns/blob/master/msg/service.go

PR #152, opened 2015-05-08
WIP: Plugins (v2) API Support

### LINKED ISSUE #62: support for plugins
I'm thinking of how I might bridge kubernetes service discovery and mesos service discovery. Part of a solution might be a mesos-dns "plugin" that watches the kubernetes service registry and manages DNS records for kubernetes services inside mesos-dns (so that mesos tasks using mesos-dns services for resolution can also see kubernetes services). The other half of a solution might be an additional kubernetes service endpoint controller that watches mesos tasks (like mesos-dns does) and manages kubernetes endpoint records (so that kubernetes pods can find mesos services hosted by other frameworks).

Any thoughts re: a plugin model?

jdef:
- supersedes #110 
- fixes #62
- high-level changelog
  - hide master polling implementation behind event (`records/config`) model
    - we still poll master :(
    - plugins don't have to poll, they can push
  - plugins!
    - can register additional HTTP containers
    - can register pre/post processing filters for records generated from mesos tasks
    - can wrap the record generation function for tasks records
    - can register additional record sources
    - HTTP and DNS servers both run as plugins
- TODO
  - [ ] document features, provide examples, add more tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ DEPS = $(shell go list -f '{{range .TestImports}}{{.}} {{end}}' ./...)
 
 default: all
 
-all: restoredeps test build
+all: test build
 
 restoredeps:
 	@echo "--> Restoring build dependencies"
--- a/app.go
+++ b/app.go
@@ -1,0 +1,196 @@
+package main
+
+import (
+	"time"
+
+	"github.com/emicklei/go-restful"
+	"github.com/mesosphere/mesos-dns/logging"
+	"github.com/mesosphere/mesos-dns/plugins"
+	"github.com/mesosphere/mesos-dns/records"
+	"github.com/mesosphere/mesos-dns/resolver"
+	"github.com/mesosphere/mesos-dns/util"
+)
+
+const (
+	zkInitialDetectionTimeout = 30 * time.Second
+)
+
+type errorHandlerFunc func(string, error)
+
+type app struct {
+	config     records.Config
+	resolver   *resolver.Resolver
+	filters    plugins.FilterSet
+	ready      chan struct{} // when closed, indicates that initialization has completed
+	done       chan struct{} // when closed, indicates that run has completed
+	errHandler errorHandlerFunc
+}
+
+type pluginContext struct {
+	*app
+	pluginName string
+}
+
+func newApp(eh errorHandlerFunc) *app {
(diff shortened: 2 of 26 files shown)


sttts:
LGTM as far as I can judge without actually testing it.

jdef:
xref #62

F21:
Any updates on this?

jdef:
@F21 we're in the process of rationalizing the decisions that drove this prototype. we'll probably end up with a v3 that reuses some of these ideas but probably somewhat more narrow in scope -- this is pretty huge PR. did you have any specific feedback to offer with respect to approach that this PR demonstrates?

F21:
@jdef Not specific feedback pertaining to this PR, but I am currently setting up a test mesos cluster with the kubernetes-mesos framework. Kubernetes appears to have a kube-dns add-on for resolving services, so I would love to see how that will get integrated with the mesos-dns project.

In order words, I would love to be able to resolve things like `blah.marathon.mesos` or `namenode1.hdfs.mesos` from within a container in a kubernetes pod as well as resolve things like `my-svc.my-namespace.svc.cluster.local` from outside kubernetes to point to a kubernetes object.

jdef:
xref https://github.com/mesosphere/kubernetes-mesos/issues/125

jdef:
@F21 kubernetes has various options for configuring DNS resolution within pod containers. I think that the default is that pod CT's inherit the node's resolve.conf settings. So if your node's already have a nameserver record that points to mesos-dns your pods should already be able to resolve mesos services.

An ideal mesos-dns/k8s plugin would provide records in mesos-dns, in the .mesos domain, that point to k8s services. There are complicating factors here: for example, k8s service IPs are magical because they're managed by kube-proxy/iptables; only those tasks running on slaves that are *also* running kube-proxy would be able to access those service IPs. Unless we changed how the k8s framework deployed kube-proxy across the cluster (and there are complications there too).

jdef:
FWIW the FrameworkInfo.Labels field is should land in mesos-0.24.0:
- https://issues.apache.org/jira/browse/MESOS-2841

this facilitates mesos-dns discovering mesos-dns-plugin capabilities offered by frameworks.

jdef:
Closing this out since no one seems interested in landing this.

PR #158, opened 2015-05-16
add resolver validation, clean up host:port joins

jdef:
DNS server should error out if ExternalOn is enabled and resolver configuration is invalid.

### DIFF
--- a/records/validation.go
+++ b/records/validation.go
@@ -1,0 +1,25 @@
+package records
+
+import (
+	"fmt"
+	"net"
+)
+
+// validateResolvers checks that each resolver in the list is a properly formatted IP address.
+// duplicate resolvers in the list are not allowed.
+// returns nil if the resolver list is empty, or else all resolvers in the list are valid.
+func validateResolvers(rs []string) error {
+	ips := map[string]struct{}{}
+	for _, r := range rs {
+		ip := net.ParseIP(r)
+		if ip == nil {
+			return fmt.Errorf("illegal IP specified for resolver %q", r)
+		}
+		ipstr := ip.String()
+		if _, found := ips[ipstr]; found {
+			return fmt.Errorf("duplicate resolver IP specified: %v", r)
+		}
+		ips[ipstr] = struct{}{}
+	}
+	return nil
+}
--- a/records/validation.go
+++ b/records/validation.go
@@ -7,9 +7,12 @@ import (
 
 // validateMasters checks that each master in the list is a properly formatted host:ip pair.
 // duplicate masters in the list are not allowed.
-// returns nil if the masters list if empty, or else all masters in the list are valid.
+// returns nil if the masters list is empty, or else all masters in the list are valid.
 func validateMasters(ms []string) error {
-	valid := map[string]struct{}{}
+	if len(ms) == 0 {
+		return nil
+	}
+	valid := make(map[string]struct{}, len(ms))
 	for i, m := range ms {
 		h, p, err := net.SplitHostPort(m)
 		if err != nil {
@@ -33,7 +36,10 @@ func validateMasters(ms []string) error {
 // duplicate resolvers in the list are not allowed.
 // returns nil if the resolver list is empty, or else all resolvers in the list are valid.
 func validateResolvers(rs []string) error {
-	ips := map[string]struct{}{}
+	if len(rs) == 0 {
+		return nil
+	}
+	ips := make(map[string]struct{}, len(rs))
 	for _, r := range rs {
 		ip := net.ParseIP(r)
 		if ip == nil {
--- a/records/validation.go
+++ b/records/validation.go
@@ -5,6 +5,30 @@ import (
 	"net"
 )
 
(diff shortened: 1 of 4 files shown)


jdef:
- resolves #161

records/validation.go
```diff
@@ -0,0 +1,49 @@
+package records
+
+import (
+	"fmt"
+	"net"
+)
+
+// validateMasters checks that each master in the list is a properly formatted host:ip pair.
+// duplicate masters in the list are not allowed.
+// returns nil if the masters list if empty, or else all masters in the list are valid.
```
tsenart: masters list **is** empty

tsenart: Wouldn't it make sense to return an error if no masters are given?

records/validation.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		valid[m] = struct{}{}
+	}
+	return nil
+}
+
+// validateResolvers checks that each resolver in the list is a properly formatted IP address.
+// duplicate resolvers in the list are not allowed.
+// returns nil if the resolver list is empty, or else all resolvers in the list are valid.
+func validateResolvers(rs []string) error {
```
tsenart: Using `rs ...string` makes for a more ergonomic API.

records/validation.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"fmt"
+	"net"
+)
+
+// validateMasters checks that each master in the list is a properly formatted host:ip pair.
+// duplicate masters in the list are not allowed.
+// returns nil if the masters list if empty, or else all masters in the list are valid.
+func validateMasters(ms []string) error {
+	valid := map[string]struct{}{}
```
tsenart: Since you know the upper bound of the size of the set, you can pre-allocate it. Same as below.
```go
valid := make(map[string]struct{}, len(ms))
```

records/validation.go
```diff
@@ -0,0 +1,49 @@
...
+	for i, m := range ms {
+		h, p, err := net.SplitHostPort(m)
+		if err != nil {
+			return fmt.Errorf("illegal host:port specified for master %q", ms[i])
+		}
+		// normalize ipv6 addresses
+		if ip := net.ParseIP(h); ip != nil {
+			h = ip.String()
+			m = h + "_" + p
```
tsenart: Not that it matters too much here but is there any reason why we can't use the more common `:` separator?

records/validation_test.go
```diff
@@ -0,0 +1,72 @@
...
+		{[]string{"a"}, false},
+		{[]string{"a", "b"}, false},
+		{[]string{"1.2.3.4"}, true},
+		{[]string{"1.2.3.4.5"}, false},
+		{[]string{"1.2.3.4", "1.2.3.4"}, false},
+		{[]string{"1.2.3.4", "5.6.7.8"}, true},
+		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b"}, true},
+		{[]string{"2001:db8:3c4d:15::1a2f:1a2b"}, true},
+		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b", "2001:db8:3c4d:15::1a2f:1a2b"}, false},
+	}
```
tsenart: Nice tests!! :+1:

records/validation.go
```diff
@@ -0,0 +1,49 @@
...
+// returns nil if the resolver list is empty, or else all resolvers in the list are valid.
+func validateResolvers(rs []string) error {
+	ips := map[string]struct{}{}
+	for _, r := range rs {
+		ip := net.ParseIP(r)
+		if ip == nil {
+			return fmt.Errorf("illegal IP specified for resolver %q", r)
+		}
+		ipstr := ip.String()
+		if _, found := ips[ipstr]; found {
```
tsenart: FWIW a common idiom is to name this variable `ok`.

tsenart:
LGTM

records/validation.go
jdef: no, because config.Masters functions as "fallback masters" if the current leading master can't be detected from zk. If zk is not configured AND masters == nil, then the config analyzer should error out -- but we don't handle that here.

records/validation.go
jdef: ipv6 addresses use `:` as segment delimiters, so I wanted to use a different delimiter when generating the map keys to avoid confusion. `_` seemed ok because it's not allowed in ipv4 or ipv6 IP's.

records/validation.go
jdef: The effective-go docs actually use "present" for this case. I've seen "exists", "found", "present", and "ok" in most (what I would consider) well written Go.
```go
_, present := timeZone[tz]
```

records/validation_test.go
jdef: thanks for the inspiration!

records/validation.go
jdef: not super-concerned about ergonomics here, it's going to be invoked either by a test case or by the config analyzer, both of which pass in slices.

PR #160, opened 2015-05-19
adding systemd tutorial for running mesos-dns

kensipe:
This is the systemd tutorial.  I tested the systemd components against SUSE 12.     The challenge as I see if is that no one is using Suse :)    It was either Suse or CoreOS on AWS or GCE.    There are no other platforms there that support systemd.     I was going to go down the CoreOS route, but it is slightly different.  

Regardless... the information here is good.   If someone is able to install a mesos cluster on a linux version that supports systemd, then the instructions here are good.

### DIFF
--- a/docs/docs/tutorial-systemd.md
+++ b/docs/docs/tutorial-systemd.md
@@ -1,0 +1,311 @@
+---
+title: Mesos-DNS using systemd
+---
+
+## Mesos-DNS using systemd
+
+This is a step-by-step tutorial for running [Mesos-DNS](https://github.com/mesosphere/mesos-dns) with [systemd](http://freedesktop.org/wiki/Software/systemd/). 
+
+### Step 1: Launch a Mesosphere cluster on a systemd platform
+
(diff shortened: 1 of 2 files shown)


kensipe:
@lloesche I would love a review.

kozyraki:
This is great. I will wait until @lloesche or @spahl can add any comments before mering this.

docs/docs/tutorial-systemd.md
```diff
@@ -0,0 +1,311 @@
...
+
+`sudo systemctl start mesos-dns`
+
+### Step 4: Configure cluster nodes
+
+Next, we will configure all nodes in our cluster to use Mesos-DNS as their DNS server. Access each node through ssh and execute: 
+
+
+```
+sudo sed -i '1s/^/nameserver 10.14.245.208\n /' /etc/resolv.conf
```
jdef: This works well if your nodes are configured with static ip. If they use dhcp, then dhclient (or something similar) will probably blow away this change at some point. For dhclient, you can modify dhclient.conf:
```
prepend domain-name-servers 10.14.245.208;
```

kensipe: smart!  thanks @jdef

lloesche:
LGTM

lloesche:
Reading about compiling mesos-dns... maybe we should tell people about mesos-dns-pkg. You can run it on OS X or Linux, anywhere you have Docker. Then it's just a simple "make el7" and you get el7 RPM packages, which include the systemd unit file. Or "make debian" and you get Debian packages.
I recently refactored and updated it to work with the most recent changes to mesos-dns.

spahl:
LGTM

kozyraki:
@lloesche we have a section on packaging in the docs that points to mesos-dns-pkg. Feel free to update any part of the documentation to link the two together. 

Of course, we can also merge them in a single repo to make it easier to see the link.

PR #169, opened 2015-06-11
Generate masterN.mesos record for current leader

### LINKED ISSUE #165: masterN.mesos record not being generated for current leader
The code in
https://github.com/mesosphere/mesos-dns/blob/master/records/generator.go#L309

makes it so that https://github.com/mesosphere/mesos-dns/blob/master/records/generator.go#L321 is being skipped and no masterN.mesos record is being generated for the current leader.

jdef:
Fixes #165

### DIFF
--- a/records/generator.go
+++ b/records/generator.go
@@ -5,6 +5,7 @@ package records
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"hash/fnv"
 	"io/ioutil"
 	"net"
@@ -286,15 +287,19 @@ func (rg *RecordGenerator) masterRecord(domain string, masters []string, leader 
 	h := strings.Split(leader, "@")
 	if len(h) < 2 {
 		logging.Error.Println(leader)
+		return // avoid a panic later
 	}
-	ip, port, err := getProto(h[1])
+	leaderAddress := h[1]
+	ip, port, err := getProto(leaderAddress)
 	if err != nil {
 		logging.Error.Println(err)
+		return
 	}
 	arec := "leader." + domain + "."
 	rg.insertRR(arec, ip, "A")
 	arec = "master." + domain + "."
 	rg.insertRR(arec, ip, "A")
+
 	// SRV records
 	tcp := "_leader._tcp." + domain + "."
 	udp := "_leader._udp." + domain + "."
@@ -303,22 +308,43 @@ func (rg *RecordGenerator) masterRecord(domain string, masters []string, leader 
 	rg.insertRR(udp, host, "SRV")
 
 	// if there is a list of masters, insert that as well
-	for i, master := range masters {
-
-		// skip leader
-		if leader == master {
-			continue
-		}
+	addedLeaderMasterN := false
+	idx := 0
+	for _, master := range masters {
 
 		ip, _, err := getProto(master)
 		if err != nil {
 			logging.Error.Println(err)
+			continue
 		}
 
 		// A records (master and masterN)
-		arec := "master." + domain + "."
+		if master != leaderAddress {
+			arec := "master." + domain + "."
+			added := rg.insertRR(arec, ip, "A")
+			if !added {
+				// duplicate master?!
+				continue
+			}
+		}
+
+		if master == leaderAddress && addedLeaderMasterN {
+			// duplicate leader in masters list?!
+			continue
+		}
+
+		arec := "master" + strconv.Itoa(idx) + "." + domain + "."
 		rg.insertRR(arec, ip, "A")
-		arec = "master" + strconv.Itoa(i) + "." + domain + "."
+		idx++
+
+		if master == leaderAddress {
+			addedLeaderMasterN = true
+		}
+	}
+	// flake: we ended up with a leader that's not in the list of all masters?
+	if !addedLeaderMasterN {
+		logging.Error.Printf("warning: leader %q is not in master list", leader)
+		arec = "master" + strconv.Itoa(idx) + "." + domain + "."
 		rg.insertRR(arec, ip, "A")
 	}
 }
@@ -375,37 +401,47 @@ func (rg *RecordGenerator) setFromLocal(host string, ns string) {
 	}
 }
 
-// insertRR inserts host to name's map
-// REFACTOR when storage is updated
-func (rg *RecordGenerator) insertRR(name string, host string, rtype string) {
-	logging.VeryVerbose.Println("[" + rtype + "]\t" + name + ": " + host)
-
+func (rg *RecordGenerator) exists(name string, host string, rtype string) bool {
 	if rtype == "A" {
 		if val, ok := rg.As[name]; ok {
 			// check if A record already exists
 			// identical tasks on same slave
 			for _, b := range val {
 				if b == host {
-					return
+					return true
 				}
 			}
-			rg.As[name] = append(val, host)
-		} else {
-			rg.As[name] = []string{host}
 		}
 	} else {
 		if val, ok := rg.SRVs[name]; ok {
 			// check if SRV record already exists
 			for _, b := range val {
 				if b == host {
-					return
+					return true
 				}
 			}
-			rg.SRVs[name] = append(val, host)
-		} else {
-			rg.SRVs[name] = []string{host}
 		}
 	}
+	return false
+}
+
+// insertRR adds a record to the appropriate record map for the given name/host pair,
+// but only if the pair is unique. returns true if added, false otherwise.
+// TODO(???): REFACTOR when storage is updated
+func (rg *RecordGenerator) insertRR(name string, host string, rtype string) bool {
+	logging.VeryVerbose.Println("[" + rtype + "]\t" + name + ": " + host)
+
+	if rg.exists(name, host, rtype) {
+		return false
+	}
+	if rtype == "A" {
+		val := rg.As[name]
+		rg.As[name] = append(val, host)
+	} else {
+		val := rg.SRVs[name]
+		rg.SRVs[name] = append(val, host)
+	}
+	return true
 }
 
 // returns an array of ports from a range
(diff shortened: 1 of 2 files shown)


records/generator.go
```diff
@@ -375,37 +401,47 @@ func (rg *RecordGenerator) setFromLocal(host string, ns string) {
...
+// TODO(???): REFACTOR when storage is updated
+func (rg *RecordGenerator) insertRR(name string, host string, rtype string) bool {
+	logging.VeryVerbose.Println("[" + rtype + "]\t" + name + ": " + host)
+
+	if rg.exists(name, host, rtype) {
+		return false
+	}
+	if rtype == "A" {
+		val := rg.As[name]
+		rg.As[name] = append(val, host)
```
tsenart: What do you think of making this a nested map?
```go
if _, ok := rg.records[rtype]; !ok {
    rg.records[rtype] = map[string][]string{}
}
rg.records[rtype][name] = append(rg.records[rtype][name], host)
```

records/generator.go
```diff
@@ -448,6 +484,9 @@ func slaveIdTail(slaveID string) string {
 // zk://username:password@host1:port1,host2:port2,.../path
 // file:///path/to/file (where file contains one of the above)
 func getProto(pair string) (string, string, error) {
```
tsenart: I find it confusing with call this function getProto when we're using it with ip:port pairs too. We're essentially just splitting and returning the tokens as different return arguments so this is generalizable.

```go
func split(s, sep string) (string, string, error) {
    ps := strings.SplitN(s, sep, 2)
    if len(ps) != 2 {
        return "", "", fmt.Errorf("%q not found in %q", sep, s)
    }
    return ps[0], ps[1], nil
}
```

records/generator_test.go
```diff
@@ -13,6 +14,129 @@ func init() {
...
+	type expectedRR struct {
+		name  string
+		host  string
+		rtype string
+	}
+	tt := []struct {
+		domain  string
+		masters []string
+		leader  string
+		exists  []expectedRR
```
tsenart: Can we call this `want` or `expect`?

records/generator.go
```diff
@@ -375,37 +401,47 @@ func (rg *RecordGenerator) setFromLocal(host string, ns string) {
...
+	return false
+}
+
+// insertRR adds a record to the appropriate record map for the given name/host pair,
+// but only if the pair is unique. returns true if added, false otherwise.
+// TODO(???): REFACTOR when storage is updated
+func (rg *RecordGenerator) insertRR(name string, host string, rtype string) bool {
+	logging.VeryVerbose.Println("[" + rtype + "]\t" + name + ": " + host)
+
+	if rg.exists(name, host, rtype) {
```
tsenart: Why wouldn't we be able to change a record? Perhaps this check could be better placed outside this function in call sites where it's actually needed.

records/generator.go
```diff
@@ -286,15 +287,19 @@ func (rg *RecordGenerator) masterRecord(domain string, masters []string, leader
 	h := strings.Split(leader, "@")
```
tsenart: I'd make the leader be the first item of the masters array.
```go
func (rg *RecordGenerator) masterRecord(domain string, masters ...string) {
```

records/generator_test.go
```diff
@@ -13,6 +14,129 @@ func init() {
...
+			[]expectedRR{
+				{"leader.foo.com.", "6", "A"},
+				{"master.foo.com.", "6", "A"},
+				{"master.foo.com.", "8", "A"},
+				{"master1.foo.com.", "6", "A"},
+				{"master0.foo.com.", "8", "A"},
+				{"_leader._tcp.foo.com.", "leader.foo.com.:7", "SRV"},
+				{"_leader._udp.foo.com.", "leader.foo.com.:7", "SRV"},
+			}},
+		{"foo.com", []string{"8:9", "6:7"}, "5@6:7",
```
tsenart: Can we add short descriptive comments explaining what each of these tests cases is verifying (where it makes sense)?

records/generator.go
```diff
@@ -375,37 +401,47 @@ func (rg *RecordGenerator) setFromLocal(host string, ns string) {
...
-			rg.SRVs[name] = []string{host}
 		}
 	}
+	return false
+}
+
+// insertRR adds a record to the appropriate record map for the given name/host pair,
+// but only if the pair is unique. returns true if added, false otherwise.
+// TODO(???): REFACTOR when storage is updated
+func (rg *RecordGenerator) insertRR(name string, host string, rtype string) bool {
```
tsenart: Argument types can be coalesced.
```go
func (rg *RecordGenerator) insertRR(name, host, rtype string) bool {
```

records/generator.go
```diff
@@ -375,37 +401,47 @@ func (rg *RecordGenerator) setFromLocal(host string, ns string) {
 	}
 }
 
-// insertRR inserts host to name's map
-// REFACTOR when storage is updated
-func (rg *RecordGenerator) insertRR(name string, host string, rtype string) {
-	logging.VeryVerbose.Println("[" + rtype + "]\t" + name + ": " + host)
-
+func (rg *RecordGenerator) exists(name string, host string, rtype string) bool {
```
tsenart:
```go
func (rg *RecordGenerator) exists(name, host, rtype string) bool
```

records/generator.go
```diff
@@ -375,37 +401,47 @@ func (rg *RecordGenerator) setFromLocal(host string, ns string) {
...
 	if rtype == "A" {
```
tsenart: I'd maintain an inverted index of host to names for this method.
```go
_, ok := rg.names[rtype][host]
return ok
```

records/generator.go
```diff
@@ -303,22 +308,43 @@ func (rg *RecordGenerator) masterRecord(domain string, masters []string, leader
 	rg.insertRR(udp, host, "SRV")
```
tsenart: I'd love to see this method have no side effects. It would make it easier to reason about and test. We'd isolate the data generation from the mutation effect.
```go
records := rg.masterRecords(tc.domain, tc.masters...) // referentially transparent: same inputs, same outputs, no side r
if err := rg.insert(records...); err != nil {
   // handle error
}
```

records/generator.go
tsenart: Or even better:
```go
if _, ok := rg.records[rtype]; !ok {
    rg.records[rtype] = map[string]map[string]struct{}{}
}

if _, ok := rg.records[rtype][name]; !ok {
    rg.records[rtype][name] = map[string]struct{}{}
    rg.records[rtype][host] = map[string]struct{}{}
}

rg.records[rtype][name][host] = struct{}{}
rg.records[rtype][host][name] = struct{}{}
```

records/generator.go
kozyraki: We should revisit the key datastructures holistically in order to support discoveryInfo, flexible naming, and http interface.

kozyraki:
@jdef @tsenart Gentlemen, is this good to go from your point of view? Cannot tell for sure.

jdef:
Not ready for merge, needs additional work.

records/generator.go
jdef: The current func implementation makes an assumption about the order of `masters`: it's the order in which you expect the ernumerated `masterX` records to be created. I think this is probably important: if a new leader is elected, you may not want it to become `master0` simply because it's the leader. You probably want your DNS records to change as little as possible. And this func should have the least impact on enumeration order, or name/IP mappings - it's just creating the records. So let the caller do the work of ordering, sorting, whatever.. the `masters` list if a different outcome is desired.

Another consequence of the current overall mesos-dns app implementation is that the leader may not even be in the `masters` list at some point in time. `masters` is really *fallback-masters* (only consider these to be masters if I can't find a leader via ZK). At some point in time, they may not actually be masters any more. Consider a cluster of 3 nodes that suffers the loss of a member, and gains a a new member (VM crashed, was replaced by another VM). And the cycle repeats several times. You end up with a set of running masters (and leader) that's different than the set of statically configured fallback masters.

So the func tries to index the masters as they're listed and begrudgingly assigns the leading master an index out-of-band if it's not actually listed in the masters list. There are probably better ways to do it. But I'm really not trying to optimize that here. I'm just trying to fix #165 without introducing additional inconsistency into the system.

I should probably document some of this in the Go comments of the func.

records/generator.go
jdef: This can be tackled in a separate PR

records/generator.go
jdef: agreed

records/generator.go
jdef: @kozyraki agreed, in a separate PR

records/generator.go
jdef: agreed

records/generator.go
jdef: i'm just maintaining existing functionality. i refactored so that I could use `exists` in a test. trying to keep the scope of this PR small.

that said, there's probably a lot of room for refactoring in the system. there's also lots of work planned for features and the bulk of this implementation is going to change at that point. probably not worth optimizing this now.

records/generator.go
jdef: good suggestions. if the system wasn't about to undergo major re-implementation to support new features, i'd agree to an approach like this.

records/generator.go
jdef: agree. `getProto` is abused.

records/generator_test.go
jdef: will do

records/generator_test.go
jdef: will do

records/generator.go
jdef: instead of `SplitN` we should probably attempt to `net.SplitHostPort` -- and if that fails we could to parse as a URL. Either way, I think that work should be done in a separate PR.

records/generator.go
jdef: I've added my prior ramblings to the masterRecord godocs

tsenart: :+1:

tsenart:
LGTM in regards to the scope of the PR. Let's review the other suggestions after we have this through.

PR #177, opened 2015-06-20
fixes #149: enforcement of rfc952 is optional, default is now rfc1123

### LINKED ISSUE #149: Make truncation to 24 characters optional
But document it is not standard

jdef:
* added config option  `EnforceRFC952` for interop w/ systems that rely on rfc952 labels
* default record generation spec (rfc1123) allows:
  * longer labels (63 chars per segment)
  * host records that start with digits


sttts:
Please add the new config option to the docs.

Otherwise, LGTM

jdef:
Thanks for the review @sttts

records/labels/dns1123.go
```diff
@@ -0,0 +1,50 @@
+package labels
+
+const DNS1123MaxLength int = 63
+
+// mangle the given name to be compliant as a DNS1123 name "component".
+// the returned result should match the regexp:
+//    ^[a-z0-9]([-a-z0-9]*?[a-z0-9])?$
+// returns "" if the name cannot be mangled.
+func AsDNS1123(name string) string {
```
tsenart: This code looks very similar to `AsDNS953`. Can we refactor?

records/labels/dns1123.go
```diff
@@ -0,0 +1,50 @@
...
+		sz = DNS1123MaxLength
+	}
+	last := sz - 1
+	label := make([]byte, sz, sz)
+	ll := 0
+	la := -1 // index of last alphanumeric
+	for _, c := range name {
+		b := dns952table[uint8(c)]
+		switch {
+		case b == -int32('-'):
```
tsenart: Why aren't we dealing with runes here? `case b == '-'`

tsenart: I think the translation table could hold runes instead ints.

docs/docs/configuration-parameters.md
```diff
@@ -71,3 +71,5 @@ It is sufficient to specify just one of the `zk` or `masters` field. If both are
 `SOAMinttl` is the minimum TTL field in the SOA record for the Mesos domain. For details, see the [RFC-2308](https://tools.ietf.org/html/rfc2308). The default value is `60`.
 
 `recurseon` controls if the DNS replies for names in the Mesos domain will indicate that recursion is available. The default value is `true`. 
+
+`enforceDNS952` will enforce an older, more strict set of rules for DNS labels. For details, see the [RFC-952](https://tools.ietf.org/html/rfc952). The default value is `false`.
```
tsenart: I think this should be `enforceRFC952`. Also, don't we want the default to be true for backwards compatibility?

records/labels/spec.go
```diff
@@ -0,0 +1,43 @@
+package labels
+
+import (
+	"fmt"
+)
+
+// LabelSeparator delimits concatenated labels.
+const LabelSeparator = '.'
```
tsenart: `Separator`

records/labels/spec.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Spec is a label specification interface
+type Spec interface {
```
tsenart: `type Mangler interface`

records/labels/spec.go
```diff
@@ -0,0 +1,43 @@
...
+	// Mangle transforms an input string to produce an output string that is
+	// compliant with the specification. If the input string is already
+	// compliant then it is returned unchanged.
+	Mangle(string) string
+}
+
+// SpecFunc implements Spec
+type SpecFunc func(string) string
```
tsenart: `type ManglerFunc func(string) string`

records/labels/spec.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (sf SpecFunc) Mangle(input string) string {
+	return sf(input)
+}
+
+type HostNameSpec int
```
tsenart: `type Spec int` or `type RFC int`

records/labels/spec.go
```diff
@@ -0,0 +1,43 @@
...
+
+const (
+	// see http://www.rfc-base.org/txt/rfc-952.txt
+	HostNameSpec952 HostNameSpec = iota
```
tsenart: I'm gonna be cute here :)

```go
const (
    RFC952 = RFC(952)
    RFC1123 = RFC(1123)
)
```

records/labels/spec.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// see http://www.rfc-base.org/txt/rfc-1123.txt
+	HostNameSpec1123
+)
+
+func (spec HostNameSpec) Spec() Spec {
```
tsenart: We can simply implement the `Mangler` interface here.

records/generator.go
```diff
@@ -81,8 +81,14 @@ func (rg *RecordGenerator) ParseState(leader string, c Config) error {
...
 
+	hostSpec := labels.HostNameSpec1123
+	if c.EnforceDNS952 {
+		hostSpec = labels.HostNameSpec952
+	}
+	nameSpec := hostSpec.Spec()
+
 	// insert state
-	rg.InsertState(sj, c.Domain, c.SOARname, c.Listener, c.Masters)
+	rg.InsertState(sj, c.Domain, c.SOARname, c.Listener, c.Masters, nameSpec)
```
tsenart: The spec or RFC should be constant across all Inserts so it makes sense for it to be a RecordGenerator field and not an argument to this method.

records/labels/dns1123.go
jdef: I like using int32 because I'm using negative's to represent certain things. A negative rune just seemed weird, or at least more weird than a negative int32.

docs/docs/configuration-parameters.md
jdef: I intentionally made the default false because I think 952 is hurting more people than those it helps. Some quick googling shows that legacy OS impls have problems with spec > 952 ... I'm less worried about breaking those. This option serves as a fallback that allows people with such systems to continue to use the old scheme. Using the new scheme by default will be less painful for most.

records/generator.go
jdef: Actually, I had initially modeled it that way. I ended up doing it this way instead because it's consistent with the existing design. It also further highlights a weakness of the current design (I think we're passing in too many parameters here, especially those that will not change) that should be refactored when we re-do the storage implementation. So it should be exceedingly obvious that this is not the right approach and we need to change it.

records/labels/spec.go
jdef: This works until you get an RFC named "802.11ac" which is no longer an int. I think we should stick to an iota-based enumeration.

records/labels/spec.go
jdef: I like `HostNameSpec` because it's more specific to the intent of the type.

records/labels/tables_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"testing"
+)
+
+func BenchmarkAsDNS952(b *testing.B) {
+	const (
+		original = "89f.gsf---g_7-fgs--d7fddg-"
+		expected = "f-gsf---g-7-fgs--d7fddg"
+	)
+
+	asDNS952 := HostNameSpec952.Mangler().Mangle
```
tsenart: I think it'd be really nice if HostNameSpec would implement the Mangler interface directly.

tsenart:
FWIW `802.11ac` is IEEE not RFC. RFCs follow a pure monotonic incremental number: http://www.rfc-editor.org/rfc-index2.html

records/labels/spec.go
tsenart: FWIW `802.11ac` is IEEE not RFC. RFCs follow a pure monotonic incremental number: http://www.rfc-editor.org/rfc-index2.html

records/labels/tables_test.go
```diff
@@ -0,0 +1,103 @@
...
+	)
+
+	asDNS1123 := HostNameSpec1123.Mangler().Mangle
+	// run the asDNS1123 func b.N times
+	for n := 0; n < b.N; n++ {
+		if actual := asDNS1123(original); actual != expected {
+			b.Fatalf("expected %q instead of %q", expected, actual)
+		}
+	}
+}
```
tsenart: It'd be interesting to compare the performance of this implementation compared to `http://golang.org/pkg/strings/#Map`.

jdef:
Yep, I conflated them this morning :)

I stand by my statement that a simple enumeration is the way to go.

records/labels/tables_test.go
jdef: Thanks for pointing out this func, I hadn't seen it before. Funny that you point it out, I actually spent some time tonight playing with the idea of using a func-based approach like this. The implementation turned out looking pretty awful and seemed much more complicated than a simple lookup table. The whole goal of the exercise was to avoid (potential) inefficiencies w/ respect to branch prediction. But the resulting impl looked so awful that I rolled back the changes before I got to the point of testing it.

jdef:
@tsenart  Took your advice re: the simpler `Mangler` impl. PTAL

records/labels/spec.go
```diff
@@ -0,0 +1,58 @@
...
+	SeparatorChar = '.'
+
+	DNS952MaxLength  int = 24
+	DNS1123MaxLength int = 63
+
+	// see http://www.rfc-base.org/txt/rfc-952.txt
+	HostNameSpecType952 HostNameSpecType = iota
+
+	// see http://www.rfc-base.org/txt/rfc-1123.txt
+	HostNameSpecType1123
```
tsenart: I believe this can be further simplified. No need for `HostNameSpecType`.
```go
var (
    RFC952   = HostNameSpec{table:  makeDNS952Table(), maxlen: 24}
    RFC1123  = HostNameSpec{table:  makeDNS1123Table(), maxlen: 63}
)
```

jdef: Yep, probably. I'll keep the constants for length though - trying to avoid
introducing magic numbers not assigned to constants.

jdef: I remember my initial reasoning for doing it this way: I didn't want to
create globals that were non-constant but that should have constant,
deterministic behavior. Upon looking at my interfaces I think they still
leak a bit, still allowing for the original structs to be modified by code
outside of the package (see http://play.golang.org/p/vVInOnjtt0). I'll stop
leaking the pointers to the original structs.

tsenart: I usually just accept this as a limitation of Go not having immutable
variables. I don't think it's necessary to consider the risk of a package
variable being overwritten by users as that is not its intended use.

jdef: I tend to make exceptions for Go in this area when it comes to error
types/implementations. Otherwise, I think the package should probably guard
against clients (possibly unintentionally) stomping on the original values.

jdef: Also, someone mentioned RFC2181 as a standard that we should look at. I
skimmed this earlier today. It looks to me like the length requirements are
not changed from 1123. Thoughts?

https://www.ietf.org/rfc/rfc2181.txt

jdef:
we have decided to convert HostNameSpec to an interface (private) so that it stays immutable outside the package and we can avoid the additional *Type constants. we can move the table init func calls inside the struct declarations.

PR #178, opened 2015-06-25
Adds Dockerfile for development

jfarrell:
Add docker build environment to simplify development

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,41 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


tsenart:
Would you please elaborate on how this simplifies development?

jfarrell:
hey @tsenart, this provides a repeatable development environment which is always the same for dev and ci purposes (travis is in the works of adding docker support as well, currently in beta). If you dont think that this would be used by a broader audience then the pr can be closed

tsenart:
As long as we don't run tests with infrastructure dependencies such as Mesos or Zookeeper, using Dockery is overkill. We already achieve repeatable code builds with dependency vendoring (Godeps).

jfarrell:
The Dockerfile does not add the source or any of its vendored dependencies listed in GoDeps, its intent is to provide a build and test environment that is always the same, so that someone developing locally or using travis/jenkins/circle can have the same base go env setup across the board. Since this does not seem to be a need i'll go ahead and close out the pr

PR #183, opened 2015-07-17
Add _container.task.framework.mesos A records with container IP

sttts:
For each existing A record for a task with a container IP another A record is added with the `_container` prefix, resolving to the container IP instead of the host IP.

The container IP is passed in a status label of a task in state.json, e.g. 

```json
{
  "executor_id": "",
  "framework_id": "20140703-014514-3041283216-5050-5348-0000",
  "id": "liquor-store.b8db9f73-562f-11e4-a088-c20493233aa5",
  "name": "liquor-store",
  "slave_id": "20140803-125133-3041283216-5050-2410-0",
  "state": "TASK_RUNNING",
  "statuses": [
    {
      "state": "TASK_RUNNING",
      "timestamp": 1413572390.09275,
      "labels": [
        {
          "key": "Docker.NetworkSettings.IPAddress",
          "value": "10.3.0.1"
        }
      ]
    }
  ]
}
```

Compare https://reviews.apache.org/r/36537/ and https://issues.apache.org/jira/browse/MESOS-3076

### DIFF
--- a/factories/fake.json
+++ b/factories/fake.json
@@ -186,7 +186,13 @@
                     "statuses": [
                         {
                             "state": "TASK_RUNNING",
-                            "timestamp": 1409096598.68885
+                            "timestamp": 1409096598.68885,
+                            "labels": [
+                                {
+                                    "key": "foo",
+                                    "value": "bar"
+                                }
+                            ]
                         }
                     ]
                 },
@@ -206,7 +212,49 @@
                     "statuses": [
                         {
                             "state": "TASK_RUNNING",
-                            "timestamp": 1413572390.09275
+                            "timestamp": 1413572371.04321,
+                            "labels": [
+                                {
+                                    "key": "foo",
+                                    "value": "bar"
+                                },
+                                {
+                                    "key": "Docker.NetworkSettings.IPAddress",
+                                    "value": "10.3.0.4"
+                                }
+                            ]
+                        },
+                        {
+                            "state": "TASK_RUNNING",
+                            "timestamp": 1413572390.09275,
+                            "labels": [
+                                {
+                                    "key": "foo",
+                                    "value": "bar"
+                                },
+                                {
+                                    "key": "Docker.NetworkSettings.IPAddress",
+                                    "value": "10.3.0.1"
+                                }
+                            ]
+                        },
+                        {
+                            "state": "TASK_RUNNING",
+                            "timestamp": 1413572381.0123,
+                            "labels": [
+                                {
+                                    "key": "foo",
+                                    "value": "bar"
+                                },
+                                {
+                                    "key": "container-ip",
+                                    "value": ""
+                                }
+                            ]
+                        },
+                        {
+                            "state": "TASK_STARTING",
+                            "timestamp": 1413572380.01234
                         }
                     ]
                 },
@@ -226,7 +274,17 @@
                     "statuses": [
                         {
                             "state": "TASK_RUNNING",
-                            "timestamp": 1413572383.04559
+                            "timestamp": 1413572383.04559,
+                            "labels": [
+                                {
+                                    "key": "foo",
+                                    "value": "bar"
+                                },
+                                {
+                                    "key": "Docker.NetworkSettings.IPAddress",
+                                    "value": "10.3.0.2"
+                                }
+                            ]
                         }
                     ]
                 },
--- a/records/generator.go
+++ b/records/generator.go
@@ -36,19 +36,33 @@ type Resources struct {
 	Ports string `json:"ports"`
 }
 
+// Label holds a key/value pair
+type Label struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+// Status holds a task status
+type Status struct {
+	Timestamp float64 `json:"timestamp"`
+	State     string  `json:"state"`
+	Labels    []Label `json:"labels,omitempty"`
+}
+
 // Tasks holds mesos task information read in from state.json
-type Tasks []struct {
-	FrameworkId string `json:"framework_id"`
-	Id          string `json:"id"`
-	Name        string `json:"name"`
-	SlaveId     string `json:"slave_id"`
-	State       string `json:"state"`
+type Task struct {
+	FrameworkId string   `json:"framework_id"`
+	Id          string   `json:"id"`
+	Name        string   `json:"name"`
+	SlaveId     string   `json:"slave_id"`
+	State       string   `json:"state"`
+	Statuses    []Status `json:"statuses"`
 	Resources   `json:"resources"`
 }
 
 // Frameworks holds mesos frameworks information read in from state.json
 type Frameworks []struct {
-	Tasks `json:"tasks"`
+	Tasks []Task `json:"tasks"`
 	Name  string `json:"name"`
 }
 
@@ -231,6 +245,30 @@ func sanitizedSlaveAddress(hostname string, spec labels.HostNameSpec) string {
 	return address
 }
 
+func (t *Task) containerIP() string {
+	const containerIPTaskStatusLabel = "Docker.NetworkSettings.IPAddress"
+
+	// find TASK_RUNNING statuses
+	var latestContainerIP string
+	var latestTimestamp float64
+	for _, status := range t.Statuses {
+		if status.State != "TASK_RUNNING" {
+			continue
+		}
+
+		// find the latest docker-inspect label
+		for _, label := range status.Labels {
+			if label.Key == containerIPTaskStatusLabel && status.Timestamp > latestTimestamp {
+				latestContainerIP = label.Value
+				latestTimestamp = status.Timestamp
+				break
+			}
+		}
+	}
+
+	return latestContainerIP
+}
+
 // InsertState transforms a StateJSON into RecordGenerator RRs
 func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
 	listener string, masters []string, spec labels.HostNameSpec) error {
@@ -248,7 +286,7 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
 	for _, f := range sj.Frameworks {
 		fname := labels.AsDomainFrag(f.Name, spec)
 		for _, task := range f.Tasks {
-			host, ok := rg.Slaves[task.SlaveId]
+			hostIP, ok := rg.SlaveIPs[task.SlaveId]
 			// skip not running or not discoverable tasks
 			if !ok || (task.State != "TASK_RUNNING") {
 				continue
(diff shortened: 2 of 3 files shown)


records/config.go
```diff
@@ -71,6 +71,9 @@ type Config struct {
 
 	// EnforceRFC952 will enforce an older, more strict set of rules for DNS labels
 	EnforceRFC952 bool
+
+	// parse docker inspect output for container ip
```
tsenart: Comment should begin with the described variable.

records/generator.go
```diff
@@ -36,13 +37,21 @@ type Resources struct {
 	Ports string `json:"ports"`
 }
 
+// Status hold a task status
```
tsenart: s/hold/represents/

records/generator.go
```diff
@@ -36,13 +37,21 @@ type Resources struct {
...
+}
+
 // Tasks holds mesos task information read in from state.json
 type Tasks []struct {
-	FrameworkId string `json:"framework_id"`
-	Id          string `json:"id"`
-	Name        string `json:"name"`
-	SlaveId     string `json:"slave_id"`
-	State       string `json:"state"`
+	FrameworkId string   `json:"framework_id"`
```
tsenart: s/Id/ID/

records/generator.go
```diff
@@ -36,13 +37,21 @@ type Resources struct {
...
+	Id          string   `json:"id"`
+	Name        string   `json:"name"`
+	SlaveId     string   `json:"slave_id"`
```
tsenart: s/Id/ID/

records/generator.go
```diff
@@ -66,6 +75,16 @@ type StateJSON struct {
 	Leader     string `json:"leader"`
 }
 
+// DockerNetworkSettings partially holds the docker container network settings
+type DockerNetworkSettings struct {
+	IPAddress string `json"IPAddress"`
```
tsenart: Typo in the field tag. Also, we're using snake casing for all other json key names. We should be consistent.

records/generator.go
sttts: not part of this commit

records/generator.go
```diff
@@ -248,12 +268,38 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
...
+						}
+
+						if len(inspectResult) > 0 {
+							containerIP = inspectResult[0].NetworkSettings.IPAddress
+						}
+						break
+					}
+				}
+			}
+
```
tsenart: Can we extract this to a private method? `InsertState` is getting quite long.

records/generator.go
```diff
@@ -248,12 +268,38 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
...
 			// skip not running or not discoverable tasks
 			if !ok || (task.State != "TASK_RUNNING") {
 				continue
 			}
 
+			// parse docker inspect output for container IP if enabled
+			containerIP := ""
+			if useDockerContainerIP {
+				for _, status := range task.Statuses {
+					if status.State == "TASK_RUNNING" && len(status.Data) > 0 {
```
tsenart: Reduce nesting by inverting the condition and continuing.

records/generator.go
```diff
@@ -248,12 +268,38 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
...
+						err = json.Unmarshal(decoded, &inspectResult)
+						if err != nil {
+							logging.VeryVerbose.Printf("error parsing docker inspect output of task %v: %v", task.Id, err)
+							continue
+						}
+
+						if len(inspectResult) > 0 {
+							containerIP = inspectResult[0].NetworkSettings.IPAddress
+						}
+						break
```
tsenart: I would think that you'd only want to break if the containerIP was actually found. Here we're breaking anyways.

records/generator_test.go
```diff
@@ -256,10 +255,18 @@ func TestInsertState(t *testing.T) {
 		t.Error("should not find this not-running task - SRV record")
 	}
 
-	_, ok = rg.As["liquor-store.marathon.mesos."]
+	As, ok := rg.As["liquor-store.marathon.mesos."]
```
tsenart: It's kind of weird to use capitalized names for local variables.

records/generator_test.go
```diff
@@ -256,10 +255,18 @@ func TestInsertState(t *testing.T) {
...
 	if !ok {
 		t.Error("should find this running task - A record")
 	}
+	if len(As) != 2 || As[0] != "1.2.3.11" || As[1] != "1.2.3.12" {
```
tsenart: Easier: `if got, want := As, []string{"1.2.3.11", "1.2.3.12"}; !reflect.DeepEqual(got, want) {`

records/generator_test.go
sttts: yes, but here it is A like A-record. kind of makes sense

tsenart: I'd just call it `rrs` for resource records.

tsenart:
@sttts: PTAL

records/generator_test.go
sttts: there are also SRV records

records/generator_test.go
sttts: well, easier can be discussed.

sttts: in fact I would prefer a set equal

tsenart: Set equal?

records/generator_test.go
sttts: ips makes sense

sttts: but rrs is used elsewhere already

records/generator_test.go
sttts: like set(list) = set(list2)

records/generator.go
```diff
@@ -231,14 +250,42 @@ func sanitizedSlaveAddress(hostname string, spec labels.HostNameSpec) string {
 	return address
 }
 
+func (task *Task) containerIP() string {
+	for _, status := range task.Statuses {
+		if status.State != "TASK_RUNNING" || len(status.Data) == 0 {
+			continue
+		}
+
+		decoded, err := base64.StdEncoding.DecodeString(status.Data)
```
tsenart:
```go
var data []DockerInspect
dec := json.NewDecoder(base64.NewDecoder(base64.StdEncoding, bytes.NewReader(status.Data))
if err := dec.Decode(data); err != nil {
    // log error
} else if len(data) > 0 {
    return data[0].NetworkSettings.IpAddress
}
```

```

Comment (sttts) [inline records/generator.go]: Why not using the StdEncoding.DecodeString?

File: records/generator_test.go [hunk 12]
```diff
@@ -317,6 +324,48 @@ func TestInsertState(t *testing.T) {
...
+	}
+	if got, want := rrs, []string{"1.2.3.11", "1.2.3.12"}; !reflect.DeepEqual(got, want) {
+		t.Errorf("should return the slave ips 1.2.3.11, 1.2.3.12 for the task, but got %v", rrs)
+	}
+
+	rrs, ok = rg.As["_container.liquor-store.marathon.mesos."]
+	if !ok {
+		t.Error("should find the container ip because ParseDockerInspectForContainerIP is enabled")
+	}
+	if len(rrs) != 1 || rrs[0] != "10.3.0.1" {
```
tsenart: If we're now using `reflect.DeepEqual` in the previous case, we should use it in the others too.

records/generator.go
tsenart: Composition FTW!

factories/fake.json
```diff
@@ -206,7 +212,49 @@
...
-                            "timestamp": 1413572390.09275
+                            "timestamp": 1413572371.04321,
+                            "labels": [
+                                {
+                                    "key": "foo",
+                                    "value": "bar"
+                                },
+                                {
+                                    "key": "docker_inspect",
+                                    "value": "[\n{\n \"NetworkSettings\": {\n \"GlobalIPv6PrefixLen\": 0,\n \"HairpinMode\": false,\n \"IPAddress\": \"10.3.0.4\" }\n}\n]\n"
```
tsenart: Are the newlines inside the JSON needed?

records/generator.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/mesosphere/mesos-dns/records/labels"
 )
 
+const (
+	DockerInspectLabel = "docker_inspect"
```
tsenart: Please add a comment describing this constant.

factories/fake.json
sttts: It's a real extract of a state.json.

records/generator.go
sttts: done.

records/generator.go
```diff
@@ -66,6 +84,16 @@ type StateJSON struct {
...
+// DockerNetworkSettings is a partial representation of the docker container network settings
+type DockerNetworkSettings struct {
+	IpAddress string `json:"IpAddress"`
+}
+
+// DockerInspect is a partial representation of a docker inspect output for a task's container
+type DockerInspect struct {
+	NetworkSettings DockerNetworkSettings `json:"NetworkSettings"`
+}
+
```
tsenart: We probably don't need to export these types. It'd be enough to declare them only in the scope they're required, inside `containerIPFromDockerInspect`.
```go
func containerIPFromDockerInspect(inspect string) (string, error) {
    var data []struct {
        Settings struct {
            IP string `json:"IpAddress"`
        } `json:"NetworkSettings"`
    }
    // ...
}
```

records/generator.go
```diff
@@ -231,14 +252,66 @@ func sanitizedSlaveAddress(hostname string, spec labels.HostNameSpec) string {
...
 }
 
+func containerIPFromDockerInspect(inspect string) (string, error) {
+	var inspectJSON []struct {
+		NetworkSettings struct {
+			IpAddress string `json:"IpAddress"`
+		} `json:"NetworkSettings"`
+	}
+
+	err := json.Unmarshal([]byte(inspect), &inspectJSON)
```
tsenart: `inspectJSON` is a reference type so no need to pass a reference in to `json.Unmarshal`

tsenart:
LGTM

tsenart:
@karya0: PTAL

records/generator.go
```diff
@@ -17,6 +17,12 @@ import (
 	"github.com/mesosphere/mesos-dns/records/labels"
 )
 
+const (
+	// containerIPTaskStatusLabel represents the key of the state.json task status label
+	// which holds the container IP.
+	containerIPTaskStatusLabel = "Docker.NetworkSettings.IPAddress"
```
tsenart: This isn't used anywhere else than the `containerIP` method. We should define it inside and get rid of the comment since its not necessary so close to where it's used.

PR #188, opened 2015-07-20
Don't silently ignore user.Current error.

alberts:
You can hit this if you build a Go without cgo.


records/config.go
```diff
@@ -97,7 +97,10 @@ func SetConfig(cjson string) (c Config) {
 	}
 
 	// read configuration file
-	usr, _ := user.Current()
+	usr, err := user.Current()
+	if err != nil {
+		logging.Error.Fatalf("user.Current: %v", err)
```
tsenart: Please log a message that doesn't leak implementation details. Something like: `failed to retrieve OS user: %s`

jdef:
this was resolved by #227

PR #192, opened 2015-07-24
Round one of fixing `gometalinter` warnings

tsenart:


records/config.go
```diff
@@ -174,8 +175,8 @@ func SetConfig(cjson string) (c Config) {
 	logging.Verbose.Println("   - SOAExpire: ", c.SOAExpire)
 	logging.Verbose.Println("   - SOAExpire: ", c.SOAMinttl)
 	logging.Verbose.Println("   - RecurseOn: ", c.RecurseOn)
-	logging.Verbose.Println("   - HttpPort: ", c.HttpPort)
```
sttts: Shouldn't we print out the JSON spelling, i.e. HttpOn?

resolver/resolver.go
```diff
@@ -581,69 +577,54 @@ func (res *Resolver) RestHost(req *restful.Request, resp *restful.Response) {
...
-		h, port, _ := net.SplitHostPort(srv)
-		p, _ := strconv.Atoi(port)
+	type service struct {
+		Service string `json:"service"`
+		Host    string `json:"host"`
+		IP      string `json:"ip"`
+		Port    string `json:"port"`
+	}
+
+	services := make([]service, 0, len(rs.SRVs[dom]))
```
sttts: Is the go compiler clever enough to optimize multiple equal lookups in rs.SRVs?

resolver/resolver.go
```diff
@@ -519,57 +517,55 @@ func (res *Resolver) LaunchHTTP() <-chan error {
...
 		logging.Error.Println(err)
 	}
-	io.WriteString(resp, string(output))
 
-	// stats
-	mesosrq := strings.HasSuffix(dom, res.config.Domain+".")
+	stats(strings.HasSuffix(dom, res.config.Domain+"."), len(rs.As[dom]) == 0)
+}
+
+func stats(mesosrq, empty bool) {
```
sttts: `empty` is a bad name, in a generic function `stats`. What is empty?

sttts: Why not `mesosRq` or even `mesosRequest`?

records/config.go
tsenart: Maybe. But I didn't think it mattered in the case of the logs. But perhaps people are integrating with them. So, I'll change this.

resolver/resolver.go
tsenart: I took "refactorism" here a little too far perhaps and maintained the previous names, even though they weren't the best. In fact, all I wanted was to remove the code duplication and fix the warning, not properly redesign the code which will be the same thing to do in the long run. Anyways, I'll make it a little better and give more descriptive names to these arguments.

resolver/resolver.go
tsenart: I don't know, but I'm not worried about that here at all. Do you think I should?

sttts: It's unnecessary work. Maybe here it does not matter much and moreover the cpu cache is pre-filled with the right data of the maps after the first access. But still it has to climb down in the btree or whatever backs the maps.

tsenart: Now that you sparkled my thought on this, I'm even wondering about the lack of concurrency control around this shared state.

sttts: I guess we only read concurrently. Hopefully the generator is decoupled from this.

tsenart: Yes, I looked into it and the whole state is entirely replaced periodically. Everything is protected by a lock. Another way would be to do something like this: http://golang.org/pkg/sync/atomic/#example_Value_config

records/config.go
tsenart: Done.

tsenart:
@sttts: Addressed your comments. PTAL

resolver/resolver.go
```diff
@@ -519,60 +517,59 @@ func (res *Resolver) LaunchHTTP() <-chan error {
...
-	mesosrq := strings.HasSuffix(dom, res.config.Domain+".")
-	if mesosrq {
+	stats(dom, res.config.Domain+".", len(aRRs) > 0)
+}
+
+func stats(domain, zone string, success bool) {
+	if strings.HasSuffix(domain, zone) {
 		logging.CurLog.MesosRequests.Inc()
-		if empty {
+		if !success {
```
sttts: Small nit: turn the cases around, makes it simpler.

sttts:
Small nit. Otherwise LGTM

PR #193, opened 2015-07-27
Fix CircleCI build (part 2)

tsenart:
This change set brings us a little closer to have a green CircleCI build. After this is merged, the remaining failures are very few:
```
records/generator_test.go:234::warning: cyclomatic complexity 19 of function TestInsertState() is high (> 12) (gocyclo)
main.go:19::warning: cyclomatic complexity 14 of function main() is high (> 12) (gocyclo)
resolver/resolver_test.go:193::warning: cyclomatic complexity 21 of function TestHandler() is high (> 12) (gocyclo)
resolver/resolver_test.go:336::warning: cyclomatic complexity 17 of function TestHTTP() is high (> 12) (gocyclo)
records/generator.go:167:8:warning: error return value not checked (defer resp.Body.Close()) (errcheck)
```

EDIT: I reduced the scope of this PR to *NOT* include the `label` package refactoring. That will follow up in a different PR. This way we can get this code in right away I hope.

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -11,29 +11,40 @@ import (
 )
 
 var (
-	VerboseFlag     bool
+	// VerboseFlag enables verbose logging if set to true.
+	VerboseFlag bool
+	// VeryVerboseFlag enables very verbose logging if set to true.
 	VeryVerboseFlag bool
-	Verbose         *log.Logger
-	VeryVerbose     *log.Logger
-	Error           *log.Logger
+	// Verbose is this package's verbose Logger.
+	Verbose *log.Logger
+	// VeryVerbose is this package's very verbose Logger.
+	VeryVerbose *log.Logger
+	// Error is this package's error Logger.
+	Error *log.Logger
 )
 
+// Counter defines an interface for a monotonically incrementing value.
 type Counter interface {
 	Inc()
 }
 
+// LogCounter implements the Counter interface with a uint64 register.
+// It's safe for concurrent use.
 type LogCounter struct {
 	value uint64
 }
 
+// Inc increments the counter by one.
 func (lc *LogCounter) Inc() {
 	atomic.AddUint64(&lc.value, 1)
 }
 
+// String returns a string represention of the counter.
 func (lc *LogCounter) String() string {
 	return strconv.FormatUint(atomic.LoadUint64(&lc.value), 10)
 }
 
+// LogOut holds metrics captured in an instrumented runtime.
 type LogOut struct {
 	MesosRequests    Counter
 	MesosSuccess     Counter
@@ -46,6 +57,7 @@ type LogOut struct {
 	NonMesosRecursed Counter
 }
 
+// CurLog is the default package level LogOut.
 var CurLog = LogOut{
 	MesosRequests:    &LogCounter{},
 	MesosSuccess:     &LogCounter{},
--- a/records/generator.go
+++ b/records/generator.go
@@ -1,4 +1,4 @@
-// package records contains functions to generate resource records from
+// Package records contains functions to generate resource records from
 // mesos master states to serve through a dns server
 package records
 
@@ -22,34 +22,33 @@ import (
 // Will likely become map[string][]discoveryinfo
 type rrs map[string][]string
 
-// Mesos-DNS state
-// Refactor when discovery id is available
+// RecordGenerator contains DNS records and methods to access and manipulate
+// them. TODO(kozyraki): Refactor when discovery id is available.
 type RecordGenerator struct {
 	As       rrs
 	SRVs     rrs
 	SlaveIPs map[string]string
 }
 
-// The following types help parse state.json
-// Resources holds our SRV ports
+// Resources holds resources as defined in the /state.json Mesos HTTP endpoint.
 type Resources struct {
 	Ports string `json:"ports"`
 }
 
-// Label holds a key/value pair
+// Label holds a label as defined in the /state.json Mesos HTTP endpoint.
 type Label struct {
 	Key   string `json:"key"`
 	Value string `json:"value"`
 }
 
-// Status holds a task status
+// Status holds a task status as defined in the /state.json Mesos HTTP endpoint.
 type Status struct {
 	Timestamp float64 `json:"timestamp"`
 	State     string  `json:"state"`
 	Labels    []Label `json:"labels,omitempty"`
 }
 
-// Tasks holds mesos task information read in from state.json
+// Task holds a task as defined in the /state.json Mesos HTTP endpoint.
 type Task struct {
 	FrameworkID string   `json:"framework_id"`
 	ID          string   `json:"id"`
@@ -60,27 +59,27 @@ type Task struct {
 	Resources   `json:"resources"`
 }
 
-// Frameworks holds mesos frameworks information read in from state.json
-type Frameworks []struct {
+// Framework holds a framework as defined in the /state.json Mesos HTTP endpoint.
+type Framework struct {
 	Tasks []Task `json:"tasks"`
 	Name  string `json:"name"`
 }
 
-// Slaves is a mapping of id to hostname read in from state.json
-type slave struct {
+// Framework holds a framework as defined in the /state.json Mesos HTTP endpoint.
+type Slave struct {
 	ID       string `json:"id"`
 	Hostname string `json:"hostname"`
 }
-type Slaves []slave
 
-// StateJSON is a representation of mesos master state.json
+// StateJSON holds the state defined in the /state.json Mesos HTTP endpoint.
 type StateJSON struct {
-	Frameworks `json:"frameworks"`
-	Slaves     `json:"slaves"`
-	Leader     string `json:"leader"`
+	Frameworks []Framework `json:"frameworks"`
+	Slaves     []Slave     `json:"slaves"`
+	Leader     string      `json:"leader"`
 }
 
-// Finds the master and inserts DNS state
+// ParseState retrieves and parses the Mesos master /state.json and converts it
+// into DNS records.
 func (rg *RecordGenerator) ParseState(leader string, c Config) error {
 
 	// find master -- return if error
@@ -274,7 +273,6 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
 		}
 		rg.SlaveIPs[slave.ID] = address
 	}
-
 
 	// complete crap - refactor me
 	for _, f := range sj.Frameworks {
--- a/records/validation_test.go
+++ b/records/validation_test.go
@@ -5,10 +5,7 @@ import (
 )
 
 func TestValidateMasters(t *testing.T) {
-	table := []struct {
-		ms    []string
-		valid bool
-	}{
+	for i, tc := range []validationTest{
 		{nil, true},
 		{[]string{}, true},
 		{[]string{""}, false},
@@ -27,24 +24,13 @@ func TestValidateMasters(t *testing.T) {
 		{[]string{"[2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b]:1"}, true},
 		{[]string{"[2001:db8:3c4d:15::1a2f:1a2b]:1"}, true},
 		{[]string{"[2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b]:1", "[2001:db8:3c4d:15::1a2f:1a2b]:1"}, false},
-	}
-	for i, tc := range table {
-		err := validateMasters(tc.ms)
-		if (err == nil && tc.valid) || (err != nil && !tc.valid) {
-			continue
-		} else if tc.valid {
-			t.Fatalf("test %d failed, unexpected error validating masters %v: %v", i+1, tc.ms, err)
-		} else {
-			t.Fatalf("test %d failed, expected validation error for masters(%d) %v", i+1, len(tc.ms), tc.ms)
-		}
+	} {
+		validate(t, i+1, tc, validateMasters)
 	}
 }
 
 func TestValidateResolvers(t *testing.T) {
-	table := []struct {
-		rs    []string
-		valid bool
-	}{
+	for i, tc := range []validationTest{
 		{nil, true},
 		{[]string{}, true},
 		{[]string{""}, false},
@@ -58,15 +44,22 @@ func TestValidateResolvers(t *testing.T) {
 		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b"}, true},
 		{[]string{"2001:db8:3c4d:15::1a2f:1a2b"}, true},
 		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b", "2001:db8:3c4d:15::1a2f:1a2b"}, false},
+	} {
+		validate(t, i+1, tc, validateResolvers)
 	}
-	for i, tc := range table {
-		err := validateResolvers(tc.rs)
-		if (err == nil && tc.valid) || (err != nil && !tc.valid) {
-			continue
-		} else if tc.valid {
-			t.Fatalf("test %d failed, unexpected error validating resolvers %v: %v", i+1, tc.rs, err)
-		} else {
-			t.Fatalf("test %d failed, expected validation error for resolvers(%d) %v", i+1, len(tc.rs), tc.rs)
-		}
+}
+
+type validationTest struct {
+	in    []string
+	valid bool
+}
+
+func validate(t *testing.T, i int, tc validationTest, f func([]string) error) {
+	if err := f(tc.in); (err == nil && tc.valid) || (err != nil && !tc.valid) {
+		return // valid
+	} else if tc.valid {
+		t.Fatalf("test %d failed, unexpected error validating resolvers %v: %v", i, tc.in, err)
+	} else {
+		t.Fatalf("test %d failed, expected validation error for resolvers(%d) %v", i, len(tc.in), tc.in)
 	}
 }
--- a/records/validation_test.go
+++ b/records/validation_test.go
@@ -55,11 +55,12 @@ type validationTest struct {
 }
 
 func validate(t *testing.T, i int, tc validationTest, f func([]string) error) {
-	if err := f(tc.in); (err == nil && tc.valid) || (err != nil && !tc.valid) {
+	switch err := f(tc.in); {
+	case (err == nil && tc.valid) || (err != nil && !tc.valid):
 		return // valid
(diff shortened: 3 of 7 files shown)


tsenart:
PTAL @sttts @jdef

jdef:
i'm curious about the benchmarking/performance of the labels refactoring. The prior table-based lookup approach really lowered label generation times (performance was key concern of @kozyraki). The new implementation looks more involved so I'm wondering what that does to label generation times.

circle.yml
```diff
@@ -9,5 +9,5 @@ dependencies:
 test:
     override:
         - ! gofmt -s -d . 2>&1 | read
-        - gometalinter ./...
+        - gometalinter --cyclo-over=12 ./...
```
jdef: how did we pick 12? why not 13, or 15?

resolver/resolver.go
```diff
@@ -167,40 +168,28 @@ func (res *Resolver) Reload() {
...
-	var in *dns.Msg
-	var err error
-
-	c := new(dns.Client)
-	c.Net = proto
+func (res *Resolver) defaultExtResolver(r *dns.Msg, nameserver string, proto string, cnt int) (in *dns.Msg, err error) {
+	defer logging.CurLog.NonMesosRecursed.Inc()
 
-	var t time.Duration = 5 * 1e9
+	timeout := 5 * time.Second
```
jdef: magic number: can we extract a constant for this?

resolver/resolver.go
```diff
@@ -354,111 +344,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
 // question with resource answer(s)
 // it can handle {A, SRV, ANY}
 func (res *Resolver) HandleMesos(w dns.ResponseWriter, r *dns.Msg) {
-	var err error
+	defer logging.CurLog.MesosRequests.Inc()
```
jdef: :+1: :+1: for refactoring this func

circle.yml
tsenart: The lower, the better. But at some point we cross a threshold that doesn't make sense anymore. I found this value through quite extensive experimentation.

tsenart:
I can run the benchmarks tomorrow for the two versions but I'd like to see valid data backed evidence of what real performance requirements are for label generation. I'd be very surprised if it would be a bottleneck and I'd optimise for maintainability, composability and modularity first.

resolver/resolver.go
tsenart: This one was the bigger beast. :-)

resolver/resolver.go
tsenart: Sure, but in a different commit, since this was a pure refactoring.

logging/logging.go
```diff
@@ -11,29 +11,40 @@ import (
...
-	Error           *log.Logger
+	// Verbose is this package's verbose Logger.
+	Verbose *log.Logger
+	// VeryVerbose is this package's very verbose Logger.
+	VeryVerbose *log.Logger
+	// Error is this package's error Logger.
+	Error *log.Logger
 )
 
+// Counter defines an interface for a monotonically incrementing counter.
```
sttts: sounds like a tautology. s/counter/integer/

records/generator.go
```diff
@@ -22,16 +22,15 @@ import (
 // Will likely become map[string][]discoveryinfo
 type rrs map[string][]string
 
-// Mesos-DNS state
-// Refactor when discovery id is available
+// RecordGenerator contains DNS records and methods to access and manipulate
+// them. TODO: Refactor when discovery id is available.
```
sttts: TODO(tsenart), I think we also assign TODOs.

records/generator.go
```diff
@@ -22,16 +22,15 @@ import (
...
 type RecordGenerator struct {
 	As       rrs
 	SRVs     rrs
 	SlaveIPs map[string]string
 }
 
-// The following types help parse state.json
-// Resources holds our SRV ports
+// Resources holds SRV ports.
 type Resources struct {
```
sttts: In fact it's not nice that these types are global. They are only used in this package, but I figure they must be public for the json conversion to work.

records/generator.go
```diff
@@ -22,16 +22,15 @@ import (
...
+// them. TODO: Refactor when discovery id is available.
 type RecordGenerator struct {
 	As       rrs
 	SRVs     rrs
 	SlaveIPs map[string]string
 }
 
-// The following types help parse state.json
-// Resources holds our SRV ports
+// Resources holds SRV ports.
```
sttts: In fact this is not about SRV ports in the DNS sense. It's just for the state.json structures from Mesos master.

records/generator.go
```diff
@@ -49,7 +48,7 @@ type Status struct {
 	Labels    []Label `json:"labels,omitempty"`
 }
 
-// Tasks holds mesos task information read in from state.json
+// Task holds mesos task information read from state.json
```
sttts: Put state.json in every of these comments.

records/generator.go
```diff
@@ -66,11 +65,12 @@ type Frameworks []struct {
...
 }
 
-// Slaves is a mapping of id to hostname read in from state.json
 type slave struct {
 	ID       string `json:"id"`
 	Hostname string `json:"hostname"`
 }
+
+// Slaves is a mapping of id to hostname read in from state.json
 type Slaves []slave
```
sttts: It's not a mapping, it's just a list.

records/generator.go
```diff
@@ -80,7 +80,8 @@ type StateJSON struct {
 	Leader     string `json:"leader"`
 }
 
-// Finds the master and inserts DNS state
+// ParseState finds the Mesos master and retrieves the state required to
```
sttts: It retrieves the state **and** processes it with insertState

records/validation_test.go
```diff
@@ -58,15 +44,22 @@ func TestValidateResolvers(t *testing.T) {
...
-		if (err == nil && tc.valid) || (err != nil && !tc.valid) {
-			continue
-		} else if tc.valid {
-			t.Fatalf("test %d failed, unexpected error validating resolvers %v: %v", i+1, tc.rs, err)
-		} else {
-			t.Fatalf("test %d failed, expected validation error for resolvers(%d) %v", i+1, len(tc.rs), tc.rs)
-		}
+}
+
+type validationTest struct {
```
sttts: I tend to put those utilities on-top. But maybe I am just used to those ancient declare-before-use languages.

records/validation_test.go
```diff
@@ -58,15 +44,22 @@ func TestValidateResolvers(t *testing.T) {
...
+	in    []string
+	valid bool
+}
+
+func validate(t *testing.T, i int, tc validationTest, f func([]string) error) {
+	if err := f(tc.in); (err == nil && tc.valid) || (err != nil && !tc.valid) {
```
sttts: that's a long if. Better split it. Makes it more readable.

records/validation_test.go
```diff
@@ -58,15 +44,22 @@ func TestValidateResolvers(t *testing.T) {
...
+		return // valid
+	} else if tc.valid {
```
sttts: Confusing logic: tc.valid => fatal error.

resolver/resolver.go
```diff
@@ -16,7 +15,7 @@ import (
 
 	"github.com/emicklei/go-restful"
 	"github.com/mesos/mesos-go/detector"
-	_ "github.com/mesos/mesos-go/detector/zoo"
+	_ "github.com/mesos/mesos-go/detector/zoo" // Registers the ZK detector
```
sttts: do we use capital letters for all comments?

resolver/resolver.go
```diff
@@ -62,7 +62,8 @@ func (res *Resolver) records() *records.RecordGenerator {
 	return res.rs
 }
 
-// launches DNS server for a resolver, returns immediately
+// LaunchDNS starts a DNS server for the Resolver, returning a error channel
```
sttts: udp and/or tcp? Should be mentioned.

resolver/resolver.go
```diff
@@ -77,7 +78,7 @@ func (res *Resolver) LaunchDNS() <-chan error {
 	return errCh
 }
 
-// starts a DNS server for net protocol (tcp/udp), returns immediately.
+// Serve starts a DNS server for net protocol (tcp/udp), returns immediately.
```
sttts: launchDNS and serve "start a DNS server". Where is the difference?

resolver/resolver.go
```diff
@@ -165,42 +166,33 @@ func (res *Resolver) Reload() {
 	}
 }
 
+// externalTimeout is the default external resolver query timeout.
```
sttts: externalQueryTimeout would be a better name then.

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
+			logging.CurLog.MesosFailed.Inc()
+			logging.Error.Println(err)
+			continue
 		}
-	}
-
-	// SOA requests
-	if (qType == dns.TypeSOA) || (qType == dns.TypeANY) {
-		rr, err := res.formatSOA(r.Question[0].Name)
+		m.Answer = append(m.Answer, srvRR)
```
sttts: I like to put some sensible blank lines, separating logical blocks of code.

resolver/resolver_test.go
```diff
@@ -289,8 +289,8 @@ func TestHandler(t *testing.T) {
 		t.Error(err)
 	}
 
-	if m.Rcode != 3 {
-		t.Error("not setting NXDOMAIN for AAAA requests")
+	if got, want := m.Rcode, dns.RcodeNameError; got != want {
+		t.Errorf("not setting NXDOMAIN for AAAA requests: got rcode: %v, want: %v", got, want)
```
sttts: rcode or Rcode as above? same for got/want vs. Answer.

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
+		// return one corresponding A record add additional info
```
sttts: correct grammar?

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
+		host := strings.Split(srv, ":")[0]
+		if len(rs.As[host]) == 0 {
+			continue
+		}
+		aRR, err := res.formatA(host, rs.As[host][0])
 		if err != nil {
+			logging.CurLog.MesosFailed.Inc()
```
sttts: This changes the metric. In the old code the counter only counted the requests. Now it's iterating over the srv's.

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
 
-	// NS requests
-	if (qType == dns.TypeNS) || (qType == dns.TypeANY) {
-		rr, err := res.formatNS(r.Question[0].Name)
-		logging.VeryVerbose.Println("NS request")
+func (res *Resolver) handleA(rs *records.RecordGenerator, name string, m *dns.Msg) {
+	for _, a := range rs.As[name] {
+		rr, err := res.formatA(name, a)
 		if err != nil {
+			logging.CurLog.MesosFailed.Inc()
```
sttts: ditto.

resolver/resolver.go
sttts: No real need for the defer

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
-					}
-				}
-
-			}
-		}
+	name := strings.ToLower(cleanWild(r.Question[0].Name))
+	switch r.Question[0].Qtype {
+	case dns.TypeSRV, dns.TypeANY:
+		res.handleSRV(rs, name, m, r)
+	case dns.TypeA:
```
sttts: dns.TypeANY missing.

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
+		res.handleA(rs, name, m)
+	case dns.TypeSOA:
```
sttts: ditto

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
+		res.handleSOA(m, r)
+	case dns.TypeNS:
```
sttts: ditto

logging/logging.go
jdef: users of the Counter API aren't meant to read the value. Counter values are rendered periodically in the logs, and so LogCounter.String() is invoked to display the value.

tsenart: I'll `s/counter/value/` to remove the tautology.

records/generator.go
tsenart: Yeah, indeed. Although the TODO was already there, without the explicit TODO :)

records/generator.go
tsenart: They don't need to be public for unmarshmaling. Only the keys need to be public.

records/generator.go
tsenart: :D Outdated comments are outdated.

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
 
-			logging.CurLog.MesosNXDomain.Inc()
-			logging.VeryVerbose.Println("total A rrs:\t" + strconv.Itoa(len(rs.As)))
-			logging.VeryVerbose.Println("failed looking for " + r.Question[0].String())
-		} else {
-			logging.CurLog.MesosSuccess.Inc()
-		}
+	m.Rcode = dns.RcodeNameError
+	if qType == dns.TypeAAAA && len(rs.SRVs[name])+len(rs.As[name]) > 0 {
+		m.Rcode = dns.RcodeSuccess
```
sttts: This code path is guarded now by `len(m.Answer)==0`. It wasn't before. Problem?

resolver/resolver.go
```diff
@@ -354,111 +347,117 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
-			if err != nil {
-				logging.Error.Println(err)
-			} else {
-				m.Ns = append(m.Ns, rr)
-			}
+func (res *Resolver) handleEmpty(rs *records.RecordGenerator, name string, m, r *dns.Msg) {
+	qType := r.Question[0].Qtype
+	switch qType {
+	case dns.TypeSOA, dns.TypeNS, dns.TypeSRV:
+		return
```
sttts: These three record types increase the MesosSuccess metric before, but don't do it now anymore. Intentional?

records/generator.go
sttts: right.

records/generator.go
sttts: oh, so git blame the code and put the name in there ;-)

resolver/resolver.go
tsenart: True.

records/generator.go
tsenart: For performance, we should really be using https://github.com/pquerna/ffjson

records/generator.go
tsenart: I was making all of these type private and realised they need to be public because we have public methods which accept them. See `InsertState` for instance.

records/validation_test.go
tsenart: It's idiomatic to have private declarations below public ones.

records/validation_test.go
tsenart: This is a pure refactoring with the intention to clear warnings in the CircleCI build. I didn't intend to clean things out unless it cleared a warning.

resolver/resolver.go
tsenart: We should start comments with a capital letter.

resolver/resolver.go
tsenart: LaunchDNS calls Serve for both TCP and UDP.

resolver/resolver.go
tsenart: That would never be reached, because if the type would be TypeANY, the first case would match.

resolver/resolver.go
tsenart: I'm not entirely sure what that comment is conveying.

resolver/resolver.go
tsenart: That's OK since `dns.RcodeSuccess` is the default value.

tsenart:
@sttts: PTAL again :-)

resolver/resolver.go
sttts: fine with that

resolver/resolver.go
sttts: looks good now

resolver/resolver.go
sttts: actually the comment is a bit strange: the import does not register anything.

resolver/resolver.go
```diff
@@ -354,111 +347,135 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
...
+	if len(m.Answer) == 0 {
+		err = res.handleEmpty(rs, name, m, r)
+	} else {
+		shuffleAnswers(m.Answer)
+		logging.CurLog.MesosSuccess.Inc()
+	}
 
-			}
-		}
+	if err != nil {
```
sttts: Shouldn't the error case take precedence over the len==0 case? Now a handler which throws an error will possibly also return an empty answer and then the error never shows up.

sttts:
A few comments, otherwise lgtm

resolver/resolver.go
tsenart: Yes it does: https://github.com/mesos/mesos-go/blob/master/detector/zoo/plugin.go#L7

Side effects... Side effects everywhere! :)

resolver/resolver.go
tsenart: Which is what was previously happening, unfortunately. But I'll change it.

resolver/resolver.go
sttts: oh, the world is crazy

sttts:
:+1:

PR #195, opened 2015-07-28
downgrade log level of debug message

jdef:
a customer was concerned about an ERROR message in the mesos-dns logs. since it's not really an error, we shouldn't log it that way. we can probably do away with this line eventually, but for now I'm simply downgrading it from "Error" to "VeryVerbose".


jdef:
@sttts @tsenart PTAL

discordianfish:
LGTM - and agree that it can probably be removed altogether.

tsenart:
I can say that this can be removed with 99.9% confidence.

PR #196, opened 2015-07-28
add A record slave.<domain> to contain list of slave IPs

lloesche:
This allows me to resolve slave.mesos and get a list of all slave IPs.
We'd like to use this for our dreamathon monitoring project which is supposed to run on DCOS and monitor the state of all mesos slaves.

In a second step I'd also like to add appropriate SRV records, but couldn't find an easy way to get the slave port out of the current data structure.

### DIFF
--- a/records/generator.go
+++ b/records/generator.go
@@ -271,12 +271,15 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
 
 	// creates a map with slave IP addresses (IPv4)
 	rg.SlaveIPs = make(map[string]string)
-	for _, slave := range sj.Slaves {
-		rg.SlaveIPs[slave.ID] = sanitizedSlaveAddress(slave.Hostname, spec)
-	}
-
 	rg.SRVs = make(rrs)
 	rg.As = make(rrs)
+
+	for _, slave := range sj.Slaves {
+		ssa := sanitizedSlaveAddress(slave.Hostname, spec)
+		rg.SlaveIPs[slave.ID] = ssa
+		rg.insertRR("slave."+domain+".", ssa, "A")
+	}
+
 
 	// complete crap - refactor me
 	for _, f := range sj.Frameworks {
--- a/records/generator_test.go
+++ b/records/generator_test.go
@@ -287,6 +287,14 @@ func TestInsertState(t *testing.T) {
 		t.Error("should find a running master - A record")
 	}
 
+	rrs, ok = rg.As["slave.mesos."]
+	if !ok {
+		t.Error("should find a running slave - A record")
+	}
+	if got, want := rrs, []string{"1.2.3.10", "1.2.3.11", "1.2.3.12"}; !reflect.DeepEqual(got, want) {
+		t.Errorf("should return the slave ips %v for the slave record, but got %v", want, got)
+	}
+
 	_, ok = rg.As["master0.mesos."]
 	if !ok {
 		t.Error("should find a running master0 - A record")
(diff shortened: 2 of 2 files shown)


jdef:
the port shows up in two places in the slave state.json: flags.port and in pid. it's probably pretty simple to add a "pid" field to the slave struct, and then parse the port from that. sample data for pid:
```
slave(1)@10.2.0.5:5051
```

you can use mesos-go/upid to parse if you want (mesos-go is already a dep):
https://github.com/mesos/mesos-go/blob/master/upid/upid.go

lloesche:
@jdef `go test ./...` now passes but the gometalinter throws lots of warnings that don't seem related to my patch?!

tsenart:
@lloesche: You can ignore those gometalinter failures for now.

Regarding your patch, it's great that you fixed the test but as far as I can see, the added functionality needs its own test.

lloesche:
@tsenart added. I'm learning so much about Go today :-)

tsenart:
It's fun isn't it?! :-)

tsenart:
I'll take another look at this once I get back home.

records/generator_test.go
```diff
@@ -287,6 +287,14 @@ func TestInsertState(t *testing.T) {
 		t.Error("should find a running master - A record")
 	}
 
+	rrs, ok = rg.As["slaves.mesos."]
+	if !ok {
+		t.Error("should find running slaves - A record")
+	}
+	if got, want := rrs, []string{"1.2.3.10", "1.2.3.11", "1.2.3.12"}; !reflect.DeepEqual(got, want) {
+		t.Errorf("should return the slave ips 1.2.3.10, 1.2.3.11, 1.2.3.12 for the slaves record, but got %v", rrs)
```
tsenart: Can we not hardcode the wanted value in this string?
```go
t.Errorf("got: %v, want: %v", got, want)
```

tsenart:
Apart from the small comment, LGTM. Please squash all commits into one once you're done.

lloesche:
Removed the hard coded IPs from the test (a pattern which I by the way copied from other tests :-)) and squashed everything into one commit. Also I updated the name to be singular slave.mesos instead of slaves.mesos to be consistent with master.mesos.

tsenart:
LGTM. @lloesche please open a new issue to track:
> In a second step I'd also like to add appropriate SRV records, but couldn't find an easy way to get the slave port out of the current data structure.

records/generator.go
```diff
@@ -271,12 +271,15 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
 
 	// creates a map with slave IP addresses (IPv4)
 	rg.SlaveIPs = make(map[string]string)
+	rg.SRVs = make(rrs)
+	rg.As = make(rrs)
+
 	for _, slave := range sj.Slaves {
-		rg.SlaveIPs[slave.ID] = sanitizedSlaveAddress(slave.Hostname, spec)
+		ssa := sanitizedSlaveAddress(slave.Hostname, spec)
```
jdef: actually there's a chance that `sanitizedSlaveAddress` doesn't return a string containing an ipv4 address, which will result in a bad A record. should probably test for that.

lloesche: @jdef does that affect rg.SlaveIPs[slave.ID] as well? or is it okay for rg.SlaveIPs to contain invalid IPs? I.e. should I only check for the part where I add an A record for the slave.domain name or also for the rg.SlaveIPs map?

lloesche: @jdef fixed in https://github.com/mesosphere/mesos-dns/pull/199

PR #197, opened 2015-07-28
Add container ip label MesosContainerizer.NetworkSettings.IPAddress and remove _container prefix

sttts:
This PR 
- adds the `MesosContainerizer.NetworkSettings.IPAddress` as alternative label to define container IPs
- and removes the _container prefix

If the container IP is passed via task status label,
- it is returned in the A record instead of the slave IPs
- no SRV records are created for that task (because the container ports are currently not available via state.json).

With DiscoveryInfo support in Mesos-DNS via https://github.com/mesosphere/mesos-dns/pull/147 and corresponding support for container IPs in frameworks like Marathon (it has to put the container ports into the DiscoveryInfo, not the host ports), SRV records can be added again (out of scope of this PR).

### DIFF
--- a/docs/docs/configuration-parameters.md
+++ b/docs/docs/configuration-parameters.md
@@ -28,7 +28,8 @@ The configuration file should include the following fields:
   "SOARefresh": 60,
   "SOARetry":   600,
   "SOAExpire":  86400,
-  "SOAMinttl": 60
+  "SOAMinttl": 60,
+  "ContainerIPsOn": true
 }
 ```
 
@@ -73,3 +74,6 @@ It is sufficient to specify just one of the `zk` or `masters` field. If both are
 `recurseon` controls if the DNS replies for names in the Mesos domain will indicate that recursion is available. The default value is `true`. 
 
 `enforceRFC952` will enforce an older, more strict set of rules for DNS labels. For details, see the [RFC-952](https://tools.ietf.org/html/rfc952). The default value is `false`.
+
+`ContainerIPsOn` controls whether A records with the container IP are created instead of the slave IP if one of task status labels `Docker.NetworkSettings.IPAddress` or `MesosContainerizer.NetworkSettings.IPAddress` is set by Mesos or the executor.
+The default value is `true`.
--- a/docs/docs/naming.md
+++ b/docs/docs/naming.md
@@ -26,11 +26,36 @@ $ dig search.marathon.mesos
 search.marathon.mesos.		60	IN	A	10.9.87.94
 ```
  
+## Container IP A Records
+
+The A records above point to the host IP of a given service. If container IPs are supported by Mesos and the executor of a test, Mesos-DNS will create another A record for each of the A records above by prefixing it with `_container` and returning the container IP:
+
+``` console
+$ dig _container.search.marathon.mesos
+
+; <<>> DiG 9.8.4-rpz2+rl005.12-P1 <<>> _container.search.marathon.mesos
+;; global options: +cmd
+;; Got answer:
+;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 24471
+;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 0
+
+;; QUESTION SECTION:
+;_container.search.marathon.mesos.         IN  A
+
+;; ANSWER SECTION:
+_container.search.marathon.mesos.      60  IN  A   10.0.3.72
+```
+
+*Note*: Container IPs must be provided by the executor of a task in one of the following task status labels:
+- `Docker.NetworkSettings.IPAddress`
+- `MesosContainerizer.NetworkSettings.IPAddress`.
+In general support for these will not be available before Mesos 0.24.
+
 ## SRV Records
 
 An SRV record associates a service name to a hostname and an IP port.  For task `task` launched by framework `framework`, Mesos-DNS generates an SRV record for service name `_task._protocol.framework.domain`, where `protocol` is `udp` or `tcp`. For example, other Mesos tasks can discover service `search` launched by the `marathon` framework with a lookup for lookup `_search._tcp.marathon.mesos`:
 
-``` console
+```console
 $ dig _search._tcp.marathon.mesos SRV
 
 ; <<>> DiG 9.8.4-rpz2+rl005.12-P1 <<>> _search._tcp.marathon.mesos SRV
@@ -44,7 +69,7 @@ $ dig _search._tcp.marathon.mesos SRV
 
 ;; ANSWER SECTION:
 _search._tcp.marathon.mesos.	60 IN SRV 0 0 31302 10.254.132.41.
-``` 
+```
 
 SRV records are generated only for tasks that have been allocated a specific port through Mesos. 
 
--- a/docs/docs/naming.md
+++ b/docs/docs/naming.md
@@ -8,7 +8,10 @@ Mesos-DNS defines a DNS domain for Mesos tasks (default `.mesos`, see [instructi
 
 ## A Records
 
-An A record associates a hostname to an IP address. For task `task` launched by framework `framework`, Mesos-DNS generates an A record for hostname `task.framework.domain` that provides the IP address of the specific slave running the task. For example, other Mesos tasks can discover the IP address for service `search` launched by the `marathon` framework with a lookup for `search.marathon.mesos`:
+An A record associates a hostname to an IP address. For task `task` launched by framework `framework`, Mesos-DNS generates an A record for hostname `task.framework.domain` that provides
+- either the IP address of the specific slave running the task
+- or the IP address of the task container itself if Mesos provides this (in the task status labels).
+For example, other Mesos tasks can discover the IP address for service `search` launched by the `marathon` framework with a lookup for `search.marathon.mesos`:
 
 ``` console
 $ dig search.marathon.mesos
@@ -25,25 +28,23 @@ $ dig search.marathon.mesos
 ;; ANSWER SECTION:
 search.marathon.mesos.		60	IN	A	10.9.87.94
 ```
- 
-## Container IP A Records
 
-The A records above point to the host IP of a given service. If container IPs are supported by Mesos and the executor of a test, Mesos-DNS will create another A record for each of the A records above by prefixing it with `_container` and returning the container IP:
+If Mesos is setup to provide a container IP `10.0.4.1` for the task `search.marathon.mesos` the lookup would give:
 
 ``` console
-$ dig _container.search.marathon.mesos
+$ dig search.marathon.mesos
 
-; <<>> DiG 9.8.4-rpz2+rl005.12-P1 <<>> _container.search.marathon.mesos
+; <<>> DiG 9.8.4-rpz2+rl005.12-P1 <<>> search.marathon.mesos
 ;; global options: +cmd
 ;; Got answer:
 ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 24471
 ;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 0
 
 ;; QUESTION SECTION:
-;_container.search.marathon.mesos.         IN  A
+;search.marathon.mesos.         IN  A
 
 ;; ANSWER SECTION:
-_container.search.marathon.mesos.      60  IN  A   10.0.3.72
+search.marathon.mesos.      60  IN  A   10.0.4.1
 ```
 
 *Note*: Container IPs must be provided by the executor of a task in one of the following task status labels:
@@ -71,7 +72,7 @@ $ dig _search._tcp.marathon.mesos SRV
 _search._tcp.marathon.mesos.	60 IN SRV 0 0 31302 10.254.132.41.
 ```
 
-SRV records are generated only for tasks that have been allocated a specific port through Mesos. 
+*Note*: SRV records are generated only for tasks that have been allocated a specific port through Mesos and that have no container IP assigned.
 
 ## Other Records
 
--- a/docs/docs/naming.md
+++ b/docs/docs/naming.md
@@ -9,8 +9,10 @@ Mesos-DNS defines a DNS domain for Mesos tasks (default `.mesos`, see [instructi
 ## A Records
 
 An A record associates a hostname to an IP address. For task `task` launched by framework `framework`, Mesos-DNS generates an A record for hostname `task.framework.domain` that provides
+
 - either the IP address of the specific slave running the task
 - or the IP address of the task container itself if Mesos provides this (in the task status labels).
+
 For example, other Mesos tasks can discover the IP address for service `search` launched by the `marathon` framework with a lookup for `search.marathon.mesos`:
 
 ``` console
@@ -48,8 +50,10 @@ search.marathon.mesos.      60  IN  A   10.0.4.1
 ```
 
 *Note*: Container IPs must be provided by the executor of a task in one of the following task status labels:
+
 - `Docker.NetworkSettings.IPAddress`
 - `MesosContainerizer.NetworkSettings.IPAddress`.
+
 In general support for these will not be available before Mesos 0.24.
 
 ## SRV Records
--- a/records/config.go
+++ b/records/config.go
@@ -72,6 +72,9 @@ type Config struct {
 
 	// EnforceRFC952 will enforce an older, more strict set of rules for DNS labels
 	EnforceRFC952 bool
+
+	// Enable serving container IP if the corresponding task status labels are found
+	ContainerIPsOn bool
 }
 
 // SetConfig instantiates a Config struct read in from config.json
@@ -95,6 +98,7 @@ func SetConfig(cjson string) (c Config) {
 		HTTPOn:         true,
 		ExternalOn:     true,
 		RecurseOn:      true,
+		ContainerIPsOn: true,
 	}
 
 	// read configuration file
@@ -179,6 +183,8 @@ func SetConfig(cjson string) (c Config) {
 	logging.Verbose.Println("   - HttpOn: ", c.HTTPOn)
 	logging.Verbose.Println("   - ConfigFile: ", c.File)
 	logging.Verbose.Println("   - EnforceRFC952: ", c.EnforceRFC952)
+	logging.Verbose.Println("   - EnforceRFC952: ", c.EnforceRFC952)
+	logging.Verbose.Println("   - ContainerIPsOn: ", c.ContainerIPsOn)
 
 	return c
 }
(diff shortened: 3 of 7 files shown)


records/generator_test.go
```diff
@@ -272,6 +272,14 @@ func TestInsertState(t *testing.T) {
 		t.Errorf("should return the container IPs 10.3.0.1, 10.3.0.2 for the task, but got %v", rrs)
 	}
 
+	rrs, ok = rg.As["_container.reviewbot.marathon.mesos."]
+	if !ok {
+		t.Error("should find this running task - A record")
+	}
+	if got, want := rrs, []string{"10.5.0.7"}; !reflect.DeepEqual(got, want) {
+		t.Errorf("should return the slave ips 10.5.0.7 for the task, but got %v", rrs)
```
tsenart: Can we not hardcode the wanted value in the string?

records/generator_test.go
```diff
@@ -272,6 +272,14 @@ func TestInsertState(t *testing.T) {
 		t.Errorf("should return the container IPs 10.3.0.1, 10.3.0.2 for the task, but got %v", rrs)
 	}
 
+	rrs, ok = rg.As["_container.reviewbot.marathon.mesos."]
+	if !ok {
+		t.Error("should find this running task - A record")
```
tsenart: Can we have a better error message here? I know the rest of the test is no example, but if we keep improving the field when we touch it, we'll be in very good shape very soon :)

records/generator.go
```diff
@@ -242,7 +242,10 @@ func sanitizedSlaveAddress(hostname string, spec labels.HostNameSpec) string {
 }
 
 func (t *Task) containerIP() string {
-	const containerIPTaskStatusLabel = "Docker.NetworkSettings.IPAddress"
+	const (
+		dockerLabel = "Docker.NetworkSettings.IPAddress"
```
jdef: are these keys doc'd somewhere?

records/generator_test.go
```diff
@@ -308,7 +316,7 @@ func TestInsertState(t *testing.T) {
 	}
 
 	// test for 5 A names
-	if got, want := len(rg.As), 16; got != want {
+	if got, want := len(rg.As), 18; got != want {
```
tsenart: Not that we should change this right now, but this test is not good. It has to change every time we add or remove a case. Better to test for the individual records.

tsenart:
Apart from the small remarks, LGTM!

tsenart:
PTAL: @sttts

records/generator.go
sttts: They are not even in Mesos yet, but will be soon.

sttts: I have added documentation for the new records.

sttts:
@tsenart ptal

docs/docs/naming.md
```diff
@@ -26,11 +26,36 @@ $ dig search.marathon.mesos
 search.marathon.mesos.		60	IN	A	10.9.87.94
 ```
  
+## Container IP A Records
+
+The A records above point to the host IP of a given service. If container IPs are supported by Mesos and the executor of a test, Mesos-DNS will create another A record for each of the A records above by prefixing it with `_container` and returning the container IP:
```
jdef: s/test/task ?

jdef: prefixing a label with an underscore `_container` is non-standard for A records, although rfc2181 seems to blow the doors wide open w/ respect to allowed characters (from the perspective of the server) and puts the onus on the clients to properly parse labels with previously-non-standard characters.

https://www.rfc-editor.org/rfc/rfc2181.txt (sec 11)

It may be worth noting this somewhere.

jdef: FYI http://www.zytrax.com/books/dns/apa/names.html

records/generator_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"testing/quick"
 
+	"fmt"
```
jdef: fmt is a std lib package, please move it up a section

docs/docs/naming.md
sttts: Thanks for pointing this out. I am not really sure how to formulate this. We use `_underscore` to avoid name clashes. But is there a danger that any client will refuse such a name? And if yes, what would be your suggestion? Something like `container.search.marathon.mesos`?

We can add a note like this:
```
*Note*: According to the original [RFC1035](https://www.ietf.org/rfc/rfc1035.txt) underscores are not allowed in A records. [RFC2181](https://www.ietf.org/rfc/rfc2181.txt) section 1 clarifies that underscores (and in fact any binary string) must be allowed by DNS servers. We use the underscore here to avoid naming clashes. The `_container` names are not meant to be hostnames (which forbit underscores as well).
```

But this feels like hand waving instead of a rigorous technical answer.

records/generator_test.go
sttts: fixed.

records/generator_test.go
```diff
@@ -250,87 +251,63 @@ func TestInsertState(t *testing.T) {
...
+	expectedARecords := []expectation{
+		{domain: "liquor-store.marathon.mesos.", num: 2, answer: []string{"1.2.3.11", "1.2.3.12"}},
+		{domain: "_container.liquor-store.marathon.mesos.", num: 2, answer: []string{"10.3.0.1", "10.3.0.2"}},
+		{domain: "_container.reviewbot.marathon.mesos.", num: 1, answer: []string{"10.5.0.7"}},
+		{domain: "poseidon.marathon.mesos.", reason: "task was not running"},
+		{domain: "_container.poseidon.marathon.mesos.", reason: "task was not running"},
+		{domain: "master.mesos.", num: 1},
+		{domain: "master0.mesos.", num: 1},
+		{domain: "leader.mesos.", num: 1},
+		{domain: "some-box.chronoswithaspaceandmixe.mesos.", num: 1}, // ensure we translate the framework name as well
```
tsenart: In table driven tests, it isn't strictly necessary to explicitly name the struct keys. It actually hinders readability IMO.

records/generator_test.go
```diff
@@ -250,87 +251,63 @@ func TestInsertState(t *testing.T) {
...
+	}
+	expectedSRVRecords := []expectation{
+		{domain: "_poseidon._tcp.marathon.mesos.", reason: "task was not running"},
+		{domain: "_leader._tcp.mesos.", num: 1},
+		{domain: "_liquor-store._tcp.marathon.mesos.", num: 1},
```
tsenart: Since you went ahead and refactored this test, we can do better still. Why would we want to test for the number of returned records instead of the records themselves? That would simplify and increase robustness.

With that in place the`reason` field seems unnecessary. All we're testing is if the returned records are what we expect. If not, `Errorf` it!

tsenart:
@sttts: Can you rebase with master, fix possible conflicts and squash the commits?
@jdef: Any remarks before we merge this in?

jdef:
I'm still not convinced about our approach of using _container for A
records. Why do we need _container at all?

Consider:

taskA.framework1.mesos (runs in container netns, has special IP)

taskB.framework2.mesos (runs in slave's netns, always maps to slave IP)

Is there really a need to distinguish between:

_container.taskA.framework1.mesos (special container IP)
taskA.framework1.mesos (slave's IP)

What am I missing?

sttts:
I think we have to distinguish. Everything else is very confusing: every container has a container IP, but depending on the setup it's not routable. How do you want to distinguish?

If all frameworks would support switching a task among those two network "modes", I would agree maybe. But we are far from that. Currently, the network model of Mesos is to have slave IPs. Then there are some special Docker setups such that a container gets a routable IP. This IP is published (with some addon Mesos plugins) via a status label. But Mesos in fact does not know, Marathon does not know. In the last case things like health checks will work with slave IP and traditional port management, while container IPs might be used between the services. In other words, we will live with both IPs existing beside each other for some time.

jdef:
It sounds like the key issue you're concerned with is routable IPs. Sadly there's no way for mesos-dns to understand a network infra without additional information.

The more I think about this, the more I don't like the _container prefix at all. I see where you were going with it but I also think that it muddies the water far too much. Users need a way to access a task at the IP that it's running on. Because they want information from it. Or to send information to it. Or both. If a user assigns a task an unroutable IP then I think they need to deal with that fact (but we can help, a little, with backward compat config options).

*Having access to container IPs turns the tables*: the case wherein a user wants to access the host address of the task becomes the edge/special case. Because normally the user just want to access the task on the container IP.

Here's what I recommend:
(a) slave-ip is the default for the task A record
(b) container-ip, if specified, overrides the default above - it's used for the task A record
(c) we introduce a config flag to mesos-dns that says "ignore container-ip if present" --> back compat, always choose (a) never do (b)

If we **really** need to identify the host of a task (this seems like a completely separate issue/PR):
(d) we introduce _node.taskA.framework1.mesos record (PTR to something) for the edge case where the user specifically wants the host of the task (even if the task has a separate container IP). I'm not convinced of the need of this.

sttts:
@jdef @tsenart PTAL

jdef:
lgtm

docs/docs/configuration-parameters.md
```diff
@@ -28,7 +28,8 @@ The configuration file should include the following fields:
   "SOARefresh": 60,
   "SOARetry":   600,
   "SOAExpire":  86400,
-  "SOAMinttl": 60
+  "SOAMinttl": 60,
+  "ContainerIPsOn": true
```
jdef: food for thought: since container-ip's are experimental we could default this to `false` for now. user's who want to play with container-ip can set this to true if they want to play with the feature.

thoughts?

sttts: On the other hand, the labels do not exist. But disabling it will speed up the record generation. Also false is fine for me.

tsenart:
FYI: I cherry-picked the `TestInsertState` table driven tests into master: https://github.com/mesosphere/mesos-dns/commit/fc28f8b889d59721893104d9f516441ea6c2b87e

jdef:
Since we now have a "slave.mesos" namespace, I'd like to propose an alternate to (d) above: instead of referring to the host addresses of a task via `_node.task.framework.mesos`, we instead use `task.framework.slave.mesos`.

jdef:
proposing an alternative to the bool flag described above in point (c):
ignoreContainerIPs := none | docker | mesos | all

none: process all container IPs for all recognized containerizers, default to slave IP if container IP is not present
docker | mesos | docker,mesos: ignore container IPs for the identified containerizers, separated by a comma if there's more than one to ignore
all: ignore all container IPs for all containerizers -- task A records are always generated with slave IPs

pyronicide:
@jdef Are you saying `slave` or `slave-id`? It makes sense as long as the user doesn't need to discover what slave the task is actually running on.

jdef:
@pyronicide I'm saying "slave". I don't expect users to know slave-id. mesos-dns just got support for "slave.mesos" --> returns A records for all slave's in the cluster. "my-task.my-framework.slave.mesos" returning A records for all the slaves on which "my-task" is running seems like a natural extension of the namespace.

pyronicide:
Then :+1:, I'd love to get rid of all the weird `_` stuff, particularly because it doesn't work with ping.

records/generator.go
```diff
@@ -293,27 +296,32 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
...
-					srvhost := trec + ":" + port
-					tcp := "_" + tname + "._tcp." + tail
-					udp := "_" + tname + "._udp." + tail
-					rg.insertRR(tcp, srvhost, "SRV")
-					rg.insertRR(udp, srvhost, "SRV")
+			if containerIP != "" {
+				// A records with container IP, no SRV records here
+				rg.insertRR(arec, containerIP, "A")
+				rg.insertRR(trec, containerIP, "A")
+			} else {
```
jdef: if we support A records for "my-task.my-framework.slave.mesos" then we need to nix this else condition and generate those records here.

we should also continue to generate SRV's here for any host ports allocated by the task, regardless of whether there's a container IP (why should mesos-dns make assumptions about such task's?)

jdef: to clarify, those SRV's would look like this:
`_my-task._tcp.my-framework.mesos` --> `my-task.my-framework.slave.mesos:{host-port}`

this should provide backwards compat for existing clients that are relying on SRV's generated for tasks that have allocated host ports.

docs/docs/naming.md
```diff
@@ -44,9 +74,9 @@ $ dig _search._tcp.marathon.mesos SRV
 
 ;; ANSWER SECTION:
 _search._tcp.marathon.mesos.	60 IN SRV 0 0 31302 10.254.132.41.
-``` 
+```
 
-SRV records are generated only for tasks that have been allocated a specific port through Mesos. 
+*Note*: SRV records are generated only for tasks that have been allocated a specific port through Mesos and that have no container IP assigned.
```
jdef: SRV records are generated for all tasks that have allocated host port resources through Mesos.

In the future, additional SRV records will be generated for tasks that provide DiscoveryInfo.

records/config.go
```diff
@@ -72,6 +72,9 @@ type Config struct {
 
 	// EnforceRFC952 will enforce an older, more strict set of rules for DNS labels
 	EnforceRFC952 bool
+
+	// Enable serving container IP if the corresponding task status labels are found
+	ContainerIPsOn bool
```
jdef:
```
IgnoreContainerIPs := 'never' | {containerizer-list} | 'always'
containerizer-list := containerizer-name | containerizer-name ',' containerizer-list
containerizer-name := 'mesos' | 'docker'
```

docs/docs/naming.md
sttts: But for container ip SRV records we have to know whether the discoveryinfo really means the container ports, not the host ports. So a framework needs knowledge about the infrastructure to publish the right ports.

jdef: my current thinking is that if DiscoveryInfo is specified then we generate SRV's:

- `_my-task._tcp.my-framework.mesos` --> `my-task.my-framework.mesos:{discovery-port}`

the framework chooses what to specify for discovery port. `my-task.my-framework...` is the name of the A record that we generate. that A record either points to a container-ip or host-ip, but SRV record generation shouldn't care.

if there's no discovery info, then we generate:

- `_my-task._tcp.my-framework.mesos` --> `{slave-host}:{host-port}`

jdef: this is also a problem that we don't have to solve just yet.

jdef:
Received a general consensus that this proposal is OK:

(a) A record(s) use container IP (for task.framework.mesos) if present; this is natural and least surprising for the user
(b) If no container IP, the A record points to the slave IP
(c) Explicit 'slave-IP' targeted A records for tasks: use task.framework.slave.mesos (slave.mesos is already a name that we return A records for).
(d) SRV: If there are host port resources allocated then _task._{proto}.framework.mesos:{host-port} --> task.framework.slave.mesos:{host-port}
(e) SRV: If there is discovery info (DI) then _task.{proto}.framework.mesos --> task.framework.mesos:{DI-port} (point to the A record, it resolves to the IP as per the rules above)
(f) mesos-dns provides a configuration option for backwards compat: IgnoreContainerIPs := 'never' | 'always' | {containerizer-list} (e.g. 'mesos' or 'mesos,docker', etc.)
(g) drop explicit 'container.task.framework.mesos' names -- they're less than ideal
(h) never use _ in A records because it breaks common tooling (like ping) on some platforms: stick with rfc952 charsets

extensions & explanations:
(d) and (e) .. there's probably room for improvement here
(f) a reasonable default is probably 'docker' since, by default, docker allocates non-cluster-routable IPs
(g) there's been a request for explicitly referencing the container IP of a task; suggested: `task.framework.container.mesos`.

we need to keep @pyronicide in the loop since customers have started developing on dcos and may have deps on `container.task.framework.mesos` now.

sttts:
@jdef about (d):

- We create SRV records for host ports even though container IP is set?
- The host name is `_task._{proto}.framework.mesos`, right? And especially not: `_task._{proto}.framework.slave.mesos`.

about (e):

- The hostname is `_task._{proto}.framework.mesos`, right?

sttts:
There is a WIP implementation (unclear (d) and (e) are missing) of the proposal in https://github.com/mesosphere/mesos-dns/pull/147

jdef:
Further clarification:

(d) SRV: If there are host port resources allocated then `_task._{proto}.framework.mesos` --> `task.framework.slave.mesos:{host-port}`

(e) SRV: If there is discovery info (DI) then `_task._{proto}.framework.mesos` --> `task.framework.mesos:{DI-port}` (point to the A record, it resolves to the IP as per the rules above)

Rule (d) is processed regardless of whether a container-IP is set (because there are consumed host-ports that likely need to be exposed). Perhaps if there is discovery-info for the task then we can ignore rule (d) since the framework is taking explicit steps to advertise services (and we'll publish those via rule (e)).

sttts:
My proposal for SRV records, might match with what you mean:

- DI available =>
  - _task._{proto}.framework.mesos -> task.framework.slave.mesos:{di-port}
- no DI available =>
  - _task._{proto}.framework.mesos -> task.framework.slave.mesos:{task-port}
- container IP =>
  - _task._{proto}.framework.slave.mesos records => task.framework.slave.mesos:{task-port}
- container IP && DI available =>
  - _task._{proto}.framework.container.mesos records => task.framework.container.mesos:{di-port}
- no container IP && DI available =>
  - _task._{proto}.framework.slave.mesos records => task.framework.slave.mesos:{di-port}

jdef:
I put together a table to help sort this out:
https://docs.google.com/spreadsheets/d/1fZcaUP2Y-MQluMl-JP_PDB1Qlg6Ad36OdSQKBj_zlu4/edit#gid=0

jdef:
@sttts @tsenart we're probably going to want an implementation of this proposal (or at least most of it) before templates are ready (they still look under development). @kozyraki is asking us to wrap this issue up by EOD Mon, Aug 10th.

sttts:
We will take a look on Monday. Should be doable, but won't be pretty.

kozyraki:
The table looks good. 

In case it helps, I don’t think we urgently need the {task}.framework.slave.mesos records.

sttts:
A clarification about the `IgnoreContainerIPs` config:
```
IgnoreContainerIPs := 'never' | 'always' | {containerizer-list} (e.g. 'mesos' or 'mesos,docker', etc.)
```
There is no containerizer field in the state.json. Hence, the only way to behave differently according to the config is by searching or ignoring the respective task status labels:
- `Docker.NetworkSettings.IPAddress`
- `MesosContainerizer.NetworkSettings.IPAddress`

jdef:
right. 'IgnoreContainerIPs=docker' says to ignore container IPs spec'd by
that label. 'IgnoreContainerIPs=mesos' says to ignore container IPs spec'd
by that label. and so on.

jdef:
replaced by #226

PR #199, opened 2015-07-29
check if slave IP string contains a valid IP address

lloesche:
This checks if the string returned by sanitizedSlaveAddress() contains a valid IP address.

### DIFF
--- a/records/generator.go
+++ b/records/generator.go
@@ -277,7 +277,12 @@ func (rg *RecordGenerator) InsertState(sj StateJSON, domain string, ns string,
 	for _, slave := range sj.Slaves {
 		ssa := sanitizedSlaveAddress(slave.Hostname, spec)
 		rg.SlaveIPs[slave.ID] = ssa
-		rg.insertRR("slave."+domain+".", ssa, "A")
+
+		if(net.ParseIP(ssa) != nil) {
+			rg.insertRR("slave."+domain+".", ssa, "A")
+		} else {
+			logging.VeryVerbose.Printf("string '%q' for slave with id %q is not a valid IP address", ssa, slave.ID)
+		}
 	}
 
 
--- a/records/generator.go
+++ b/records/generator.go
@@ -233,12 +233,12 @@ func hostToIP4(hostname string) (string, bool) {
 
 // attempt to convert the slave hostname to an IP4 address. if that fails, then
 // sanitize the hostname for DNS compat.
-func sanitizedSlaveAddress(hostname string, spec labels.HostNameSpec) string {
+func sanitizedSlaveAddress(hostname string, spec labels.HostNameSpec) (string, bool) {
 	address, ok := hostToIP4(hostname)
 	if !ok {
 		address = labels.AsDomainFrag(address, spec)
 	}
-	return address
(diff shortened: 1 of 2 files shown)


jdef:
Thanks for this. I think I'd rather see `sanitizedSlaveAddress` return `(string,bool)` where the bool indicates "valid ip address" (basically just forward `ok`); avoids redundant parsing.

lloesche:
Like so?

jdef:
yep. looks like a unit test fails to build now..

tsenart:
Why is this required?

lloesche:
Right sorry forgot about the tests. Should be fixed.

lloesche:
@tsenart if I understand it correctly there's cases where sanitizedSlaveAddress() will return 
labels.AsDomainFrag() instead of a valid IPv4 address. Which would then create invalid A records.

tsenart:
@lloesche: Then what you want to use is not `sanitizedSlaveAddress` but `address, ok := hostToIP4(hostname)` directly!

jdef:
@tsenart an implication of my suggested approach would preserve the existing behavior of assigning the result of sanitizedSlaveAddress to SlaveIPs, and address the corner case whereby a non-IP-address would be used to construct an A record.

I don't know where else SlaveIPs is used without further code review.

lloesche:
@tsenart but that's what sanitizedSlaveAddress() is doing already. Given a cluster of 10.000 slave nodes I would suddenly do 20k hostname lookups instead of 10k every 5 sec.

tsenart:
@lloesche: I'm not suggesting that :) I'm saying that I'd get rid of the `sanitizedSlaveAddress` and in-line it.

lloesche:
@tsenart but sanitizedSlaveAddress() was already used before I added my slave.mesos code to it.
The original code had an assignment
```
rg.SlaveIPs[slave.ID] = sanitizedSlaveAddress(slave.Hostname, spec)
```
And I just added my slave.mesos code in there. I don't know what rg.SlaveIPs[slave.ID] does or where it's used so can't just get rid of sanitizedSlaveAddress().

jdef:
I vote for keeping this change as is and we can open a new PR to refactor sanitizeSlaveAddress

tsenart:
@lloesche: Sure you can. It isn't used anywhere else.
```go
for _, slave := range sj.Slaves {
    address, ok := hostToIP4(slave.Hostname)
    if ok {
        rg.insertRR("slave."+domain+".", address, "A")
    } else {
        address = labels.AsDomainFrag(address, spec)
    }
    rg.SlaveIPs[slave.ID] = address
}
```

lloesche:
@tsenart ha cool, updated. How would I know that it's not used anywhere outside generator.go? Just grep the entire sources? I'm using vim to edit. Is there an IDE or some other way that would show me all references?

tsenart:
Yes, you can use grep. I personally use [Ag](https://github.com/ggreer/the_silver_searcher).
For smarter Go static analysis tools, install https://github.com/fatih/vim-go
Then you can call `:GoCallers` on your function.

tsenart:
LGTM

PR #207, opened 2015-07-31
Refactor labels package to reduce cyclomatic complexity

tsenart:
This change set reduces the cyclomatic complexity of the labels package and reduces its API surface.
With this patch, we're trading label generation performance for maintainability and easier correctness under the premise that the 3x slow-down due to allocations and garbage collection is acceptable even at large scales. On my Macbook, 1M tasks take 1.59748894s to generate without any code optimisation done yet. Here are the more detailed benchmarks and profiling information:

## Benchmarks
```
benchmark                       old ns/op     new ns/op     delta
BenchmarkDomainFrag_RFC952      1192          4198          +252.18%
BenchmarkDomainFrag_RFC1123     1206          3937          +226.45%

benchmark                       old allocs    new allocs    delta
BenchmarkDomainFrag_RFC952      13            48            +269.23%
BenchmarkDomainFrag_RFC1123     14            49            +250.00%

benchmark                       old bytes     new bytes     delta
BenchmarkDomainFrag_RFC952      192           984           +412.50%
BenchmarkDomainFrag_RFC1123     192           992           +416.67%
```

### Previous implementation
cpu

### New implementation
cpu

## Other ideas
1) Why don't we cache the results of label generation? These should never change for the same input.
2) Why is the whole state rebuilt on every state load? It'd be more efficient to incrementally apply state changes.


kozyraki:
Hi Tomas, 

first of all, there are really cool graphs and major kudos for measuring the performance impact of your changes. Great engineering discipline my fried, congrats!

Can you explain a little what where does the improvement in maintainability and correctness comes from? The graph in the bottom looks more complex than the one in the top. Where do the benefits come from? 

About the 3x slowdown in generation: it is probably ok. You are not changing the load on the master, which would be really bad with the current implementation of Mesos-DNS for large systems. 1.6s of generation for 1M tasks every 60s is not bad. We may want to double-check what happens to serving during generation. If Mesos-DNS has a couple of cores in most deployments, the impact of longer record generation on record serving should not be that bad. But we should double-check it. If this works out and the maintainability advantages are major, it is ok with me. 

Now, about your other questions:
- The reason we don’t cache the results of record generation is the following. Our only way of interacting with the master is to poll for the whole task state. If we wanted to cache records, we would have to scan through the new task state each time, find the task that are the same and skip them (most of them), find the tasks that changed and update them, and then find the missing tasks and erase their records. It’s actually simpler to regenerate everything from scratch and get it over with. I also bet it is faster as the diff would have non trivial overheads. 
- Ideally, we should not get the whole task state every 60 sec. We should just get task updates on launch or finish/failure and that would allow us to immediately generate/remove records accordingly. Unfortunately, Mesos does not have such an interface right now. There are many discussions on how to do this. Once there is one of them, we will switch Mesos-DNS. It will solve many issues around scaling. 

Christos

records/labels/labels.go
```diff
@@ -0,0 +1,83 @@
...
+const Sep = "."
+
+// DomainFrag mangles the given name in order to produce a valid domain fragment.
+// A valid domain fragment will consist of one or more host name labels
+// concatenated by the given separator.
+func DomainFrag(name, sep string, label Func) string {
+	var labels []string
+	for _, part := range strings.Split(name, sep) {
+		if lab := label(part); lab != "" {
+			labels = append(labels, lab)
```
sttts: **1st step:** This line causes a lot of mallocs by growing the slice. We can easily pre-allocate a `labels` array big enough for any sensible domain name.
**2nd step:** why collect the labels in an array at all? Join them in place => no malloc for the labels slice at all.

records/labels/labels.go
```diff
@@ -0,0 +1,83 @@
...
+		return '-'
+	default:
+		return -1
+	}
+}
+
+// trimCut cuts the given label at min(maxlen, len(label)) and ensures the left
+// and right cutsets are trimmed from their respective ends.
+func trimCut(label []byte, maxlen int, left, right string) []byte {
+	trim := bytes.TrimLeft(label, left)
```
sttts: We should avoid the generic bytes.TrimLeft/Right functions. They are far too generic. Our cutsets are trivial and the default impl creates yet another closure function for the cutset everytime we call it.

records/labels/labels.go
tsenart: I'm 100% sure all of this code can be optimised in a myriad of ways. That, however, was not the purpose of this PR.

sttts: Of course. But throwing around with arrays and closures when a local optimization is not more complex....

tsenart: Yes, we can do better indeed. I was hoping to defer that to a different PR but OK, I'll tackle it here.

records/labels/labels.go
tsenart: @sttts: I'm aware of the closure creation. But again, optimisation was not the purpose of this PR.

sttts:
Moving the optimizations to another PR is fine, especially because the performance is still "good enough".

sttts:
With that said, lgtm

tsenart:
@kozyraki: I had replied via email, but apparently nothing showed up here. So, here it goes!

> Hi Tomas, 

> first of all, there are really cool graphs and major kudos for measuring the performance impact of your changes. Great engineering discipline my fried, congrats!

> Can you explain a little what where does the improvement in maintainability and correctness comes from? The graph in the bottom looks more complex than the one in the top. Where do the benefits come from?

The cyclomatic complexity score of the previous implementation was high (15). In fact, this was a good metric since I had a quite hard time following and understanding the code in order to refactor it. The improvement in maintainability comes from the simplicity of the new implementation through reuse of standard library functions and hence substantial complexity reduction. The graph at the bottom isn't representative of the complexity exposed to us since it mostly reflects standard library code which isn't our responsibility to maintain.

The correctness improvements come from the reduced complexity, increased reasoning capacity and reliance on battle tested, well known standard library functions. On top of that, we're now doing serious property based testing which increase the confidence of the invariants that need to be honoured.
See for instance: https://github.com/mesosphere/mesos-dns/pull/207/files#diff-f970b921471415ae3df5fe81ccea93ddR80


> About the 3x slowdown in generation: it is probably ok. You are not changing the load on the master, which would be really bad with the current implementation of Mesos-DNS for large systems. 1.6s of generation for 1M tasks every 60s is not bad. We may want to double-check what happens to serving during generation. If Mesos-DNS has a couple of cores in most deployments, the impact of longer record generation on record serving should not be that bad. But we should double-check it. If this works out and the maintainability advantages are major, it is ok with me.

We're using a RWLock to protect the whole state. This means that once we Lock, all new readers will block until Unlock is called. This can lead to latency spikes in responding to DNS queries that happen during reloads.

This can be mitigated by using less restrictive concurrency control strategies. We could simply not hold the lock for so long, but only to update the reference, or we could get rid of the lock all together by using a sync.Value: http://golang.org/pkg/sync/atomic/#example_Value_config

> Now, about your other questions:
> - The reason we don’t cache the results of record generation is the following. Our only way of interacting with the master is to poll for the whole task state. If we wanted to cache records, we would have to scan through the new task state each time, find the task that are the same and skip them (most of them), find the tasks that changed and update them, and then find the missing tasks and erase their records. It’s actually simpler to regenerate everything from scratch and get it over with. I also bet it is faster as the diff would have non trivial overheads.
> - Ideally, we should not get the whole task state every 60 sec. We should just get task updates on launch or finish/failure and that would allow us to immediately generate/remove records accordingly. Unfortunately, Mesos does not have such an interface right now. There are many discussions on how to do this. Once there is one of them, we will switch Mesos-DNS. It will solve many issues around scaling.

I wonder how much slower such a diff algorithm would be. One advantage of doing this would be to refactor the code to already handle incremental state updates which, for now, would be generated by mesos-dns through diffing, but once Mesos itself exposes an event stream, could be very easily integrated with.

And finally, @kozyraki, thanks for taking the time to review this. :-)

jdef:
Couple of thoughts:

- I really like the changes to the test code -- very nice!
- I'm not convinced about the changes to the labels code. Trying to be unbiased here (since I wrote the original code) but the new code is actually harder for me to understand than the old stuff. "12" as a target complexity threshold seems somewhat arbitrary -- 15 isn't that might higher. I just don't see the gain here.
- I'll need to review the RW lock stuff, but last I checked locking/unlocking was happening around very tiny bits of code (get/set) vs. the entire record generation step.

tsenart:
The algorithm you implemented is fast and hard (to understand). IMO using common standard library functions to achieve the same thing massively reduces the cognitive burden of maintaining such code. I found it really hard to follow. We should actually strive to push the complexity threshold down, not up!

Regarding the RW lock, you're right indeed. Somehow I registered that the lock was being held longer than it is. That's good! Anyways, the lock-free version mentioned is the perfect use case for this. I opened an issue to experiment with it.

jdef:
LGTM

jdef:
:shipit:

PR #208, opened 2015-08-01
Generate slave SRV records

### LINKED ISSUE #200: create slave.mesos SRV records
https://github.com/mesosphere/mesos-dns/commit/659d1dc164f9359fc501157d0efb8335b4835676 added the ability to resolve slave.mesos and get a list of all slave IPs. It would be nice to also get SRV records including the slave ports.

Hints from @jdef 
```
the port shows up in two places in the slave state.json: flags.port and in pid. it's probably pretty simple to add a "pid" field to the slave struct, and then parse the port from that. sample data for pid:

slave(1)@10.2.0.5:5051
you can use mesos-go/upid to parse if you want (mesos-go is already a dep):
https://github.com/mesos/mesos-go/blob/master/upid/upid.go
```

tsenart:
Depends on #193 and closes #200.


tsenart:
PTAL: @lloesche, @sttts, @jdef

tsenart:
Only the last commit is relevant to this PR.

sttts:
Only one comment, otherwise lgtm

tsenart:
I suppose that in this case we really know that slaves don't use UDP at all. I'll drop it.

PR #209, opened 2015-08-01
Generate framework A and SRV records

### LINKED ISSUE #202: Resolve schedulers as well as tasks.
I would like to resolve where my scheduler is running. `ping marathon.mesos` should return the IP of the scheduler that is connected to mesos.

tsenart:
Depends on #208 and closes #202.


tsenart:
PTAL: @sttts, @jdef, @pyronicide

tsenart:
Only the last commit is relevant.

sttts:
One comment, otherwise lgtm

sttts:
Same question as for slave: why udp. Otherwise lgtm

tsenart:
Oh, I see. That's scheduler Mesos port which is TCP only. Cool, I'll drop the UDP.

jdef:
Why did we add an SRV record for schedulers? The PID port is the driver
port, and should only be used by mesos -- what need is there for it to be
discoverable?

pyronicide:
@jdef you could use the webui_url? That's an actual url though, so I'm not sure it works out right.

jdef:
we could probably attempt to parse webui_url for a host:port. if not found,
we could default to the scheduler PID info (this merged PR). how does that
sound?

On Sat, Aug 1, 2015 at 10:56 AM, Thomas Rampelberg <notifications@github.com

pyronicide:
:+1: that sounds perfect.

PR #212, opened 2015-08-01
[WIP] Dnstap support

jdef:
What queries is mesos-dns processing? What responses is it yielding? Someone invented [dnstap](http://dnstap.info/) a while back to introspect DNS server query processing. This is a prototype for such support. _dnstap provides wire-speed logging for DNS queries and responses._

To use:

``` shell
$ ./mesos-dns -config=/tmp/config.json -logtostderr &
...
$ ./dnstap &
...
$ dig @localhost -p 8053 leader.mesos >/dev/null # discard output from dig
# .. what follows is the output of the dnstap program ..
type: MESSAGE
message:
  type: CLIENT_QUERY
  query_time: !!timestamp 2015-08-01 15:02:21.401094109
  socket_family: INET
  socket_protocol: UDP
  response_address: ::
  query_message: |
    ;; opcode: QUERY, status: NOERROR, id: 37374
    ;; flags: rd ad; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

    ;; QUESTION SECTION:
    ;leader.mesos.      IN       A

    ;; ADDITIONAL SECTION:

    ;; OPT PSEUDOSECTION:
    ; EDNS: version 0; flags: ; udp: 4096

---
type: MESSAGE
message:
  type: CLIENT_RESPONSE
  query_time: !!timestamp 2015-08-01 15:02:21.401094109
  response_time: !!timestamp 2015-08-01 15:02:21.401135991
  socket_family: INET
  socket_protocol: UDP
  response_address: ::
  response_message: |
    ;; opcode: QUERY, status: NOERROR, id: 37374
    ;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

    ;; QUESTION SECTION:
    ;leader.mesos.      IN       A

    ;; ANSWER SECTION:
    leader.mesos.       60      IN      A       10.2.0.7

---
```

This prototype still needs some work:
- [ ] avoid introducing a secondary protobuf lib: use the one we already have (will require regeneration of the dnstap protos)
- [x] the DNS server setup code is pretty awful. need a cleaner solution
- [x] dns.Msg objects are re-Pack()'d so that we can introspect. Not good for performance
  - see https://github.com/miekg/dns/pull/236
- [ ] would be nice to avoid the overhead of building tap protos if no one is actually listening
- [ ] ... probably other things that I'm not thinking of ATM


tsenart:
YES. Monitoring for Mesos DNS!

Now, I'd love to understand how monitoring a DNS server is different from monitoring any other server software and why we shouldn't use a general monitoring solution such as Prometheus for it. What are the advantages of using `dnstap`? **+16,904** git additions need to be well justified! :-)

jdef:
The short of it is: prometheus is great for metrics. dnstap actually get's you copies of the DNS query + response messages. Can prometheus do something that I'm not aware of?

There's a nice presentation on daptap here: http://dnstap.info/slides/dnstap.html#(1)

tsenart:
Why would we need full copies of all DNS query and response messages? What is the use case? For monitoring, that's really not necessary. I also have other initial concerns:
1. One more special purpose infrastructure piece whose entire life cycle needs to be managed.
2. How is the data going to be secured and its access controlled?
3. How is the data going to be garbage collected?
4. What is the performance implication of having all this data written to the local disk?
5. How would this piece fit in the bigger DCOS monitoring story?
6. That looks like a very good deal of non trivial code to maintain..

jdef:
It's possible that the intent of this PR is somewhat misunderstood. This is not a monitoring solution. This is not a metrics solution. It's a wire-speed DNS logging solution.

(James's primary) use case: as a sysadmin I want tools to help me debug a complicated DNS network that I've installed: it consists of mesos-dns operating in conjunction with other DNS servers in my infrastructure. mesos-dns doesn't log request/responses of *any* of the queries it processes, which makes my job harder. Some of my dns servers already support a technology called dnstap, and for those that don't it's a [planned feature](https://la51.icann.org/en/schedule/mon-tech/presentation-isc-bind-13oct14-en.pdf). It would be great if I could use *one* set of tools for all of my DNS servers when I'm debugging a problem.

- I'd much prefer this tooling vs. tcpdump'ing tcp/udp port 53 in production, which is extra hard to do if I don't have root on a system. An alternative to tcpdump would be to grep the logs for request/response details --- oh wait, **we don't log that** ?! :p Probably better that we don't log it: clutters the logs and hides possibly useful ERROR messages.
- It's possible to imagine that data generated from this tool could be stored to disk (for deferred audits, etc). That's not required, nor is it the only use case. Access (secured) to persistent data is therefore a related problem, but not one that this tooling intends to deal with -- at all. A sysadmin decides what to do with any data extracted from a DNS server (as they do with data extracted from any of their servers -- not sure how this is different?).
- The mesos-dns/dnstap buffers are kept in memory -- nothing is paged to disk. They're ring buffers, so old data is cycled out automatically as the buffer overflows. When dnstap client's go away, so do their ring buffers. I haven't profiled this code yet, but I'm expecting that overhead is not very high (except for cases where DNS messages are re-Pack'd -- this overhead has not been measured, but it's obviously not ideal).
- It's possible that this *might* be used for ongoing system monitoring. For me, having a tool like this would be helpful for real-time debugging (vs. 24x7 real-time monitoring). When I'm done, I kill the client and the server cleans up the session resources. No more overhead (this is the goal .. there's some work to do here still, as noted in the code and in the TODO's of this PR).
- I'm more than happy to clean up the design so that it's more maintainable. This initial rev is a toy to play with in order to assess usefulness. Much of the code is, in fact, quite trivial. The only somewhat sticky part so far has been proper dead-client-connection detection in Go to avoid TCP sockets stuck in a CLOSE_WAIT state. But that's been resolved.

jdef:
Closing this out since no one seems interested in landing this.

PR #215, opened 2015-08-04
Fix build (at last)

tsenart:
This change set fixes the last failures of the our CI.


tsenart:
PTAL: @jdef, @sttts

tsenart:
Weird: I got one test which fails in CI but not locally. Investigating!

tsenart:
@jdef, @sttts: The build is GREEN :-)

sttts:
:+1:

main.go
```diff
@@ -65,15 +66,6 @@ func main() {
 		newLeader = leader
```
sttts: why this temporary leader variable?

main.go
```diff
@@ -101,12 +93,8 @@ func main() {
...
 		case <-newLeader:
 			tryReload()
-		case err := <-dnsErr:
-			handleServerErr("DNS server", err)
-		case err := <-httpErr:
-			handleServerErr("HTTP server", err)
-		case err := <-zkErr:
-			handleServerErr("ZK watcher", err)
+		case err := <-errch:
+			logging.Error.Fatal(err)
```
sttts: err could be nil. The Fatal(nil) will do what then?

resolver/resolver_test.go
```diff
@@ -481,3 +297,179 @@ func TestLaunchZK(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	var sj records.StateJSON
+	err = json.Unmarshal(b, &sj)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	masters := []string{"144.76.157.37:5050"}
```
sttts: res.config.Masters

resolver/resolver_test.go
```diff
@@ -481,3 +297,179 @@ func TestLaunchZK(t *testing.T) {
...
+	spec := labels.RFC952
+	res.rs = &records.RecordGenerator{}
```
sttts: This is done by the New call above already.

resolver/resolver_test.go
```diff
@@ -481,3 +297,179 @@ func TestLaunchZK(t *testing.T) {
...
+		Masters:    []string{"144.76.157.37:5050"},
+		TTL:        60,
+		Port:       53,
+		Domain:     "mesos",
+		Resolvers:  records.GetLocalDNS(),
+		Listener:   "127.0.0.1",
+		SOARname:   "root.ns1.mesos.",
+		SOAMname:   "ns1.mesos.",
+		HTTPPort:   8123,
+		ExternalOn: true,
```
sttts: EnforceRFC952: true

circle.yml
```diff
@@ -10,4 +10,5 @@ test:
     override:
         - ! gofmt -s -d . 2>&1 | read
         - gometalinter --cyclo-over=12 ./...
-        - go test -v -race -timeout=120s ./...
+        - go test -v -timeout=2m ./...
```
sttts: The first call for quick feedback and the second for slow race detection?

main.go
```diff
@@ -41,23 +41,24 @@ func main() {
 	// initialize resolver
 	config := records.SetConfig(*cjson)
 	resolver := resolver.New(version, config)
-
-	var dnsErr, httpErr, zkErr <-chan error
-	var newLeader <-chan struct{}
+	errch := make(chan error)
```
sttts: errCh

resolver/resolver_test.go
```diff
@@ -309,127 +84,168 @@ func TestNonMesosHandler(t *testing.T) {
...
+	for i, tt := range []struct {
+		handle dns.HandlerFunc
+		query  dns.Msg
+		assert func(*dns.Msg) []error
+	}{
+		{
+			res.HandleMesos,
+			query("chronos.marathon.mesos.", dns.TypeA),
+			message(
+				header(0, 0, true, dns.RcodeSuccess),
```
sttts: 3 essentially untyped parameters... always have to lookup what they mean. How about some constants?

resolver/resolver_test.go
```diff
@@ -309,127 +84,168 @@ func TestNonMesosHandler(t *testing.T) {
...
+				"host": "leader.mesos.",
+				"ip":   "1.2.3.4",
+			}},
+		},
+	} {
+		if resp, err := http.Get(ts.URL + tt.path); err != nil {
+			t.Error(err)
+		} else if got, want := resp.StatusCode, tt.code; got != want {
+			t.Errorf("GET %s: StatusCode: got %d, want %d", tt.path, got, want)
+		} else if err := json.NewDecoder(resp.Body).Decode(&tt.got); err != nil {
```
sttts: Not really readable. Needs some seconds to find the `Decode(&tt.got)` with side-effect on `tt.got`.

resolver/resolver_test.go
```diff
@@ -309,127 +84,168 @@ func TestNonMesosHandler(t *testing.T) {
...
-	if err != nil {
-		t.Error(err)
-	}
-	var got5 []map[string]interface{}
-	err = json.Unmarshal(g5, &got5)
-	correct5 := []map[string]interface{}{{"host": "leader.mesos.", "ip": "1.2.3.4"}}
-	eq5 := reflect.DeepEqual(got5, correct5)
-	if !eq5 {
-		t.Error("HTTP hosts API failure")
+	for _, tt := range []struct {
```
sttts: tt means what? Table test? test or tc (=test-case) or even t would be more intuitive.

resolver/resolver_test.go
```diff
@@ -309,127 +84,168 @@ func TestNonMesosHandler(t *testing.T) {
...
 	// setup DNS server (just http)
-	res, err := fakeDNS(8053)
-	if err != nil {
-		t.Error(err)
-	}
+	res := fakeDNS(t)
 	res.version = "0.1.1"
 
 	res.configureHTTP()
 	ts := httptest.NewServer(http.DefaultServeMux)
```
sttts: ts stands for test server? srv or server would be more intuitive.

resolver/resolver_test.go
```diff
@@ -481,3 +297,179 @@ func TestLaunchZK(t *testing.T) {
 		// expected
 	}
 }
+
+func message(asserts ...func(*dns.Msg) []error) func(*dns.Msg) []error {
```
sttts: why call it message? messageAssertions would be more verbose and actually explain that it's about assertions.

sttts: or alternatively use a list of assertions in the test table struct above?

sttts:
Comments a few smaller things. Overall it looks good.

circle.yml
tsenart: The problem is that race enabled property based tests with the `testing/quick` package take FOREVER. Hence, those are skipped with the `-short` flag below.

main.go
tsenart: I'm not so fond of that naming scheme for channels.

main.go
tsenart: It's not temporary as it's defined above on line 56. I suppose the comment left there is trying to explain the why.

main.go
tsenart: It will log `<nil>`.

sttts: It should tell about the terminated service I guess. Or are we sure this can only happen intentionally on shutdown?

resolver/resolver_test.go
tsenart: They are typed. However, just like any other API, it takes time learning. Since the first two arguments are constant as of now, we could simply drop them.

resolver/resolver_test.go
tsenart: I suppose so. I haven't change it. I'll rename it to srv.

resolver/resolver_test.go
tsenart: It's a convention from the Go stdlib codebase. I think it stands for table test.

tsenart: You can't use `t` since it's the test function argument. I'd stick to the convention.

resolver/resolver_test.go
tsenart: What do you suggest?

resolver/resolver_test.go
tsenart: `messageAssertions` doesn't give any more information than the context in which this function is used. I had it named `assert` before but fell back to `message` since we're constructing a `message` assertion out of many `assertions` which have type `func(*dns.Msg) []error`

resolver/resolver_test.go
tsenart: Cool, thanks. This function was pretty much left unchanged, but I'll address these comments.

resolver/resolver_test.go
tsenart: I mean, I'm not sure why it's unreadable. What exactly makes it unreadable for you?

resolver/resolver_test.go
tsenart: Done.

main.go
tsenart: It doesn't matter because `logging.Error.Fatal` will terminate the whole process.

main.go
sttts: fine with me. What about giving the other channels also a ch postfix then?

main.go
sttts: No, I meant `leader`, not `newLeader`.

sttts:
lgtm

records/generator.go
```diff
@@ -176,12 +176,12 @@ func (rg *RecordGenerator) loadFromMaster(ip string, port string) (sj StateJSON)
 	if err != nil {
 		logging.Error.Println(err)
 	}
-	defer resp.Body.Close()
 
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		logging.Error.Println(err)
 	}
+	_ = resp.Body.Close()
```
jdef: the docs use `defer resp.Body.Close()`. not sure why this was changed. using `defer` provides "finally"-like semantics, which I think are appropriate in this case. If for some reason there's a panic that bubbles up we probably want to make sure that the Body is closed.

resolver/resolver_test.go
```diff
@@ -309,127 +83,173 @@ func TestNonMesosHandler(t *testing.T) {
...
 
-	// test A records
-	msg, err = fakeQuery("google.com", dns.TypeA, "udp", port)
-	if err != nil {
-		t.Error(err)
-	}
-
-	if len(msg) < 1 {
-		t.Errorf("not serving up A records, expected 2 records instead of %d: %+v", len(msg), msg)
+	for i, tt := range []struct {
```
jdef: :+1: :+1: :+1:

records/generator.go
tsenart: Yes, and we should totally use defer whenever we can use it. It's my fault I didn't explain this change well enough. So, here's the deal: One of the linters in our CI is `errcheck` which detects ignored errors in Go programs. Unfortunately, it creates tension with the common `defer r.Close()` Go idiom. Now, errors returned by `Close` methods can be safely ignored *as long as* what was closed was read-only. Whenever there are writes involved, errors returned by close should absolutely not be ignored.

What is lacking, then, is the appropriate tooling to handles such intricacies intelligently. In this PR, I did the easiest thing that would silence the CI failure, but I knew that it wasn't the right thing to do. I will follow up in https://github.com/mesosphere/mesos-dns/issues/223

PR #216, opened 2015-08-06
Modulerize records package

sttts:
- split `InsertState` as much as possible
- move all state.json logic into subpackage `state`

This is the preparation for the DiscoveryInfo support.


records/generator.go
```diff
@@ -244,106 +183,105 @@ func hostToIP4(hostname string) (string, bool) {
...
-			address = labels.DomainFrag(address, labels.Sep, spec)
-		}
-		rg.SlaveIPs[slave.ID] = address
-	}
+	rg.frameworkRecords(sj, domain, spec)
+	rg.slaveRecords(sj, domain, spec)
+	rg.listenerRecord(listener, ns)
+	rg.masterRecord(domain, masters, sj.Leader)
 
 	// complete crap - refactor me
```
tsenart: Why don't we have an `rg.taskRecords(...)` ?

records/generator.go
```diff
@@ -244,106 +183,105 @@ func hostToIP4(hostname string) (string, bool) {
...
-				for _, port := range ports {
-					srvhost := trec + ":" + port
-					tcp := "_" + tname + "._tcp." + tail
-					udp := "_" + tname + "._udp." + tail
-					rg.insertRR(tcp, srvhost, "SRV")
-					rg.insertRR(udp, srvhost, "SRV")
-				}
-			}
+			// Add RFC 2782 SRV records
+			task.ForEachPort(func(port string) {
```
tsenart: This sort of closure iteration isn't very idiomatic in Go. What prevents us from having a method which simply returns a slice of ports and ranging over that instead?

records/generator.go
```diff
@@ -244,106 +183,105 @@ func hostToIP4(hostname string) (string, bool) {
...
 			if !ok || (task.State != "TASK_RUNNING") {
 				continue
 			}
 
-			tname := spec(task.Name)
-			sid := slaveIDTail(task.SlaveID)
-			tag := hashString(task.ID)
-			tail := fname + "." + domain + "."
+			// context used to build domain names
+			context := struct {
```
tsenart: s/context/ctx/`? It'd make lines shorter without sacrificing any information value.

tsenart: Also, more compact:
```go
ctx := struct{ TaskName, TaskID, SlaveID string }{
	spec(task.Name), hashString(task.ID), slaveIDTail(task.SlaveID),
}
```

records/generator.go
```diff
@@ -244,106 +183,105 @@ func hostToIP4(hostname string) (string, bool) {
...
+			}{
+				SlaveID:  slaveIDTail(task.SlaveID),
+				TaskID:   hashString(task.ID),
+				TaskName: spec(task.Name),
+			}
 
-			// A records for task and task-sid
-			arec := tname + "." + tail
+			// insert canonical A records
+			trec := context.TaskName + "-" + context.TaskID + "-" + context.SlaveID + "." + tail
```
tsenart: Eventually I'd like to have these name constructions extracted into well tested, single purpose functions.

records/generator_test.go
```diff
@@ -261,19 +212,19 @@ func TestInsertState(t *testing.T) {
 }
 
 func TestPID_UnmarshalJSON(t *testing.T) {
```
tsenart: This test should move to the `state` package.

records/state/utils.go
```diff
@@ -0,0 +1,63 @@
+package state
```
tsenart: IMO `utils`, `misc` and such packages are a smell. They're the trash bin of programming taxonomy. :-)

These methods belong where the types they act on are. Let's put them in `types.go`. Better still, I'd rename `types.go` to `state.go`

records/state/utils.go
```diff
@@ -0,0 +1,63 @@
...
+		ports := yankPorts(t.Resources.Ports)
+		for _, port := range ports {
+			f(port)
+		}
+	}
+}
+
+// ContainerIP extracts a container ip from a Mesos state.json task. If not
+// container ip is provided, an empty string is returned.
+func (t *Task) ContainerIP() string {
```
tsenart: I know you just moved these methods, but there's no reason to pass a pointer to a Task when we're not mutating it.

records/state/utils_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+func TestRangePorts(t *testing.T) {
+	p := "[31115-31117]"
+
+	ports := yankPorts(p)
+
+	if len(ports) != 3 {
+		t.Error("not parsing ports")
```
tsenart: You should `Fatal` here, otherwise the following code might panic.

resolver/resolver_test.go
```diff
@@ -419,13 +420,12 @@ func fakeDNS(t *testing.T) *Resolver {
...
 	}
 
-	var sj records.StateJSON
+	var sj state.State
 	err = json.Unmarshal(b, &sj)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	res.config.Masters = []string{"144.76.157.37:5050"}
```
tsenart: What happened here?

records/state/utils.go
```diff
@@ -0,0 +1,63 @@
...
+
+		for t := lo; t <= hi; t++ {
+			yports = append(yports, strconv.Itoa(t))
+		}
+	}
+	return yports
+}
+
+// ForEachPort calls the given function f for each defined port in the given task
+func (t *Task) ForEachPort(f func(port string)) {
```
tsenart: As previously mentioned, let's just have a `func (t Task) Ports() []string`.

records/generator.go
tsenart: Something like `func CanonicalName(taskName, taskID, slaveID, frameworkName, domain string) string`

records/generator.go
sttts: you are right.

records/generator.go
sttts: looks good

records/generator.go
sttts: This will get replaced by patterns. Just leave it like it is here.

records/generator_test.go
sttts: right

records/state/utils.go
sttts: yes, please. In fact it was the trash bin while refactoring initially ;-)

records/state/utils.go
sttts: What do you propose? t Task? a task is a pretty big struct. Is go clever enough not to copy everything?

records/state/utils_test.go
sttts: in fact it's the old code, only moved. But please change it.

resolver/resolver_test.go
sttts: It's in the config already above.

records/state/utils.go
sttts: same question as above.

tsenart:
LGTM

PR #218, opened 2015-08-06
records: Fallback to framework hostname if PID not present

tsenart:
Apparently, some Mesos installations don't provide PIDs for frameworks. This patch makes the code fallback to the `hostname` field if not present. In such cases, no SRV records will be generated.

Fixes #127

/cc @lloesche @sttts

### DIFF
--- a/factories/fake.json
+++ b/factories/fake.json
@@ -130,6 +130,27 @@
         },
         {
             "active": true,
+            "checkpoint": false,
+            "completed_tasks": [],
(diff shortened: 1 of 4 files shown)


tsenart:
After some research, I found that framework PIDs are only returned from Mesos after this commit: https://github.com/apache/mesos/commit/f3691eafa200a0de551a019adf2eeb2495ec4494

PR #226, opened 2015-08-10
Quick implementation of container IP records

sttts:
- SRV records according to https://github.com/mesosphere/mesos-dns/pull/197#issuecomment-128594330 and https://docs.google.com/spreadsheets/d/1fZcaUP2Y-MQluMl-JP_PDB1Qlg6Ad36OdSQKBj_zlu4/edit#gid=0
- remove _container prefix and add slave.mesos A records instead for the non-container IP.


records/generator.go
```diff
@@ -344,42 +343,76 @@ func (rg *RecordGenerator) listenerRecord(listener string, ns string) {
...
+
+			if !task.HasDiscoveryInfo() {
+				continue
+			}
+
+			for _, port := range task.DiscoveryInfo.Ports.DiscoveryPorts {
+				target := canonical + tail + ":" + strconv.Itoa(port.Number)
+
+				// use protocol if defined, fallback to tcp+udp
+				if port.Protocol != "" {
```
jdef: should we attempt to sanitize the `port.Protocol` label before splicing it in?

tsenart:
@jdef: Why did you add the WIP label? Planning to do some work on it?

jdef:
@tsenart yes, I'm going to update docs for A and SRV records

tsenart:
:+1:

records/config.go
```diff
@@ -73,6 +73,34 @@ type Config struct {
...
+		SOAExpire:      86400,
+		SOAMinttl:      60,
+		Resolvers:      []string{"8.8.8.8"},
+		Listener:       "0.0.0.0",
+		HTTPPort:       8123,
+		DNSOn:          true,
+		HTTPOn:         true,
+		ExternalOn:     true,
+		RecurseOn:      true,
+		IPSources:      []string{"docker", "mesos", "host"},
```
jdef: the default should not include "docker" since Docker, by default, uses IPs that are not cluster routable

PR #227, opened 2015-08-10
Misc bugfix

jdef:
Fix some core bugs that I stumbled into while working on dnstap stuff.

### DIFF
--- a/records/config.go
+++ b/records/config.go
@@ -159,12 +159,15 @@ func readConfig(file string) (*Config, error) {
 		RecurseOn:      true,
 	}
 
+	workingDir := "."
 	usr, err := user.Current()
 	if err != nil {
-		return nil, fmt.Errorf("couldn't retrieve current user: %v", err)
+		logging.Error.Println("Failed to determine current user, translating ~/ to ./, error was", err)
+	} else {
+		workingDir = usr.HomeDir
 	}
 
-	c.File, err = filepath.Abs(strings.Replace(file, "~/", usr.HomeDir+"/", 1))
+	c.File, err = filepath.Abs(strings.Replace(file, "~/", workingDir+"/", 1))
 	if err != nil {
 		return nil, fmt.Errorf("cannot find configuration file")
 	} else if bs, err := ioutil.ReadFile(c.File); err != nil {
--- a/records/config.go
+++ b/records/config.go
@@ -162,6 +162,7 @@ func readConfig(file string) (*Config, error) {
 	workingDir := "."
 	usr, err := user.Current()
 	if err != nil {
+		// this can happen (on Linux) if you're running mesos-dns as a non-root user.
 		logging.Error.Println("Failed to determine current user, translating ~/ to ./, error was", err)
 	} else {
 		workingDir = usr.HomeDir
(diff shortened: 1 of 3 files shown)


resolver/resolver.go
```diff
@@ -360,15 +360,15 @@ func (res *Resolver) HandleMesos(w dns.ResponseWriter, r *dns.Msg) {
 	name := strings.ToLower(cleanWild(r.Question[0].Name))
 	switch r.Question[0].Qtype {
 	case dns.TypeSRV:
-		errs = append(errs, res.handleSRV(rs, name, m, r))
+		errs.Add(res.handleSRV(rs, name, m, r))
```
tsenart: What was the bug you found here?

resolver/resolver.go
```diff
@@ -710,6 +710,22 @@ func cleanWild(name string) string {
...
+func (e multiError) Add(err error) multiError {
+	if me, ok := err.(multiError); ok {
+		return append(e, me...)
+	} else if err != nil {
+		return append(e, err)
+	}
+	return e
+}
+
+func (e multiError) AddAll(err ...error) multiError {
```
tsenart: There's no need for having both `AddAll` and `Add`.

records/config.go
```diff
@@ -159,12 +159,15 @@ func readConfig(file string) (*Config, error) {
 		RecurseOn:      true,
 	}
 
+	workingDir := "."
 	usr, err := user.Current()
 	if err != nil {
-		return nil, fmt.Errorf("couldn't retrieve current user: %v", err)
+		logging.Error.Println("Failed to determine current user, translating ~/ to ./, error was", err)
```
tsenart: Why was this a bug before?

jdef: if I start mesos-dns as non-root user then Current() returns nil for me

resolver/resolver.go
```diff
@@ -377,14 +377,14 @@ func (res *Resolver) HandleMesos(w dns.ResponseWriter, r *dns.Msg) {
...
-		errs = append(errs, res.handleEmpty(rs, name, m, r))
+		errs.Add(res.handleEmpty(rs, name, m, r))
 	} else {
 		shuffleAnswers(res.rng, m.Answer)
 		logging.CurLog.MesosSuccess.Inc()
 	}
 
 	if !errs.Nil() {
-		logging.Error.Println(errs)
+		logging.Error.Println(errs.Error())
```
jdef: the symptom of the bug was showing here: blank error lines

resolver/resolver.go
jdef: agreed, will consolidate

jdef: and i'll add a test case for the bug that i'm fixing with this impl

tsenart: :clap:

resolver/resolver.go
jdef:
```
slave1_1      | ERROR: 2015/08/12 19:55:55 resolver.go:387: ; 
slave1_1      | ERROR: 2015/08/12 19:55:55 resolver.go:387: ; 
slave1_1      | ERROR: 2015/08/12 19:55:55 resolver.go:387: ; 
```

PR #240, opened 2015-09-04
nameserver address needs a port when attempting to leverage soa NS records

jdef:
/cc @s-urbaniak


tsenart:
LGTM

tsenart:
I'll follow up with writing a test for this case.

jdef:
Awesome - thanks!

PR #242, opened 2015-09-04
Exchanger

tsenart:
This PR introduces the exchanger package which refactors and tests a lot of the code used in the "external resolver".

Additionally, I extracted testing utilities and convenience functions I had in the `resolver` package to a new `dnstest` package.

### DIFF
--- a/exchanger/exchanger.go
+++ b/exchanger/exchanger.go
@@ -1,0 +1,119 @@
+package exchanger
+
+import (
+	"log"
+	"net"
+	"time"
+
+	"github.com/mesosphere/mesos-dns/logging"
+	"github.com/miekg/dns"
+)
+
+// Exchanger is an interface capturing a dns.Client Exchange method.
+type Exchanger interface {
+	// Exchange performs an synchronous query. It sends the message m to the address
+	// contained in addr (host:port) and waits for a reply.
+	Exchange(m *dns.Msg, addr string) (r *dns.Msg, rtt time.Duration, err error)
+}
+
+// Func is a function type that implements the Exchanger interface.
+type Func func(*dns.Msg, string) (*dns.Msg, time.Duration, error)
+
+// Exchange implements the Exchanger interface.
+func (f Func) Exchange(m *dns.Msg, addr string) (*dns.Msg, time.Duration, error) {
+	return f(m, addr)
+}
+
+// A Decorator adds a layer of behaviour to a given Exchanger.
+type Decorator func(Exchanger) Exchanger
+
+// Decorate decorates an Exchanger with the given Decorators.
+func Decorate(ex Exchanger, ds ...Decorator) Exchanger {
+	decorated := ex
+	for _, decorate := range ds {
+		decorated = decorate(decorated)
+	}
+	return decorated
+}
+
+// Pred is a predicate function type for dns.Msgs.
+type Pred func(*dns.Msg) bool
+
+// While returns an Exchanger which attempts the given Exchangers while the given
+// predicate function returns true for the returned dns.Msg, an error is returned,
+// or all Exchangers are attempted, in which case the return values of the last
+// one are returned.
+func While(p Pred, exs ...Exchanger) Exchanger {
+	return Func(func(m *dns.Msg, a string) (r *dns.Msg, rtt time.Duration, err error) {
+		for _, ex := range exs {
+			if r, rtt, err = ex.Exchange(m, a); err != nil || !p(r) {
+				break
+			}
+		}
+		return
+	})
+}
+
+// ErrorLogging returns a Decorator which logs an Exchanger's errors to the given
+// logger.
+func ErrorLogging(l *log.Logger) Decorator {
+	return func(ex Exchanger) Exchanger {
+		return Func(func(m *dns.Msg, a string) (r *dns.Msg, rtt time.Duration, err error) {
+			defer func() {
+				if err != nil {
+					l.Printf("error exchanging with %q: %v", a, err)
+				}
+			}()
+			return ex.Exchange(m, a)
+		})
+	}
+}
+
+// Instrumentation returns a Decorator which instruments an Exchanger with the given
+// counter.
+func Instrumentation(c logging.Counter) Decorator {
+	return func(ex Exchanger) Exchanger {
+		return Func(func(m *dns.Msg, a string) (*dns.Msg, time.Duration, error) {
+			defer c.Inc()
+			return ex.Exchange(m, a)
+		})
+	}
+}
+
+// A Recurser returns the addr (host:port) of the next DNS server to recurse a
+// Msg to. Empty returns signal that further recursion isn't possible or needed.
+type Recurser func(*dns.Msg) string
+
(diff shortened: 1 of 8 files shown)


resolver/resolver.go
```diff
@@ -311,15 +287,9 @@ func (res *Resolver) HandleNonMesos(w dns.ResponseWriter, r *dns.Msg) {
 		// set refused
 		m.SetRcode(r, 5)
 	} else {
-
-		proto := "udp"
-		if _, ok := w.RemoteAddr().(*net.TCPAddr); ok {
```
tsenart: I'm don't think that using the same protocol as the incoming DNS message used to be transported is required when recursing. Please advise otherwise.

exchanger/exchanger.go
```diff
@@ -0,0 +1,83 @@
...
+			return ex.Exchange(m, a)
+		})
+	}
+}
+
+// A Recurser returns the addr of the next DNS server to recurse a Msg to. Empty
+// returns signal that further recursion isn't possible or needed.
+type Recurser func(*dns.Msg) string
+
+// Recurse is the default Mesos-DNS Recurser which returns an addr only
```
jdef: can we clarify that the addresses returned here are nameserver host:port addresses?

exchanger/exchanger.go
```diff
@@ -0,0 +1,83 @@
...
+	"net"
+	"time"
+
+	"github.com/mesosphere/mesos-dns/logging"
+	"github.com/miekg/dns"
+)
+
+// Exchanger is an interface capturing a dns.Client Exchange method.
+type Exchanger interface {
+	Exchange(m *dns.Msg, addr string) (r *dns.Msg, rtt time.Duration, err error)
```
jdef: Can we document that addr is the host:port of a nameserver?

exchanger/exchanger_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/miekg/dns"
+)
+
+func TestRecurse(t *testing.T) {
+	for i, tt := range []struct {
+		*dns.Msg
+		want string
+	}{
+		{ // Authoritative with answers
+			Message(
```
jdef: if only it were this clean to build protobufs!

resolver/resolver.go
```diff
@@ -51,8 +47,25 @@ func New(version string, config records.Config) *Resolver {
...
+		timeout = time.Duration(config.Timeout) * time.Second
+	}
+
+	client := &dns.Client{
+		DialTimeout:  timeout,
+		ReadTimeout:  timeout,
+		WriteTimeout: timeout,
+	}
+
+	r.extResolver = exchanger.Decorate(client,
```
jdef: what do you think about another decorator that accepts a list of resolvers to iterate through? it would get rid of the loop in HandleNonMesos.

this "resolver builder" section could then be wrapped with an `if config.ExternalOn {}` and then in HandleNonMesos the check could if `if r.extResolver != nil {}`. Helps to make config just a little less leaky since HandleNonMesos would no longer reference it.

resolver/resolver.go
jdef: not sure why this was done this way. maybe someone on the ops team has some insight. i'm a little leery of changing behavior like this without a good reason.

jdef: i'd feel much better if we left `proto` as part of the interface for now. we can always yank it out in a later PR once we've verified that there no users depending on this. this PR seems like the wrong place to embed a behavioral change like this.

jdef:
filed a couple of comments/suggestions. overall :+1: :+1: :+1:

exchanger/exchanger.go
tsenart: In the Go stdlib, the name addr always means a host + port. I used the same naming here for that reason. With this in mind, do you still think it needs further explanation?

resolver/resolver.go
tsenart: Taking a protocol into account means that we have to create a new client on every request. If we think about it, clients can't possibly depend on the outgoing recursion transport protocol. That's entirely transparent to them. What I think we should do, instead, is to attempt to exchange the same message with both UDP and, if that fails, due to issues such as excessive payload size, TCP.

exchanger/exchanger.go
tsenart: Same as above.

resolver/resolver.go
tsenart: I attempted that before and realised that the `Exchanger` interface doesn't play well with that idea since it accepts the address to direct a DNS query to in its `Exchange` method.

+1 on the second suggestion though, which is still doable .

exchanger/exchanger_test.go
tsenart: We can make that happen :-)

resolver/resolver.go
jdef: ok, i'll buy that

exchanger/exchanger.go
jdef: yes, even something simple like changing s/addr/addr (host:port)/ in the
docs would be helpful.

tsenart:
@jdef: Back to you :-)

exchanger/exchanger.go
```diff
@@ -0,0 +1,114 @@
...
+	decorated := ex
+	for _, decorate := range ds {
+		decorated = decorate(decorated)
+	}
+	return decorated
+}
+
+// Multi returns an Exchanger which sequentially attempts the given Exchangers
+// until a successfull response is returned or an error in case of failure.
+func Multi(exs ...Exchanger) Exchanger {
```
jdef: :+1:

jdef:
lgtm. nice job.

tsenart:
Thanks James!

resolver/resolver.go
```diff
@@ -51,8 +47,34 @@ func New(version string, config records.Config) *Resolver {
...
+		clients[i] = exchanger.Decorate(&dns.Client{
+			Net:          proto,
+			DialTimeout:  timeout,
+			ReadTimeout:  timeout,
+			WriteTimeout: timeout,
+		}, exchanger.ErrorLogging(logging.Error))
+	}
+
+	r.extResolver = exchanger.Decorate(
+		exchanger.Multi(clients...),
```
tsenart: This approach to support https://tools.ietf.org/html/rfc5966 isn't the best. We should instead check the `Truncated` bit and re-try with TCP. Current status, reading: https://tools.ietf.org/html/rfc5966 and https://tools.ietf.org/html/rfc6891

tsenart:
@jdef: PTAL again

resolver/resolver.go
```diff
@@ -51,11 +47,43 @@ func New(version string, config records.Config) *Resolver {
...
+	for i, proto := range [...]string{"udp", "tcp"} { // See RFC5966
+		clients[i] = &dns.Client{
+			Net:          proto,
+			DialTimeout:  timeout,
+			ReadTimeout:  timeout,
+			WriteTimeout: timeout,
+		}
+	}
+	return exchanger.Decorate(
+		exchanger.While(truncated, clients...),
```

PR #246, opened 2015-09-08
generator keep consistency

njuettner:
To keep consistency I would suggest to use len(leader).

url concat is maybe confusing (at least for me), instead it could be replaced by fmt.Sprintf function


tsenart:
@njuettner: It's idiomatic go to compare strings to the empty strings instead of using the builtin `len`. Regarding the use of the  `fmt.Sprintf` function, I can get behind that, but it'd be even better to generate the URL string from a `net/url.URL` struct.

tsenart:
http://golang.org/pkg/net/url/

njuettner:
ok :), I only saw it on line 76

 		if len(masters) == 0 {
 			return sj, errors.New("no master")
 		}

yes you're right it makes more sense to use url.URL

tsenart:
The compiler optimizes `s == ""` to `len(s) == 0`.

records/generator.go
```diff
@@ -105,9 +106,13 @@ func (rg *RecordGenerator) findMaster(leader string, masters []string) (state.St
 // Loads state.json from mesos master
 func (rg *RecordGenerator) loadFromMaster(ip string, port string) (sj state.State) {
 	// REFACTOR: state.json security
-	url := "http://" + ip + ":" + port + "/master/state.json"
-
-	req, err := http.NewRequest("GET", url, nil)
+	master := fmt.Sprintf("%s:%s", ip, port)
```
tsenart: You have `net.JoinHostPort` for this exact use case. We should use that instead.

records/generator.go
```diff
@@ -105,9 +106,13 @@ func (rg *RecordGenerator) findMaster(leader string, masters []string) (state.St
...
+	u := &url.URL{
```
tsenart: No need for heap assignment. `s/&url/url/`

tsenart:
After addressing my comments, please rebase all commits into one.

PR #251, opened 2015-09-08
Allow hostnames for delegate resolvers

karlkfi:
### Use Case

As a deployment operator, I want to be able to chain DNS together with hostnames (not just IPs).

Note: masters can already be specified by host or domain name.
### Example Usage

Because deployments may contain more than just Mesos (e.g. Marathon), another DNS may be required that can resolve non-Mesos machines. In order to resolve Mesos, non-Mesos cluster machines, and internet domains, three DNS must be chained together. 

Example chain: `mesosdns` > `dnsdock` > `8.8.8.8`. 

Because Mesos-DNS depends on both zookeeper and mesos-master, it may not be ready as soon as the DNS it delegates to. So all three DNS need to be configured for each machine in the above chain, as well as for the DNS themselves.
#### Example docker-compose fragment

```
dnsdock:
  hostname: dnsdock
  image: tonistiigi/dnsdock
  command: [ "-nameserver=8.8.8.8" ]
  environment:
  - DNSDOCK_NAME=dnsdock
  - DNSDOCK_IMAGE=mesos
  volumes:
  - /var/run/docker.sock:/run/docker.sock
  dns:
  - 127.0.0.1
  - 8.8.8.8
mesosmaster1:
  hostname: master1.mesos.docker
  image: mesosphere/mesos:0.23.0-1.0.ubuntu1404
  entrypoint: [ "mesos-master" ]
  environment:
  - DNSDOCK_NAME=master1
  - DNSDOCK_IMAGE=mesos
  dns:
  - mesosdns.mesos.docker
  - dnsdock
  - 8.8.8.8
mesosdns:
  hostname: mesosdns.mesos.docker
  image: mesosphere/mesos-dns
  command: [ "/usr/bin/mesos-dns", "-v=2", "-config=/config.json" ]
  environment:
  - DNSDOCK_NAME=mesosdns
  - DNSDOCK_IMAGE=mesos
  volumes:
  - /var/tmp/config.json:/config.json
  links:
  - dnsdock
  - mesosmaster1
  dns:
  - localhost
  - dnsdock
  - 8.8.8.8
```
#### Example Mesos-DNS fragment

```
{
  "zk": "zk://zookeeper:2181/mesos",
  "masters": ["mesosmaster1:5050"],
  "domain": "mesos",
  "resolvers": ["dnsdock"],
}
```

### DIFF
--- a/records/validation.go
+++ b/records/validation.go
@@ -25,31 +25,27 @@ func validateMasters(ms []string) error {
 		}
 		//TODO(jdef) distinguish between intended hostnames and invalid ip addresses
 		if _, found := valid[m]; found {
-			return fmt.Errorf("duplicate master specified: %v", ms[i])
+			return fmt.Errorf("duplicate master specified: %q", ms[i])
 		}
 		valid[m] = struct{}{}
 	}
 	return nil
 }
 
-// validateResolvers checks that each resolver in the list is a properly formatted IP address.
-// duplicate resolvers in the list are not allowed.
-// returns nil if the resolver list is empty, or else all resolvers in the list are valid.
+// validateResolvers errors if there are duplicate resolvers, otherwise returns nil.
 func validateResolvers(rs []string) error {
 	if len(rs) == 0 {
 		return nil
 	}
 	ips := make(map[string]struct{}, len(rs))
 	for _, r := range rs {
-		ip := net.ParseIP(r)
-		if ip == nil {
-			return fmt.Errorf("illegal IP specified for resolver %q", r)
+		if r == "" {
+			return fmt.Errorf("empty resolver IP specified: %q", r)
 		}
-		ipstr := ip.String()
-		if _, found := ips[ipstr]; found {
-			return fmt.Errorf("duplicate resolver IP specified: %v", r)
+		if _, found := ips[r]; found {
+			return fmt.Errorf("duplicate resolver IP specified: %q", r)
 		}
-		ips[ipstr] = struct{}{}
+		ips[r] = struct{}{}
 	}
 	return nil
 }
--- a/records/validation.go
+++ b/records/validation.go
@@ -3,7 +3,13 @@ package records
 import (
 	"fmt"
 	"net"
+	"regexp"
+	"strconv"
 )
+
+const hostnamePattern = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
+
+var hostnameRegexp *regexp.Regexp
 
 // validateMasters checks that each master in the list is a properly formatted host:ip pair.
 // duplicate masters in the list are not allowed.
@@ -16,14 +22,14 @@ func validateMasters(ms []string) error {
 	for i, m := range ms {
 		h, p, err := net.SplitHostPort(m)
 		if err != nil {
-			return fmt.Errorf("illegal host:port specified for master %q", ms[i])
+			return fmt.Errorf("invalid host:port specified for master %q", ms[i])
 		}
-		// normalize ipv6 addresses
 		if ip := net.ParseIP(h); ip != nil {
-			h = ip.String()
-			m = h + "_" + p
+			// normalize ipv6 addresses
+			m = net.JoinHostPort(ip.String(), p)
+		} else if !isHostname(h) {
+			return fmt.Errorf("invalid hostname or IP specified for master %q", ms[i])
 		}
-		//TODO(jdef) distinguish between intended hostnames and invalid ip addresses
 		if _, found := valid[m]; found {
 			return fmt.Errorf("duplicate master specified: %q", ms[i])
 		}
@@ -39,11 +45,14 @@ func validateResolvers(rs []string) error {
 	}
 	ips := make(map[string]struct{}, len(rs))
 	for _, r := range rs {
-		if r == "" {
-			return fmt.Errorf("empty resolver IP specified: %q", r)
+		if ip := net.ParseIP(r); ip != nil {
+			// normalize ipv6 addresses
+			r = ip.String()
+		} else if !isHostname(r) {
+			return fmt.Errorf("invalid hostname or IP specified for resolver %q", r)
 		}
 		if _, found := ips[r]; found {
-			return fmt.Errorf("duplicate resolver IP specified: %q", r)
+			return fmt.Errorf("duplicate resolver specified: %q", r)
 		}
 		ips[r] = struct{}{}
 	}
@@ -67,4 +76,20 @@ func validateIPSources(srcs []string) error {
 	}
 
 	return nil
+}
+
+func isHostname(host string) bool {
+	if hostnameRegexp == nil {
+		hostnameRegexp = regexp.MustCompile(hostnamePattern)
+	}
+	matchSubstrings := hostnameRegexp.FindStringSubmatch(host)
+	if matchSubstrings == nil {
+		return false
+	}
+	tld := matchSubstrings[len(matchSubstrings)-1]
+	// TLD must not be all digits
+	if _, err := strconv.Atoi(tld); err == nil {
+		return false
+	}
+	return true
 }
--- a/records/validation.go
+++ b/records/validation.go
@@ -11,9 +11,9 @@ const hostnamePattern = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)
 
 var hostnameRegexp *regexp.Regexp
 
-// validateMasters checks that each master in the list is a properly formatted host:ip pair.
-// duplicate masters in the list are not allowed.
-// returns nil if the masters list is empty, or else all masters in the list are valid.
+// validateMasters checks that each element in the list is valid.
+// List elements must be either a valid IP:port or a valid hostname:port.
+// List must not contain duplicates (after IPv6 normalization).
 func validateMasters(ms []string) error {
 	if len(ms) == 0 {
 		return nil
@@ -38,7 +38,9 @@ func validateMasters(ms []string) error {
 	return nil
 }
 
-// validateResolvers errors if there are duplicate resolvers, otherwise returns nil.
+// validateResolvers checks that each element in the list is valid.
+// List elements must be either a valid IP or a valid hostname.
+// List must not contain duplicates (after IPv6 normalization).
 func validateResolvers(rs []string) error {
 	if len(rs) == 0 {
 		return nil
--- a/records/validation_test.go
+++ b/records/validation_test.go
@@ -10,10 +10,11 @@ func TestValidateMasters(t *testing.T) {
 		{[]string{}, true},
 		{[]string{""}, false},
 		{[]string{"", ""}, false},
-		{[]string{"a"}, false},
-		{[]string{"a:1234"}, true},
-		{[]string{"a", "b"}, false},
-		{[]string{"a:1", "b:1"}, true},
+		{[]string{"localhost"}, false},
+		{[]string{"localhost:1234"}, true},
+		{[]string{"example.com"}, false},
+		{[]string{"example.com:1234"}, true},
+		{[]string{"example.com:1234", "localhost:1234"}, true},
 		{[]string{"1.2.3.4"}, false},
 		{[]string{"1.2.3.4:5"}, true},
 		{[]string{"1.2.3.4.5"}, false},
@@ -21,6 +22,7 @@ func TestValidateMasters(t *testing.T) {
 		{[]string{"1.2.3.4", "1.2.3.4"}, false},
 		{[]string{"1.2.3.4:1", "1.2.3.4:1"}, false},
 		{[]string{"1.2.3.4:1", "5.6.7.8:1"}, true},
+		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b"}, false},
 		{[]string{"[2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b]:1"}, true},
 		{[]string{"[2001:db8:3c4d:15::1a2f:1a2b]:1"}, true},
 		{[]string{"[2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b]:1", "[2001:db8:3c4d:15::1a2f:1a2b]:1"}, false},
@@ -35,15 +37,17 @@ func TestValidateResolvers(t *testing.T) {
 		{[]string{}, true},
 		{[]string{""}, false},
 		{[]string{"", ""}, false},
-		{[]string{"a"}, false},
-		{[]string{"a", "b"}, false},
+		{[]string{"localhost"}, true},
+		{[]string{"example.com"}, true},
+		{[]string{"example.com", "localhost"}, true},
+		{[]string{"localhost", "localhost"}, false},
 		{[]string{"1.2.3.4"}, true},
-		{[]string{"1.2.3.4.5"}, false},
+//		{[]string{"1.2.3.4.5"}, false},
 		{[]string{"1.2.3.4", "1.2.3.4"}, false},
 		{[]string{"1.2.3.4", "5.6.7.8"}, true},
 		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b"}, true},
 		{[]string{"2001:db8:3c4d:15::1a2f:1a2b"}, true},
-		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b", "2001:db8:3c4d:15::1a2f:1a2b"}, false},
+//		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b", "2001:db8:3c4d:15::1a2f:1a2b"}, false},
 	} {
 		validate(t, i+1, tc, validateResolvers)
 	}
@@ -59,8 +63,8 @@ func validate(t *testing.T, i int, tc validationTest, f func([]string) error) {
 	case (err == nil && tc.valid) || (err != nil && !tc.valid):
 		return // valid
 	case tc.valid:
-		t.Fatalf("test %d failed, unexpected error validating resolvers %v: %v", i, tc.in, err)
+		t.Fatalf("test %d failed, unexpected error validating resolvers %q: %v", i, tc.in, err)
 	default:
-		t.Fatalf("test %d failed, expected validation error for resolvers(%d) %v", i, len(tc.in), tc.in)
+		t.Fatalf("test %d failed, expected validation error for resolvers %#v", i, tc.in)
 	}
 }
--- a/records/validation_test.go
+++ b/records/validation_test.go
@@ -18,7 +18,8 @@ func TestValidateMasters(t *testing.T) {
 		{[]string{"1.2.3.4"}, false},
 		{[]string{"1.2.3.4:5"}, true},
 		{[]string{"1.2.3.4.5"}, false},
-		{[]string{"1.2.3.4.5:6"}, true}, // no validation of hostnames
+		{[]string{"1.2.3.4.5:6"}, false}, // TLD must not be all digits
+		{[]string{"1.2.3.4.5x:6"}, true}, // custom TLD may start with a digit
 		{[]string{"1.2.3.4", "1.2.3.4"}, false},
 		{[]string{"1.2.3.4:1", "1.2.3.4:1"}, false},
 		{[]string{"1.2.3.4:1", "5.6.7.8:1"}, true},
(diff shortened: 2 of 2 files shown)


karlkfi:
```
$ GOPATH=~/workspace/mesos-dns make test
--> Testing mesos-dns
?   	github.com/mesosphere/mesos-dns	[no test files]
?   	github.com/mesosphere/mesos-dns/dnstest	[no test files]
ok  	github.com/mesosphere/mesos-dns/exchanger	0.009s
?   	github.com/mesosphere/mesos-dns/logging	[no test files]
ok  	github.com/mesosphere/mesos-dns/records	0.012s
ok  	github.com/mesosphere/mesos-dns/records/labels	7.363s
ok  	github.com/mesosphere/mesos-dns/records/state	0.007s
ok  	github.com/mesosphere/mesos-dns/resolver	0.222s
?   	github.com/mesosphere/mesos-dns/tools	[no test files]
ok  	github.com/mesosphere/mesos-dns/util	0.008s
```

records/validation_test.go
```diff
@@ -35,15 +37,17 @@ func TestValidateResolvers(t *testing.T) {
...
 		{[]string{"", ""}, false},
-		{[]string{"a"}, false},
-		{[]string{"a", "b"}, false},
+		{[]string{"localhost"}, true},
+		{[]string{"example.com"}, true},
+		{[]string{"example.com", "localhost"}, true},
+		{[]string{"localhost", "localhost"}, false},
 		{[]string{"1.2.3.4"}, true},
-		{[]string{"1.2.3.4.5"}, false},
+//		{[]string{"1.2.3.4.5"}, false},
```
jdef: this needs to be uncommented, 1.2.3.4.5 is not a valid host name

records/validation_test.go
```diff
@@ -35,15 +37,17 @@ func TestValidateResolvers(t *testing.T) {
...
 		{[]string{"1.2.3.4", "1.2.3.4"}, false},
 		{[]string{"1.2.3.4", "5.6.7.8"}, true},
 		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b"}, true},
 		{[]string{"2001:db8:3c4d:15::1a2f:1a2b"}, true},
-		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b", "2001:db8:3c4d:15::1a2f:1a2b"}, false},
+//		{[]string{"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b", "2001:db8:3c4d:15::1a2f:1a2b"}, false},
```
jdef: this needs to be uncommented, duplicate IPs are not allowed

records/validation_test.go
karlkfi: How do you validate both 1.2.3.4.5 AND hostnames? I don't know of any way to do it.

records/validation.go
```diff
@@ -25,31 +25,27 @@ func validateMasters(ms []string) error {
...
 		}
 		valid[m] = struct{}{}
 	}
 	return nil
 }
 
-// validateResolvers checks that each resolver in the list is a properly formatted IP address.
-// duplicate resolvers in the list are not allowed.
-// returns nil if the resolver list is empty, or else all resolvers in the list are valid.
+// validateResolvers errors if there are duplicate resolvers, otherwise returns nil.
```
jdef: this refactoring is too naive:
- I want duplicate IP's to error
- I want syntactically invalid hostnames to error

records/validation_test.go
karlkfi: I assume you mean duplicate normalized IPs are not allowed.

jdef: correct. this test case should not have been commented out

karlkfi: Agreed. I misunderstood what it was testing.

records/validation_test.go
jdef: try to parse as an IP (there's a golang lib func for this); if it's not an
IP then validate hostname syntax (we'll probably have to write a func for
this).

karlkfi: Sure, it sounds plausible. I'm just not sure a good hostname matcher exists... I've seen a lot fo different ones and they all have problems. It all goes down the tubes when you consider recent changes that allow custom domain suffixes and non-ascii utf-8 urls.

tsenart:
@karlkfi: Why can't one know the IP address of dnsdock? How is mesos-dns supposed to resolve a hostname to the IP address itself?

jdef:
An alternative would be to attempt to use the host's resolver during
validation and swap out resolver hostnames for IPs. Then we maintain that
the []resolvers slice used by the Exchanger only contains IPs. The rule
being that user-specified resolver hostnames must be resolvable at
mesos-dns start time, otherwise :(

tsenart:
It doesn't feel right that a DNS server needs to rely on the availability of another DNS server to be able to recurse messages to yet another DNS server. Resolver configuration ought to not introduce another failure mode which is propagated into runtime.

@karlkfi: Why can't we have a static IP / port for dnsdock and use that instead?

karlkfi:
@tsenart The IP of dnsdock is not known until docker assigns it (dhcp-like). Since dnsdock and mesosdns are deployed at the same time, there's no way to know what the IP is at "compile time" when I created the mesos-dns config.json. Docker injects the hostnames of linked containers, because links also imply dependency. So mesosdns come sup after dnsdock is up and knows that `dnsdock` resolves to the right IP.

If there were a way to configure mesos-dns by command line argument it might be possible to pre-resolve the hostname IP, but it's annoying to do because it also means the entrypoint or command need to be a `bash -c` string, because docker otherwise wont run the command in a shell.

karlkfi:
> Resolver configuration ought to not introduce another failure mode which is propagated into runtime.

This isn't a new failure mode. The resolver IP already may or may not be up/ready/accessible. It's already possible to fail delegation. This is no different. `failed to reach '8.8.8.8'` vs `failed to resolve 'dnsdock'`.

jdef:
I'd like to be able to maintain the invariant that post-validation the
Config.Resolvers object contain only IPs. I think we can do this, and also
allow users to specify hostnames in the dns config file (we can just
translate them pre- or during resolver validation).

karlkfi:
`1.2.3.4.5` is almost a valid hostname... 

The only rule I can find that disallows it is that TLDs may not be exclusively made up of digits. I guess I can validate this, but I have yet to see a commonly accepted regex for validate all the following edge cases correctly.

### Hostnames may start with a digit

  > The original specification of hostnames in RFC 952, mandated that labels could not start with a digit or with a hyphen, and must not end with a hyphen. However, a subsequent specification (RFC 1123) permitted hostname labels to start with digits. No other symbols, punctuation characters, or white space are permitted. 

<https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_host_names>

### Hostnames may start with a digit (but not be exclusively digits)

> An additional rule requires that top-level domain names should not be all-numeric.

<https://en.wikipedia.org/wiki/Domain_Name_System#Domain_name_syntax>

### Multiple sub-domains are allowed

> In theory this subdivision can go down to 127 levels deep, and each DNS label can contain up to 63 characters, as long as the whole domain name does not exceed a total length of 255 characters. But in practice most domain registries limit at 253 characters.

<https://en.wikipedia.org/wiki/Subdomain#Overview>

karlkfi:
Modifying values at validation-time is a good way to get really confusing runtime errors when the value doesn't match the input value.

tsenart:
@karlkfi: I can get behind accepting hostnames for resolvers as long as they can be resolved to IPs at start-up time and re-used throughout the program lifetime. Delaying the resolution of the resolver name to run-time *does* introduce a failure mode that could manifest through lack of DNS availability, mis-configuration, etc...

tsenart:
You shouldn't really modify the values in a validation function. Another function is needed for such.

karlkfi:
I don't think they need to be resolved during bootstrap. In fact, if they do then they aren't resilient to intermediate DNS downtime. If the user wants such a guarantee, they can resolve the IP themselves. The whole point here is to delay resolution, because it can't be done at configuration time.

tsenart:
> In fact, if they do then they aren't resilient to intermediate DNS downtime. 

Would you please elaborate on this? I'm a bit confused. :/

jdef:
Agreed. We can do hostname lookups before we validate the resolvers slice,
and keep modification separate from validation.

FWIW I've used hostname's for delegates in bind before, specifying a
delegate hostname in the bind config, and specifying a hostname-to-IP
mapping for the delegate in /etc/hosts ... which is essentially the same
thing that @karlkfi is trying to achieve. so, overall, I think it's a valid
use case.

jdef:
Couple of points here:
- Configuration time != container run time
- "if the user wants a guarantee ... resolve [it] themselves"; this is what
docker is supposed to be doing, right, updating /etc/hosts? that would let
mesos-dns resolve the name at startup time, yes?

tsenart:
@jdef: What I meant by the *mis-configuration* failure mode is exactly this one: What happens when `/etc/hosts` changes under the covers by operator error through some configuration management system?

karlkfi:
> this is what docker is supposed to be doing, right, updating /etc/hosts? that would let mesos-dns resolve the name at startup time, yes?

For **hostnames** under the specified use case, yes. But not for **domain names** that would need to be resolved by an intermediate DNS.

karlkfi:
> What happens when /etc/hosts changes under the covers by operator error through some configuration management system?

Modification of `/etc/hosts/` at runtime is a perfectly legitimate use case. Granted it could be accidental, but we have no way of knowing that.  If `/etc/hosts/` changes, the resolver should continue delegating to the configured resolver hostname, regardless of where it currently points.

tsenart:
> Modification of /etc/hosts/ at runtime is a perfectly legitimate use case. Granted it could be accidental, but we have no way of knowing that. If /etc/hosts/ changes, the resolver should continue delegating to the configured resolver hostname, regardless of where it currently points.

That's just global, mutable state. Not something I'm comfortable with dealing with if it can be avoided.

jdef:
then that configuration management system should be prepared to restart
impacted services, like mesos-dns

karlkfi:
The OS already has plenty of global state. That's why we're delegating resolution to the OS already. Golang checks `/etc/hosts`, `/etc/hostnames`, and `/etc/resolv.conf` under the covers.

jdef:
we have a couple of options here; if the resolver-hostname is not actually
resolvable at mesos-dns start-time:
- abort; fatal error .. or
- log the error and skip the resolver
- make resolution more complicated, introduce a lazy resolver lookup so
that we get the IP the first time it's needed (still racy, and this adds
too much complication so it's a bad option that we shouldn't really
consider but I felt compelled to include for completeness)

You highlighted **domain names** so you're thinking that an intermediate
DNS is going to resolve the hostname of a resolver with an external domain
name that it (the internal DNS server) can't resolve?! dragons...
everywhere. don't do it that way.

karlkfi:
We don't need to introduce a lazy resolver. Lazy resolution already functions with the current impl.

To throw more fire on the validation:

> a host domain name is now allowed to begin with a digit and could legally be entirely numeric

<http://serverfault.com/questions/638260/is-it-valid-for-a-hostname-to-start-with-a-digit/638270#638270>

That said, I'm close to a validator that handles all the cases I've documented.

jdef:
>  *However, a valid host name can never have the dotted-decimal form
#.#.#.#, since at least the highest-level component label will be not
all-numeric*

discordianfish:
I would either support only IPs or leave the resolution to Exchange(): https://github.com/mesosphere/mesos-dns/blob/31f5539636a8bf591bf702fe686f63b37e690fea/resolver/resolver.go#L307
I would avoid adding the complexity of resolving and managing resolvers.
If people want to use names for their resolver configuration, they are probably okay with the additional roundtrip.

In general though I agree with @tsenart and wouldn't run it in production like that. For the Docker use case you can also use the link environment variables and use them to write the config before running mesos-dns. This can be done in bash, just need to make sure to exec mesos-dns to exit bash and only mesos-dns runs in the container. Or - shameless plug - https://github.com/docker-infra/reefer which I created to solve similar issues.

karlkfi:
If mesos-dns had command-line arguments to set zks, masters, and resolvers I wouldn't be pushing quite as hard on this. But I do think this is the right way to allow more flexibility in configuration. If the user doesn't want the added risk or cost, they can configure using IPs.

That said, I've now pushed an update that validates that the input is **either a valid IP or a valid hostname**.  I'm not 100% sure that the regex covers all possible use cases, but I did add a couple test cases to improve our confidence. 

I also added the same hostname validate to the master validation, which had a TODO to do so.

tsenart:
In order to solve this problem without sacrificing the properties and invariants I want this code base to maintain, I see two possible solutions:

1. Resolve names at start-up time *outside* Mesos-DNS and write IPs to the config. As @discordianfish mentioned, this seems to be possible by wrapping Mesos-DNS execution in a BASH script or similar. **I vote for this one.**
2. Resolve names at start-up time *inside* Mesos-DNS.

Deferring resolution is *not* an option due to a few reasons which I'm synthesising here:

1. Go 1.4 and Go 1.5 have a global `sync.Mutex` protecting the static host cache read from `/etc/hosts`. This can hurt our throughput with large numbers of concurrent queries.

2. Go 1.4 and Go 1.5 have a global cache static host cache timeout of 5 minutes. No changes to `/etc/hosts` are available for a maximum of that period. As for availability of a recursing DNS server goes, this can go *pretty* bad. If operators need to change configuration of static hosts, they'd have to restart Mesos-DNS for it to pickup the new static hosts entries immediately.

3. In case the given name isn't found in the static hosts cache, a DNS query will be issued against the system configured DNS servers (read from `/etc/resolv.conf`), only to find the IP of the resolver we want to forward our query to. Hence, for such simple query, we have to perform one extra round-trip (or even two, in case of truncation where we have to fallback to TCP), which besides hurting performance, can hurt our availability.

4. From an operational simplicity standpoint, having configuration be static throughout the lifetime of a process, is a major advantage. Side-effect free programs are invariantly easier to debug than ones with lots of side-effects and global mutable state.

In sum, in my view, the extra flexibility of deferred resolver resolution isn't worth the costs. I'll also be revisiting these same conclusions in other places in the codebase where we're relying on the system resolver!

jdef:
It would be nice to not have to rely on BASH to get resolver hostname
resolution functionality. It's not that hard to add to the Go code, the
changes are fairly isolated, and it's much more convenient to package
mesos-dns as a binary vs. binary+BASH

tsenart:
@jdef: Fair point. As I said, I'm OK with either of those two, your comment made me reconsider which one I prefer :-)

karlkfi:
> Go 1.4 and Go 1.5 have a global sync.Mutex protecting the static host cache read from /etc/hosts.

This makes me sad. We should submit a PR to make it a RWMutex.

> Go 1.4 and Go 1.5 have a global static host cache timeout of 5 minutes.

Is this hard coded?

> In case the given name isn't found in the static hosts cache, a DNS query will be issued against the system configured DNS servers (read from /etc/resolv.conf).

I don't think this is a downside. If the user configured a hostname then it seem clear they want it resolved on-demand, not up front.

> From an operational simplicity standpoint, having configuration be static throughout the lifetime of a process, is a major advantage. Side-effect free programs are invariantly easier to debug than ones with lots of side-effects and global mutable state.

I completely agree that static configurations are simpler to deal with. I don't agree that that makes mutable config always a bad idea. And I definitely do not agree with your application of that reasoning as sufficient cause to make mesos-dns harder to use. (Are you really gonna write a bash script to templatize the json config and write it to file so you can pass it to mesos-dns? That's terrible!) 

A configured hostname IS static. It's just not a static IP. If the user is willing to accept the costs that come with on-demand resolution, why would you not want to give them that flexibility?

Configuring a non-IP implies deliberately asking mesos-dns to do on-demand resolution based on a "normal" hostname resolution procedure. Hostname resolution normally includes caching. Ideally the cache timeout would be configurable, but it's not unreasonable to punt on that as a know caveat. 

Thank you for notifying me of the downsides of this configuration. but I still don't care. None of those reasons make me want to not use a hostname in a testing environment.

I would rather completely retract my PR than implement on-bootstrap hostname resolution. I think that option is like lying to the user, taking their hostname and discarding it. Not only that, but it requires extra dev work to make the UX worse.

jdef:
oops, meant to cc @spacejam

tsenart:
> This makes me sad. We should submit a PR to make it a RWMutex.

See [more](https://sourcegraph.com/github.com/golang/go@master/.tree/src/net/hosts.go#startline=32&endline=32).

> Is this hard coded?

Unfortunately, [yes](https://sourcegraph.com/github.com/golang/go@master/.tree/src/net/hosts.go#startline=12&endline=12).

> I don't think this is a downside. If the user configured a hostname then it seem clear they want it resolved on-demand, not up front.

You quoted just a portion of the whole argument. There's more to it than this.

> (Are you really gonna write a bash script to templatize the json config and write it to file so you can pass it to mesos-dns? That's terrible!)

[As said before](https://github.com/mesosphere/mesos-dns/pull/251#issuecomment-138933759), I'm tending to the second described solution (resolution at start-up time), not the first which you're rightfully critiquing.

> A configured hostname IS static. It's just not a static IP. If the user is willing to accept the costs that come with on-demand resolution, why would you not want to give them that flexibility?

Yes, a hostname is static just like the number 3 is static. However, what the hostname resolves to, isn't, in this case. A user wouldn't always understand the trade-offs and costs involved with this decision and I'd rather sacrifice flexibility and feature "richness" for operational and architectural simplicity and scalability.

> Thank you for notifying me of the downsides of this configuration. but I still don't care. None of those reasons make me want to not use a hostname in a testing environment.

> I would rather completely retract my PR than implement on-bootstrap hostname resolution. I think that option is like lying to the user, taking their hostname and discarding it. Not only that, but it requires extra dev work to make the UX worse.

Such resolution on bootstrap behaviour can be as well documented as the caveats you're willing to accept. That wouldn't be lying to the user IMO. I'm curious as to how you feel the UX is worse with that option though.

tsenart:
> /cc @tyler since you were wondering how Go clients cache DNS lookup results

This isn't really about caching all DNS results, only static hosts defined in `/etc/hosts` on UNIX based OSes.

karlkfi:
I don't think I've ever used an application that accepted a hostname, looked it up once, discarded the hostname and saved the resolved IP. It doesn't match expectations, and UX is all about reducing cognitive overhead of having to remember things that don't match your expectations. If something "just works" that means it works the way you expect it to, so you don't need to read the manual or bend over backwards to configure it.

Take for example, Mesos. If you configure a hostname, it announces that. It doesn't resolve it to an IP and then announce the IP. It might resolve it to an IP to decide what IP to listen on, but it only listens once. No one expects it to start listening on a new IP after startup, but they DO expect other applications to be able to find it on its announced hostname, even if its IP changes, as long as the hostname does not. It may use a simple host for that, or it may use a full domain name, or it may use a hostname that gets resolved with DNS search. Those are all normal things people expect of an address that they don't expect of an IP.

tsenart:
> I don't think I've ever used an application that accepted a hostname, looked it up once, discarded the hostname and saved the resolved IP. It doesn't match expectations, and UX is all about reducing cognitive overhead of having to remember things that don't match your expectations. If something "just works" that means it works the way you expect it to, so you don't need to read the manual or bend over backwards to configure it.

That's a very good definition of UX and I fully agree with it. However, this project is a DNS server, not an  app of other sorts. Hence, different reasoning is being applied here due to the nature of the project.

> Take for example, Mesos. If you configure a hostname, it announces that. It doesn't resolve it to an IP and then announce the IP. It might resolve it to an IP to decide what IP to listen on, but it only listens once.

I wouldn't take that as a good example. A hostname can have multiple interfaces and hence *multiple* IPs but only one of those will be usually listening on the socket. Binding a socket to such a name isn't something I'd consider a good practice.

> No one expects it to start listening on a new IP after startup, but they DO expect other applications to be able to find it on its announced hostname, even if its IP changes, as long as the hostname does not.

Yeah, sure, that's the DNS infrastructure you're describing and what something like Mesos-DNS is meant to provide.

> It may use a simple host for that, or it may use a full domain name, or it may use a hostname that gets resolved with DNS search. Those are all normal things people expect of an address that they don't expect of an IP.

All valid expectations which I'd support if this wasn't a DNS server.

karlkfi:
Have it your way.

tsenart:
@karlkfi: So, you don't intend to implement start-up time resolution? What are you going to do instead? This issue isn't really closed as you still have the problem. Please advise otherwise.

karlkfi:
I have my own branch that functions just fine. Since no one is packaging mesos-dns, I've had to do that on a branch too. For the time being that suffices, because I was only building a test deployment anyway. If I have to maintain my own fork to do so, so be it. At some point I'll probably have to completely rewrite the way mesos-dns does configuration, because it's a PITA to require a json file on disk when running mesos-dns in a container. At that point I expect I'll be able to use an external ip resolver that I'll have to bake into the docker image.

cmaloney:
> I don't think I've ever used an application that accepted a hostname, looked it up once, discarded the hostname and saved the resolved IP. It doesn't match expectations, and UX is all about reducing cognitive overhead of having to remember things that don't match your expectations. If something "just works" that means it works the way you expect it to, so you don't need to read the manual or bend over backwards to configure it.

Nginx does it for all DNS in the config. HAProxy does as well. Old Java did that always, new java does it [unless you configure it otherwise] (http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/java-dg-jvm-ttl.html). Definitely not a new novel behavior.

tsenart:
@karlkfi: I'll be working on flag based configuration tomorrow. I hope that'll help your cause.

jdef:
@tsenart shifting to flag-based from file-based is something that will
impact DCOS. Please reach out to that team.

tsenart:
@jdef: Thanks, I'll certainly reach out. Please note, though, that as described in #257, we'll have one version with support for both file and flags, before deprecating file.

jdef:
@koziraki was a pretty staunch advocate of the current file-based config.
he had concerns about the "command runner" (who executes mesos-dns) having
to know the configuration details of mesos-dns at the time the binary was
executed, vs just pointing mesos-dns at a location that someone else deals
with (the config file).

That said, I'm generally not a fan of having tons of flags for a binary,
but I understand the arguments for flags and/or envvar in the context of
containerization. I'd actually lean towards envvar to either override, or
provide fallback values for, values set in an optional config file (and
don't implement the flags approach)-- that keeps about the same distance
between configuration and the execution of the binary (at least, more
distance than flags) and also allows for easier containerization.

tsenart:
@jdef: Can you comment on the relevant issue please? Just for tracking.
Thanks

jdef:
Closing since this PR is quite stale and doesn't seem to be in high demand.

PR #253, opened 2015-09-09
using net url instead of concat

njuettner:
My apologies, I had to close the old pull request (https://github.com/mesosphere/mesos-dns/pull/246), cause some issues with squash.

This basically changes what we've discussed.

Instead of concatenation @tsenart suggested to use net.JoinHostPort and URL struct as well.

### DIFF
--- a/records/generator.go
+++ b/records/generator.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 
@@ -109,9 +110,18 @@ func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
 // Loads state.json from mesos master
 func (rg *RecordGenerator) loadFromMaster(ip string, port string) (sj state.State) {
 	// REFACTOR: state.json security
-	url := "http://" + ip + ":" + port + "/master/state.json"
 
-	req, err := http.NewRequest("GET", url, nil)
+	u := url.URL{
+		Scheme: "http",
+		Host:   net.JoinHostPort(ip, port),
+		Path:   "/master/state.json",
+	}
+
+	req, err := http.NewRequest("GET", u.String(), nil)
+	if err != nil {
+		logging.Error.Println(err)
+	}
+
 	req.Header.Set("Content-Type", "application/json")
 
 	client := &http.Client{}


records/generator.go
```diff
@@ -105,9 +106,15 @@ func (rg *RecordGenerator) findMaster(leader string, masters []string) (state.St
 // Loads state.json from mesos master
 func (rg *RecordGenerator) loadFromMaster(ip string, port string) (sj state.State) {
 	// REFACTOR: state.json security
-	url := "http://" + ip + ":" + port + "/master/state.json"
 
-	req, err := http.NewRequest("GET", url, nil)
+	master := net.JoinHostPort(ip, port)
```
tsenart: This variable is unnecessary. You can inline it in the url struct instantiation.

records/generator.go
```diff
@@ -105,9 +106,14 @@ func (rg *RecordGenerator) findMaster(leader string, masters []string) (state.St
...
-	url := "http://" + ip + ":" + port + "/master/state.json"
 
-	req, err := http.NewRequest("GET", url, nil)
+	u := url.URL{
+		Scheme: "http",
+		Host:   net.JoinHostPort(ip, port),
+		Path:   "/master/state.json",
+	}
+
+	req, err := http.NewRequest("GET", u.String(), nil)
```
jdef: since we're mucking around in here, mind logging the `err`?

njuettner:
You're totally right, I added it.

tsenart:
@njuettner: Coo, thanks. Can you squash your commits into one and make sure your commits message complies with http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html?

tsenart:
@njuettner: There seems to be some merge conflict. Can you resolve this on your side so that we can get it merged?

tsenart:
Please squash your commits into one.

tsenart:
Thanks! LGTM

karlkfi:
It's not new for this change, but why doesn't `loadFromMaster` return an error? It doesn't make sense to log and continue. It's almost guaranteed to panic later on down the line...

tsenart:
@karlkfi: You're absolutely right. Unfortunately, there are many areas of the codebase where you'll find this kind of surprises. Lots of technical debt to address.

PR #255, opened 2015-09-10
Ignore Coveralls API failures in Circle CI

tsenart:
This commit will go hand in hand with disabling the protected branch check for Coveralls since external forks wouldn't be able to pass the them due to security restrictions in CirceCI which doesn't provide these builds with the required private environment variables.

I trust we'll always look at how the coverage changes on every PR and take that into consideration on reviews.


s-urbaniak:
Are you referring to the failure in https://github.com/mesosphere/mesos-dns/pull/253? I guess I can live with that as long as the unit tests let a given PR fail, fortunately this process is independent of any tokens/provider.

s-urbaniak:
LGTM

tsenart:
Closed by 64f21a0

PR #261, opened 2015-09-14
updated to latest mesos-go, adds support for JSON MasterInfo in ZK

### LINKED ISSUE #228: Change in Mesos leader info in ZK may break Mesos-DNS
Mesos 0.24 will follow this convention for the leading master info in ZK: https://github.com/mesos/commons/blob/master/docs/leader_discovery.rst
https://issues.apache.org/jira/browse/MESOS-2340
This will break existing Mesos-DNS code which assumes the old protobuf. 

We need to update the code to handle correctly both 0.24 and older versions :(

jdef:
tested against mesos:tip, seems to do the trick


jdef:
cc @cmaloney @nqn

tsenart:
LGTM

Godeps/_workspace/src/github.com/mesos/mesos-go/mesosutil/constants.go
```diff
@@ -2,5 +2,5 @@ package mesosutil
 
 const (
 	// MesosVersion indicates the supported mesos version.
-	MesosVersion = "0.20.0"
+	MesosVersion = "0.23.0"
```
nqn: Don't want it to be 0.24.0?

tsenart: We'll have to update that in mesos-go.

tsenart:
Closing this to be re-opened with https://github.com/mesos/mesos-go/pull/178 vendored.

tsenart:
I'll also add a CHANGELOG, tag things properly and do a Github release.

PR #263, opened 2015-09-15
Release v0.2.0

### LINKED ISSUE #228: Change in Mesos leader info in ZK may break Mesos-DNS

tsenart:
This change set releases **v0.2.0**. Once merged, I'll tag master with `mesos-0.24.0` and `v0.2.0`. This will also be the first release to include cross-compiled binaries in Github releases which shall be our official distribution mechanism from now on.

Fixes #228


jdef:
LGTM

See that you're building w/ go1.5 - that's cool. We've had complains about breaking 1.3 and 1.4 compat in the past. Not sure how relevant that is for DCOS builds. May want to check with the DCOS image team.

tsenart:
@jdef: I'm going to change dcos-image to use the cross-compiled binaries instead of building from source.

tsenart:
And as mentioned above, our official distribution method shall be Github releases with cross-compiled binaries from now on. I've filed https://github.com/mesosphere/mesos-dns/issues/264 to do this today.

PR #266, opened 2015-09-16
Added logic to read NetworkInfo from /state.json.

ConnorDoyle:
## Overview

Mesos 0.25.0 will include a new way for network isolation modules to communicate assigned IP addresses assigned to containers at executor launch time.  Where previously Mesos-DNS picked up the values of some special labels appended to a task status, Mesos 0.25.0 introduces a new structure called `NetworkInfo`.

## Change summary

This change preserves the previous behavior of reading special labels to ease transition to users who relied on that method of IP address discovery.  However, the information in the new `NetworkInfo` structure should take precedence if present.

## Additional references

- For general information on the `NetworkInfo` message and its intended use, see the public [networking design document](https://docs.google.com/document/d/17mXtAmdAXcNBwp_JfrxmZcQrs7EO6ancSbejrqjLQ0g).
- For a different angle on how Framework developers will interact with new network isolation strategies in Mesos, see the public [Framework writer’s guide to IP-per-container](https://docs.google.com/document/d/13mXNziHQPkmHdH6g2srtE2k0PaJSIWuBvRgmJ7Jvqh4).
- For implementation details around `NetworkInfo` and its representation in the Mesos master state endpoint, see [Review Request 38367: Added NetworkInfo message to ContainerInfo and TaskStatus](https://reviews.apache.org/r/38367) by @karya0 .

### DIFF
--- a/records/state/state.go
+++ b/records/state/state.go
@@ -123,7 +123,7 @@ func (t *Task) containerIP(src string) string {
 			// TODO(CD): create AAAA records if the address is IPv6
 			latestContainerIP = status.ContainerStatus.NetworkInfos[0].IPAddress
 			latestTimestamp = status.Timestamp
-			break
+			continue // Skip label inspection step below for this element
 		}
 
 		// next, fall back to the docker-inspect label
--- a/records/state/state.go
+++ b/records/state/state.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"bytes"
+	"net"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


records/state/state.go
```diff
@@ -103,7 +116,18 @@ func (t *Task) containerIP(src string) string {
 			continue
 		}
 
-		// find the latest docker-inspect label
+		// first try to extract the address from the NetworkInfo structure
+		if len(status.ContainerStatus.NetworkInfos) > 0 {
+			// TODO(CD): handle multiple NetworkInfo objects
+			// TODO(CD): only create A records if the address is IPv4
+			// TODO(CD): create AAAA records if the address is IPv6
+			latestContainerIP = status.ContainerStatus.NetworkInfos[0].IPAddress
```
jdef: may be interesting to filter on label here at some point in the future when there are multiple IPs

ConnorDoyle: I'm not even sure what Mesos-DNS should do when there are multiple addreses assigned.  I think we'd need conventions for handling additional metadata to give us clues (e.g. labels on NetworkInfo, etc.)

records/state/state.go
```diff
@@ -46,9 +46,22 @@ type Label struct {
 
 // Status holds a task status as defined in the /state.json Mesos HTTP endpoint.
 type Status struct {
-	Timestamp float64 `json:"timestamp"`
-	State     string  `json:"state"`
-	Labels    []Label `json:"labels,omitempty"`
+	Timestamp       float64         `json:"timestamp"`
+	State           string          `json:"state"`
+	Labels          []Label         `json:"labels,omitempty"`
+	ContainerStatus ContainerStatus `json:"container_status"`
```
jdef: nit: this is an optional field, should probably apply `omitempty`

records/state/state.go
```diff
@@ -46,9 +46,22 @@ type Label struct {
...
+}
+
+// ContainerStatus holds container metadata as defined in the /state.json
+// Mesos HTTP endpoint.
+type ContainerStatus struct {
+	NetworkInfos []NetworkInfo `json:"network_infos"`
```
jdef: omitempty?

records/state/state.go
```diff
@@ -46,9 +46,22 @@ type Label struct {
...
+}
+
+// NetworkInfo holds a network address resolution as defined in the
+// /state.json Mesos HTTP endpoint.
+type NetworkInfo struct {
+	IPAddress string `json:"ip_address"`
```
jdef: omitempty?

records/state/state.go
```diff
@@ -103,7 +116,18 @@ func (t *Task) containerIP(src string) string {
...
 
-		// find the latest docker-inspect label
+		// first try to extract the address from the NetworkInfo structure
+		if len(status.ContainerStatus.NetworkInfos) > 0 {
+			// TODO(CD): handle multiple NetworkInfo objects
+			// TODO(CD): only create A records if the address is IPv4
+			// TODO(CD): create AAAA records if the address is IPv6
+			latestContainerIP = status.ContainerStatus.NetworkInfos[0].IPAddress
+			latestTimestamp = status.Timestamp
+			break
```
jdef: might be breaking too early here if there are multiple TASK_RUNNING status entries and they're out of order w/ respect to timestamps (this break has a different effect than the one in the label-driven loop below)

jdef:
Great writeup! Thanks for all the context. Couple of minor comments but overall looking good. LGTM after comments addressed.

records/state/state.go
jdef: Right, it's not clear to me either how mesos-dns should respond in
situations like that. It also just occurred to me that it might be nice to
include mention of this new behavior somewhere in the mesos-dns docs pages.

records/state/state.go
ConnorDoyle: You're right, I misread the following break statement.  I did wonder whether we were assuming sorted updates, and if so why bother to check the timestamp.  That makes more sense.  This should be `continue` instead to jump over the label inspection (to maintain preference for NetworkInfo addresses), but still consider the other elements.  Good catch.

ConnorDoyle:
Hey @tsenart, could you PTAL and merge if you don't find anything objectionable?

records/state/state.go
tsenart: I personally don't like to use `omitempty` because when a field is absent in a JSON payload, one can't tell the difference between the field not existing and being empty for a particular instance.

records/state/state.go
```diff
@@ -46,9 +46,22 @@ type Label struct {
...
+// ContainerStatus holds container metadata as defined in the /state.json
+// Mesos HTTP endpoint.
+type ContainerStatus struct {
+	NetworkInfos []NetworkInfo `json:"network_infos,omitempty"`
+}
+
+// NetworkInfo holds a network address resolution as defined in the
+// /state.json Mesos HTTP endpoint.
+type NetworkInfo struct {
+	IPAddress string `json:"ip_address,omitempty"`
```
tsenart: How do we know if the `IPAddress` is v4 or v6? Are all clients supposed to try parsing one and then fallback?

ConnorDoyle: Yes, exactly.  Using a string here was seen as a better alternative than trying to come up with a clever packed representation.

records/state/state.go
ConnorDoyle: In this case there's no difference; either way of representing "not supplied" seems equally OK to me.  This is because proper defensive practice would dictate checks for the key as well as a check for null (it's a valid JSON value after all).  Either way, I don't have a strong opinion.  As mentioned above, this doesn't appear to be used for serialization.  Please advise one way to proceed.

tsenart:
@ConnorDoyle: I'm briefly working on suggested change set that would set in place the constructs required to address your TODOs.

tsenart:
PTAL: https://github.com/mesosphere/mesos-dns/pull/267

PR #267, opened 2015-09-17
state: Refactor Task IP sourcing

tsenart:
This commit refactors the Task IP sourcing code to gracefully account
for the new `NetworkInfo` source with the same structure and be easily
changed to distinguish between IPv4 and IPv6.

### DIFF
--- a/records/config.go
+++ b/records/config.go
@@ -103,7 +103,7 @@ func NewConfig() Config {
 		HTTPOn:             true,
 		ExternalOn:         true,
 		RecurseOn:          true,
-		IPSources:          []string{"mesos", "host"},
+		IPSources:          []string{"netinfo", "mesos", "host"},
 	}
 }
 
--- a/records/state/state.go
+++ b/records/state/state.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"bytes"
+	"net"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


records/state/state_test.go
```diff
@@ -1,11 +1,13 @@
-package state
+package state_test
```
jdef: what is the advantage of doing this?

tsenart: Privacy enforcement.

records/state/state_test.go
```diff
@@ -39,62 +41,146 @@ func TestPID_UnmarshalJSON(t *testing.T) {
...
+					status(state("TASK_STOPPED"), netinfo("2.3.4.5")),
+				),
+			),
+			srcs: []string{"netinfo"},
+			want: ips("1.2.3.4"),
+		},
+		{ // statuses ordering
+			Task: task(
+				statuses(
+					status(state("TASK_RUNNING"), netinfo("1.2.3.4")),
```
jdef: this is interesting; if an IP-address for a task changes, how to we know the difference between a task that has multiple IP addresses (reported in different task status entries) vs. an IP-address change? The old code used to take the latest IP-address according to the timestamp on the status. This PR looks like it aggregates -- this is a significant change.

records/state/state_test.go
tsenart: Package `state_test` has no access to private identifiers of package `state`.

records/state/state_test.go
jdef: @nqn care to comment on this?

records/config.go
```diff
@@ -103,7 +103,7 @@ func NewConfig() Config {
 		HTTPOn:             true,
 		ExternalOn:         true,
 		RecurseOn:          true,
-		IPSources:          []string{"mesos", "host"},
+		IPSources:          []string{"netinfo", "mesos", "host"},
```
jdef: :+1:

records/state/state.go
```diff
@@ -78,66 +79,103 @@ type Task struct {
...
 	}
-	return ""
+	return ips
 }
 
-// HasDiscoveryInfo return whether the DiscoveryInfo was provided in the state.json
-func (t *Task) HasDiscoveryInfo() bool {
-	return t.DiscoveryInfo.Name != ""
+// sources maps the string representation of IP sources to their functions.
+var sources = map[string]func(*Task) []string{
```
jdef: :+1: :+1:

records/state/state.go
```diff
@@ -78,66 +79,103 @@ type Task struct {
...
+// mesosIPs returns IP addresses from the values of all
+// Task.[]Status.[]Labels whose keys are equal to
+// "MesosContainerizer.NetworkSettings.IPAddress".
+func mesosIPs(t *Task) []string {
+	return statusIPs(t, labels(MesosIPLabel))
+}
+
+// statusIPs returns IPs with the given status src from a Task's Statuses.
+func statusIPs(t *Task, src func(*Status) []string) (ips []string) {
+	for i := len(t.Statuses) - 1; i >= 0; i-- {
```
jdef: old code used to consider timestamps. see the comment i left in the test case

records/state/state_test.go
jdef: that's pretty interesting. i'm wondering if I can use this in the k8s-mesos "km" main package since it looks like the _test.go stuff in package main is bleeding into the generated non-test executable.

records/state/state_test.go
tsenart: @jdef: This PR aggregates indeed, but the ordering was preserved. Later statuses come first in the result set. This exactly what this case is testing.

jdef:
very elegant. some concerns about aggregation of IPs.

records/state/state.go
tsenart: Fortunately, statuses are sorted by timestamps by default.

records/state/state_test.go
jdef: aggregation is a new behavior, I'm concerned about making it part of this PR -- can we extract that into a follow-up PR for further discussion next week when @ConnorDoyle is back?

tsenart: @jdef: The APIs [are left intact](https://github.com/mesosphere/mesos-dns/pull/267/files#diff-d051a24769d8e34d3e0269e6398731ccL87). The aggregation is the whole point of this PR which addresses the TODOs that @ConnorDoyle left in the code.

tsenart: Sorry, it doesn't address the TODOs directly, but it sets everything in place for them to be easily done.

jdef: I think I'm OK with aggregating across source types. Across statuses though (from newest to oldest) seems super-fragile, mostly because we have no way of knowing which newer IPs are replacements for others, and is new behavior that does not belong in this PR.

tsenart: @jdef: Perhaps I let a bug through, but as far as I understand the code, semantically, nothing really changed. We were iterating though all statuses and getting the newest one before which we're also doing now.

tsenart: Communication channel upgrade: https://talky.io/mesos-dns

tsenart: @jdef: Done.

PR #277, opened 2015-09-23
records: Find user's home directory through env

### LINKED ISSUE #275: Error (Current not implemented on linux/amd64)
Hello,

I downloaded the correct version [1] but when I run I got strange errors.

[1] - https://github.com/mesosphere/mesos-dns/releases/download/v0.2.0/mesos-dns-v0.2.0-linux-amd64.tgz

For config.json:
```
{
  "zk": "zk://173.255.192.122:2181/mesos",
  "masters": ["173.255.192.122:5050"],
  "refreshSeconds": 30,
  "ttl": 60,
  "domain": "mesos",
  "port": 53,
  "resolvers": ["8.8.8.8"],
  "timeout": 10
}
```
I got (with mesos-dns moved to /usr/local/bin and running with user root):
```
sudo mesos-dns --config=/root/mesos-dns/config.json
ERROR: 2015/09/23 01:16:55 config.go:185: Failed to determine current user, translating ~/ to ./, error was user: Current not implemented on linux/amd64
```
Linux li202-122 4.1.5-x86_64-linode61 #7 SMP Mon Aug 24 13:46:31 EDT 2015 x86_64 x86_64 x86_64 GNU/Linux
Ubuntu 14.04.1 LTS

Thank you

tsenart:
As reported in https://github.com/golang/go/issues/6376,
`os/user.Current` doesn't work with cross-compilation.
This commit replaces its use with `os.Getenv` of the `$HOME` and
`$USERPROFILE` environment variables.

Fixes #275

### DIFF
--- a/records/config.go
+++ b/records/config.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
-	"os/user"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -177,14 +177,13 @@ func readConfig(file string) (*Config, error) {
 	c := NewConfig()
 
 	workingDir := "."
-	usr, err := user.Current()
-	if err != nil {
-		// this can happen (on Linux) if you're running mesos-dns as a non-root user.
-		logging.Error.Println("Failed to determine current user, translating ~/ to ./, error was", err)
-	} else {
-		workingDir = usr.HomeDir
+	for _, name := range []string{"HOME", "USERPROFILE"} { // *nix, windows
+		if dir := os.Getenv(name); dir != "" {
+			workingDir = dir
+		}
 	}
 
+	var err error
 	c.File, err = filepath.Abs(strings.Replace(file, "~/", workingDir+"/", 1))
 	if err != nil {
 		return nil, fmt.Errorf("cannot find configuration file")


records/config.go
```diff
@@ -177,14 +177,13 @@ func readConfig(file string) (*Config, error) {
 	c := NewConfig()
 
 	workingDir := "."
-	usr, err := user.Current()
-	if err != nil {
-		// this can happen (on Linux) if you're running mesos-dns as a non-root user.
-		logging.Error.Println("Failed to determine current user, translating ~/ to ./, error was", err)
```
s-urbaniak: W don't log this anymore when `HOME` and `USERPROFILE` is unset.

records/config.go
```diff
@@ -177,14 +177,13 @@ func readConfig(file string) (*Config, error) {
...
-	} else {
-		workingDir = usr.HomeDir
+	for _, name := range []string{"HOME", "USERPROFILE"} { // *nix, windows
```
s-urbaniak: I believe in windows there is also `HOMEPATH`, I wonder whether we should check this one too?!

records/config.go
tsenart: Hm, OK, this gives users some insight in those conditions. I'll add it back.

records/config.go
tsenart: http://stackoverflow.com/questions/606483/what-is-the-meaning-of-these-windows-enviroment-variables-homedrive-homepath

s-urbaniak: That's a lot of env vars to choose from, I suggest to be inspired by the go std pkg which uses the user profile directory (https://github.com/golang/go/blob/release-branch.go1.5/src/os/user/lookup_windows.go#L115).

So I suggest to leave the `USERPROFILE` env var as is.

tsenart:
Oops. Forgot to add that log line back. I'll open a small PR for that.

PR #268, opened 2015-09-18
add netinfo to validateIPSources config validation

jdef:
without this, mesos-dns does not start with default config options

/cc @tsenart @nqn @ConnorDoyle

PR #279, opened 2015-09-23
Release v0.3.0

tsenart:
#### CHANGELOG
- Support Mesos 0.25 NetworkInfo as an IPSource (#266, #267, #268)
- Find home directory with environment variables (#275, #277, #278)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+2015-09-23: v0.3.0
+  Add support for Mesos 0.25 "NetworkInfo" as an IPSource (#266, #267, #268)
+  Find home directory with environment variables (#275, #277, #278)
+
 2015-09-15: v0.2.0
   Mesos 0.24.0 compatibility (#263, #228)
   Configurable ZooKeeper detection timeout (#260)


CHANGELOG
```diff
@@ -1,3 +1,7 @@
+2015-09-23: v0.3.0
+  Support Mesos 0.25 NetworkInfo as an IPSource (#266, #267, #268)
```
sttts: It's not really clear to me what this means. "Add Mesos 0.25 support including NetworkInfo" ?

tsenart: There might be other Mesos 0.25 things that we still need to do so we can't say that.

sttts: Then make it clean in the changelog.

tsenart: ?

tsenart: Isn't it already saying exactly what feature from Mesos 0.25 we added support for?

sttts: People will read "0.25" is supported. Write something like "Experimental support ....".

tsenart: I'll rephrase it.

Issue #276, opened 2015-09-23
length of masters array always bigger than 0?


crucidal:
When using a zk string instead of a masters string the console gets spammed with the following message:
ERROR: 2015/09/23 10:55:03 generator.go:342: warning: leader "master@xxx.xxx.xxx.xxx:yyyy" is not in master list

omitting the masters variable or creating an empty array ("masters": []) generates this error/warning.

crucidal:
after some testing I noticed something obvious which I should have seen before...
I've defined DNS names in my zk string and master string.
However, the error message complains that some ip adress is not in the masters string.
I'd expect there to be no error message when I omit the masters string but at least it makes sense when I do define one with dns names.

Can this be fixed? :-)

s-urbaniak:
I was able to reproduce the issue with a local 3-master setup. 

When parsing the current leader from `state.json` during the periodic reload we get:
```
$ curl -s 127.0.0.1:5050/master/state.json | jq '.leader'
"master@127.0.0.1:5050"
```

This doesn't quite match what zookeeper gives us:
```
I0924 16:07:03.726162    2763 detect.go:219] notifying of master membership change: [
  &MasterInfo{
    Id:*20150924-150935-50331775-5050-1411,
    Ip:*50331775,Port:*5050,
    Pid:*master@127.0.0.3:5050,
    Hostname:*127.0.0.3,
    Version:*0.23.0,XXX_unrecognized:[],}
  &MasterInfo{
    Id:*20150924-150932-33554559-5050-1410,
    Ip:*33554559,Port:*5050,
    Pid:*master@127.0.0.2:5050,
    Hostname:*127.0.0.2,
    Version:*0.23.0,XXX_unrecognized:[],}
  &MasterInfo{
    Id:*20150924-150927-16777343-5050-1402,
    Ip:*16777343,
    Port:*5050,
    Pid:*master@127.0.0.1:5050,
    Hostname:*localhost,
    Version:*0.23.0,XXX_unrecognized:[],}
]
```

There is an apparent string mismatch between `"localhost"` (what zookeper gives us) and `"127.0.0.1"` (what `state.json` gives us) causing the observed warning because "127.0.0.1" is not in the master list ["localhost", "127.0.0.2", "127.0.0.3"].

jdef:
why do you have one master registering with --hostname=localhost and the
others with an IP address? or is that not the case?

s-urbaniak:
[...]

PR #290, opened 2015-09-28
detect: prefer ip address over hostname

### LINKED ISSUE #289: Prefer IP addresses over hostnames
What version of the project are you using?
https://github.com/mesosphere/mesos-dns/commit/b8922a49e14bbf346f0048ab571f688e1ee47813

What operating system and processor architecture are you using?
OSX/amd64

What did you do?
I started a 1-zookeeper/3-mesos-master/1-mesos-dns testbed as described in https://github.com/mesosphere/mesos-dns/issues/276#issuecomment-143468209.

I issued an A-record query:
`$ dig @127.0.0.1 -p 5253 master.mesos +short A`

What did you expect to see?
I expect to get the complete list of master IPs:
```
127.0.0.2
127.0.0.3
127.0.0.4
```

What did you see instead?
I get only one IP address:
```
127.0.0.2
```

s-urbaniak:
Fixes #289, #276


s-urbaniak:
@tsenart PTAL

tsenart:
LGTM

jdef:
Probably worth adding a TODO in here to upgrade to MasterInfo.Address as soon as mesos-go makes that available via updated protos. That way you don't need to be in the business of big-vs-little endian decoding (which there were some plans to change at some point).

jdef:
It sounds like little vs. big endian encoding depends on the arch that the remote UPID process is running on: https://issues.apache.org/jira/browse/MESOS-1201

jdef:
The "trivial" fix uses `ntohl` for the ordering problem (copied from the xref'd MESOS issue above):
```
 // you get the minfo from … magic!
  MasterInfo minfo = getMyMasterInfo(...);

  // then convert the bytes into host-order:
  net::IP ip{ ntohl(minfo.ip()) };
  // you can then generate back the original string ("192.168.1.10")
  std::ostringstream out;
  out << ip;
  return out.str();
```

s-urbaniak:
@jdef PTAL (just added a TODO)

tsenart:
@s-urbaniak: TODOs in code get easily lost. Please create an issue to track this instead.

s-urbaniak:
@tsenart you are very right, I just created #291

jdef:
actually wondering if it makes sense to extract this packing logic into a
set of separate go files controlled by build flags to target different
arch's since each may have different endian-ness:
https://golang.org/doc/install/source

tsenart:
@jdef: No need to do that after implementing https://github.com/mesosphere/mesos-dns/issues/291 since it is a string, not an int.

jdef:
lgtm. we're probably not going to make things any worse for people since
it's already broken. there's a (slight) chance that by fixing this for 99%
of people maybe it breaks for 1% of others. this issue, in general, isn't
fully resolvable without upgrading to the v0.24 bindings. which I don't
think mesos-go has generated protos for yet.

tsenart:
Even if we update the bindings, this new address field would only be set in more recent versions of Mesos, so older clusters would still be affected.

jdef:
true, but given the current deprecation cycle systems need to upgrade
quickly to stay compat. @nqn sent something to the mesos-dev list recently
about the fast-pace of the current deprecation process and how it might
make sense to slow it down.

jdef:
like the byte order detection

s-urbaniak:
small nit, otherwise I really like the refactorings, LGTM

PR #293, opened 2015-09-30
typo fixed

joel-hamill:


tsenart:
@joel-hamill: Thanks for the PR! I'll regenerate the docs according to https://github.com/mesosphere/mesos-dns/blob/master/docs/README.md and deploy the site.

PR #296, opened 2015-10-02
Generate correct SOA and NS records

### LINKED ISSUE #295: mesos-dns server is not responding properly to zone NS request (ERROR: mesos-dns.... resolver.go:393: NS request
Good morning everyone

We've installed and configured mesos-dns on server named compute1.mad.roche.com, to act as DNS server for some services running on top of marathon/mesos ecosystem

**cat config.json**
{
  "zk": "zk://centostest61.mad.roche.com:2181,centostest71.mad.roche.com:2181,centostest85.mad.roche.com:2181/mesos",
  "refreshSeconds": 60,
  "ttl": 60,
  "domain": "prhpc.roche.com",
  "port": 53,
  "timeout": 5,
  "email": "root.prhpc.roche.com",
  "httpon": true,
  "dnson": true,
  "httpport": 8123,
  "externalon": false,
  "listener": "10.120.39.154",
  "SOAMname": "root.ns1.prhpc.roche.com",
  "SOARname": "ns1.prhpc.roche.com",
  "SOARefresh": 60,
  "SOARetry":   600,
  "SOAExpire":  86400,
  "SOAMinttl": 60
}

mesos-dns is answering properly to requests related with services running on mesos

But when we run 
* dig @compute1.mad.roche.com prhpc.roche.com NS

We got 
**# dig @compute1.mad.roche.com prhpc.roche.com NS**

; <<>> DiG 9.9.4-RedHat-9.9.4-18.el7 <<>> @compute1.mad.roche.com prhpc.roche.com NS
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 41391
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0

;; QUESTION SECTION:
;prhpc.roche.com.               IN      NS

;; AUTHORITY SECTION:
prhpc.roche.com.        60      IN      NS      root.ns1.prhpc.roche.com.

;; Query time: 0 msec
;; SERVER: 10.120.39.154#53(10.120.39.154)
;; WHEN: Thu Oct 01 23:15:30 PDT 2015
;; MSG SIZE  rcvd: 56

**And in mesos-dns log we got**
*ERROR: 2015/10/02 08:15:20 resolver.go:393: NS request

I've tried setting the parameter SOARName to compute1.mad.roche.com, but behavior remains identical

I've read the thread https://github.com/mesosphere/mesos-dns/issues/73, and it looks like this behavior shall be addressed already.

Could any of you guys please help me set the NS resolution properly or confirm me if this behavior is expected one?

Kind Regards

tsenart:
The SOAMname and SOARname were switched around in a few places which
caused SOA and NS records to be wrong.

Fixes #295


tsenart:
Note to self: Update the `gh-pages` branch with the doc changes after this is merged.

sttts:
As far as I understand the matter, it lgtm

config.json.sample
```diff
@@ -9,8 +9,8 @@
   "resolvers": ["8.8.8.8"],
   "timeout": 5,
   "listener": "0.0.0.0",
-  "SOAMname": "root.ns1.mesos",
-  "SOARname": "ns1.mesos",
+  "SOAMname": "ns1.mesos",
```
jdef: should we specify a sample name that's fully qualified? (ends in a dot)

e.g. `ns1.mesos.`

config.json.sample
```diff
@@ -9,8 +9,8 @@
...
+  "SOARname": "root.ns1.mesos",
```
jdef: ditto

see http://www.zytrax.com/books/dns/ch8/soa.html

docs/docs/configuration-parameters.md
```diff
@@ -61,7 +61,9 @@ It is sufficient to specify just one of the `zk` or `masters` field. If both are
 
 `externalon` is a boolean field that controls whether Mesos-DNS serves requests outside of the Mesos domain. The default value is `true`. 
 
-`SOAMname` is the MNAME field in the SOA record for the Mesos domain. The format is `mailbox.domain`, using a `.` instead of `@`. For example, if the email address is `root@ns1.mesos`, the `email` field should be `root.mesos-dns.mesos`. For details, see the [RFC-1035](http://tools.ietf.org/html/rfc1035#page-18). The default value is `root.ns1.mesos`. 
+`SOAMname` specifies the domain name of the name server that was the original or primary source of data for the configured domain. The default value is `ns1.mesos`.
```
jdef: as per my prior comment, should the default value be fully qualified?

karlkfi:
@jdef I don't know enough about SOAs or DNS zones to answer that. But I have documented the symptoms here for the time being: https://github.com/kubernetes/kubernetes/issues/14394#issuecomment-145640550

docs/docs/configuration-parameters.md
tsenart: Internally all names are fully qualified, but we haven't been requiring users to do it explicitly.

jdef: Our code seems wonky.

Our default values (and docs) specify relative names:
https://github.com/mesosphere/mesos-dns/blob/master/records/config.go#L76

Then we later **assume** that the config should use fully qualified names
(even if a user doesn't want it!):
https://github.com/mesosphere/mesos-dns/blob/master/records/config.go#L125

So basically there's no way to use a relative name. Is that OK?

tsenart: I think it's OK to enforce FQDNs for the SOA record.

jdef: Perhaps we should document that it's enforced this way?

tsenart: https://github.com/mesosphere/mesos-dns/pull/304

jdef: thanks!

Issue #301, opened 2015-10-06
Research and possibly remove recursion logic from external resolver (forwarder)


tsenart:
Talking to @discordianfish brought some light to the weird practice that we're undertaking here. Usually a DNS server won't be both a forwarder and a unorthodox recurser. I'll have to do some code archeology to understand the rationale behind this.

tsenart:
Ref: https://github.com/mesosphere/mesos-dns/issues/302

tsenart:
Read this: https://tools.ietf.org/html/rfc5625

tsenart:
It turns out that this logic came from the first commit in the repo, so no real explanation.

discordianfish:
To sum up what I said: mesos-dns should be an authoritative nameserver for .mesos and it should simply forward queries for all other zones to a recursive resolving nameserver (aka something like 8.8.8.8 or everything that end up in resolv.conf). It shouldn't do any other logic beside forwarding, similar to the bind option 'forwarders'. This is also what tools like consul are doing.

PR #302, opened 2015-10-06
exchanger: Fix Recurse logic

tsenart:
This commit fixes a logic bug introduced with the refactoring of the external
resolver code: https://github.com/mesosphere/mesos-dns/commit/31f5539

The bug manifested when external queries would be sent to configured resolvers
that would reply with non-authoritative answers. These answers weren't
considered valid and if the message would contain an SOA record in its
authoritative section then it would be erroneously recursed.

### DIFF
--- a/exchanger/exchanger.go
+++ b/exchanger/exchanger.go
@@ -85,10 +85,10 @@ func Instrumentation(c logging.Counter) Decorator {
 type Recurser func(*dns.Msg) string
 
 // Recurse is the default Mesos-DNS Recurser which returns an addr (host:port)
-// only when the given dns.Msg doesn't contain authoritative answers and has at
-// least one SOA record in its NS section.
+// only when the given dns.Msg doesn't doesn't contain answers or isn't
+// authoritative and has at least one SOA record in its NS section.
 func Recurse(r *dns.Msg) string {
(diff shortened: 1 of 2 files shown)


tsenart:
Ref: https://github.com/mesosphere/mesos-dns/issues/301

exchanger/exchanger.go
```diff
@@ -85,10 +85,10 @@ func Instrumentation(c logging.Counter) Decorator {
 type Recurser func(*dns.Msg) string
 
 // Recurse is the default Mesos-DNS Recurser which returns an addr (host:port)
-// only when the given dns.Msg doesn't contain authoritative answers and has at
-// least one SOA record in its NS section.
+// only when the given dns.Msg doesn't doesn't contain answers or isn't
+// authoritative and has at least one SOA record in its NS section.
 func Recurse(r *dns.Msg) string {
-	if r.Authoritative && len(r.Answer) > 0 {
+	if r.Authoritative || len(r.Answer) > 0 {
```
sttts: The comment now **does not** match the logic:

not-authorative or no answer = !r.Authoritative || len(r.Answer)==0 = !(r.Authoritative &&  len(r.Answer)>0)

tsenart:
Closed in favour of #307.

Issue #298, opened 2015-10-04
Server misbehaving


rafaelcapucho:
Hello,

I setup Mesos-DNS in a linode and I prepend my 2 Mesos-DNS apps in resolv.conf:

```
domain members.linode.com
search members.linode.com
nameserver 72.14.179.25 # Atlas004
nameserver 198.58.99.72 # Atlas005
nameserver 23.239.24.5
nameserver 173.255.199.5
nameserver 66.228.53.5
nameserver 96.126.122.5
nameserver 96.126.124.5
nameserver 96.126.127.5
nameserver 198.58.107.5
nameserver 198.58.111.5
nameserver 72.14.179.5
nameserver 72.14.188.5
#options rotate
```

Its work good, dig working, resolv master.mesos properly but there're a lot of errors in Mesos Sandbox Log, a lot (per second) like that: 

```error exchanging with "23.239.24.5:53": dial udp: lookup dns.jxnc.cncnet.net. on 198.58.111.5:53: server misbehaving```

![4](https://cloud.githubusercontent.com/assets/163367/10266533/d9186db8-6a41-11e5-98d6-f7cc1f95bab8.png)

The config used:

```
{
  "zk": "zk://atlas001:2181,atlas002:2181,atlas003:2181/mesos",
  "masters": ["atlas001:5050", "atlas002:5050", "atlas003:5050"],
  "refreshSeconds": 60,
  "ttl": 120,
  "domain": "mesos",
  "port": 53,
  "resolvers": ["173.255.199.5","66.228.53.5","96.126.122.5","96.126.124.5","96.126.127.5","198.58.107.5","198.58.111.5","23.239.24.5","72.14.179.5","72.14.188.5","8.8.8.8"],
  "timeout": 15,
  "email": "root.mesos-dns.mesos"
}
```
All IPs except the last are from linode DNS Server... I'm launching using marathon, with this code:

```
{
  "cmd": "/bin/mesos-dns -config=/mnt/mesos/sandbox/config.json",
  "cpus": 0.2,
  "mem": 50,
  "ports": [53],
  "id": "mesos-dns",
  "instances": 1,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "nexusedge/docker-mesos-dns",
      "network": "HOST",
      "privileged": true
    }
  }, "uris":[
    "http://blog.rafaelcapucho.com/download/config.json"
  ],
[...]

Issue #297, opened 2015-10-02
Upstream i/o timeouts with mesos slave authentication (and docker registry)


mcoffin:
# Overview

1. I'm currently using `v0.3.0`, but this issue appeared somewhere between `v0.1.2` and `v0.2.0`
2. I am running this on ubuntu 15.04 (amd64) on EC2. I have tried running mesos-dns both from the released binaries from the github releases as well as through a [docker container](https://hub.docker.com/r/mcoffin/mesos-dns). Docker version is 1.8.1.
3. I'm running mesos-dns on 3 mesos-masters. When a new slave is spun up, it's resolv.conf nameservers are replaced with the ips of the 3 mesos masters. When a slave tries to authenticate with these masters, timeouts will be logged by mesos-dns and the slave will fail to authenticate (as the timeouts in mesos-dns also cause the authentication to time out.). As soon as I kill mesos-dns, the slave & master will fall back to using the each other's ip rather than hostname resolution and connect successfully. This issue was not observed with mesos 0.23.0 and mesos-dns 0.1.2 under an identical configuration.
4. I expected to see the hostnames *and* other DNS queries successfully execute without timeouts.
5. Instead, we see that *some* DNS queries, but not others, cause upstream i/o timeouts. This appears to only happen with recursive queries, and only a consistent subset of them.

# Original Post

I'm not sure what's unique about the resolving that mesos slave authentication and docker registry so, but somewhere between 0.1.2 and 0.2.0 (and still present in 0.3.0), something broke that makes creates upstream timeouts from these requests that produce logs like the following:

```
Oct 02 20:43:58 ip-10-0-0-254 docker[5812]: ERROR: 2015/10/02 20:43:58 resolver.go:281: failed external DNS lookup of "registry-1.docker.io.us-west-2.compute.internal.": read udp 10.0.0.254:34856->172.16.0.23:5
Oct 02 20:43:58 ip-10-0-0-254 docker[5812]: ERROR: 2015/10/02 20:43:58 resolver.go:280: registry-1.docker.io.us-west-2.compute.internal.
Oct 02 20:43:58 ip-10-0-0-254 docker[5812]: ERROR: 2015/10/02 20:43:58 exchanger.go:64: error exchanging with "10.0.0.2:53": read udp 10.0.0.254:34856->172.16.0.23:53: i/o timeout
[...]

PR #303, opened 2015-10-06
exchanger: Improve ErrorLogging

tsenart:
While debugging #297 and #298, it became apparent that this extra
information would be very useful in aiding understanding of the issue.

### DIFF
--- a/exchanger/exchanger.go
+++ b/exchanger/exchanger.go
@@ -61,7 +61,7 @@ func ErrorLogging(l *log.Logger) Decorator {
 		return Func(func(m *dns.Msg, a string) (r *dns.Msg, rtt time.Duration, err error) {
 			defer func() {
 				if err != nil {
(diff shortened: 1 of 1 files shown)


exchanger/exchanger.go
```diff
@@ -61,7 +61,7 @@ func ErrorLogging(l *log.Logger) Decorator {
 		return Func(func(m *dns.Msg, a string) (r *dns.Msg, rtt time.Duration, err error) {
 			defer func() {
 				if err != nil {
-					l.Printf("error exchanging with %q: %v", a, err)
+					l.Printf("error exchanging %v with %q: %v", m.Question[0], a, err)
```
sttts: m.Question is always non-nil and len(m)>1?

tsenart: Good catch, we may have no question, but it will never be nil.

PR #305, opened 2015-10-06
docs: s/SOARname/SOAMname/g

tsenart:


sttts:
lgtm, matching https://github.com/mesosphere/mesos-dns/blob/master/resolver/resolver.go#L213

Issue #284, opened 2015-09-24
Proposal: Mesos-DNS should stay running even when disconnected from Master/ZK


sepiroth887:
Currently 2 issues exist: 

1.) If the initial ZK connection doesn't success mesos-dns will not start
2.) If the master detection doesn't find a master in 30sec it will crash

Issue 2 is probably the most critical as it has a chance to be fatal during a network partition and will not recover automatically after the network is restored. 

Suggestion: 

If ZK connection doesn't succeed keep retrying (possibly with a capped backoff to reduce chatter) indefinitely. Continue service requests (e.g. to allow Resolvers to work)

If Master detection fails, don't panic :) just log it and wait for the next reload cycle to try again.

tsenart:
SGTM

sepiroth887:
Ugh. this actually bit me hard today when trying to migrate a ZK cluster on a running mesos. I will try and see if I can fix this quickly ^^ any suggestions? 

What i'm thinking right now is to do a reload regardless of master/zk connection and ensure that the data structures holding the records are kept around. 

Also making the first reload async of the connection to masters may be an option i will look into.

sepiroth887:
Ok quick update. 

It seems to be relatively simple as most of the logic is async already (yay go!) 

I think i can get a clean PR for review and feedback ready pretty quickly. 

Basically whats can be done is: 

Dont log fatal, instead just very verbose. 
Dont return on parseState when no master is found. Instead continue on but with an error sent back to ensure its logged and further logic can be placed in the future. 

E.g. For static records after the error is checked those could still be added to the rrs maps but outside the InsertState mehod to avoid recreating the maps.

tsenart:
Please keep the static records proposal apart from this one.

sepiroth887:
I am no worries. Just making sure that the logic is sound and compatible for any non-master driven records if there are/will be any.

PR #306, opened 2015-10-06
Prevent exit on ZK detection timeout

sepiroth887:
Addresses: #284 

This one line fix will do it. There are other instances where mesos-dns could fail fatally due to dependent libraries but thats outside the scope of this fix.

### DIFF
--- a/main.go
+++ b/main.go
@@ -66,11 +66,6 @@ func main() {
 
 	reload := time.NewTicker(time.Second * time.Duration(config.RefreshSeconds))
 
-	zkTimeout := time.Second * time.Duration(config.ZkDetectionTimeout)
-	timeout := time.AfterFunc(zkTimeout, func() {
-		errch <- fmt.Errorf("master detection timed out after %s", zkTimeout)
-	})
-
 	defer reload.Stop()
 	defer util.HandleCrash()
 	for {
@@ -78,7 +73,6 @@ func main() {
 		case <-reload.C:
 			res.Reload()
 		case masters := <-changed:
-			timeout.Stop()
 			logging.VeryVerbose.Printf("new masters detected: %v", masters)
 			res.SetMasters(masters)
 			res.Reload()
--- a/records/config.go
+++ b/records/config.go
@@ -23,8 +23,6 @@ type Config struct {
 	// Timeout is the default connect/read/write timeout for outbound
 	// queries
 	Timeout int
-	// Zookeeper Detection Timeout: how long in seconds to wait for Zookeeper to be initially responsive (default 30)
-	ZkDetectionTimeout int
 	// NOTE(tsenart): HTTPPort, DNSOn and HTTPOn have defined JSON keys for
 	// backwards compatibility with external API clients.
 	HTTPPort int `json:"HttpPort"`
@@ -66,26 +64,25 @@ type Config struct {
 // NewConfig return the default config of the resolver
 func NewConfig() Config {
 	return Config{
-		ZkDetectionTimeout: 30,
-		RefreshSeconds:     60,
-		TTL:                60,
-		Domain:             "mesos",
-		Port:               53,
-		Timeout:            5,
-		SOARname:           "root.ns1.mesos",
-		SOAMname:           "ns1.mesos",
-		SOARefresh:         60,
-		SOARetry:           600,
-		SOAExpire:          86400,
-		SOAMinttl:          60,
-		Resolvers:          []string{"8.8.8.8"},
-		Listener:           "0.0.0.0",
-		HTTPPort:           8123,
-		DNSOn:              true,
-		HTTPOn:             true,
-		ExternalOn:         true,
-		RecurseOn:          true,
-		IPSources:          []string{"netinfo", "mesos", "host"},
+		RefreshSeconds: 60,
+		TTL:            60,
+		Domain:         "mesos",
+		Port:           53,
+		Timeout:        5,
+		SOARname:       "root.ns1.mesos",
+		SOAMname:       "ns1.mesos",
+		SOARefresh:     60,
+		SOARetry:       600,
+		SOAExpire:      86400,
+		SOAMinttl:      60,
+		Resolvers:      []string{"8.8.8.8"},
+		Listener:       "0.0.0.0",
+		HTTPPort:       8123,
+		DNSOn:          true,
+		HTTPOn:         true,
+		ExternalOn:     true,
+		RecurseOn:      true,
+		IPSources:      []string{"netinfo", "mesos", "host"},
 	}
 }
 
@@ -129,7 +126,6 @@ func SetConfig(cjson string) Config {
 	logging.Verbose.Println("Mesos-DNS configuration:")
 	logging.Verbose.Println("   - Masters: " + strings.Join(c.Masters, ", "))
 	logging.Verbose.Println("   - Zookeeper: ", c.Zk)
-	logging.Verbose.Println("   - ZookeeperDetectionTimeout: ", c.ZkDetectionTimeout)
 	logging.Verbose.Println("   - RefreshSeconds: ", c.RefreshSeconds)
 	logging.Verbose.Println("   - Domain: " + c.Domain)
 	logging.Verbose.Println("   - Listener: " + c.Listener)


jdef:
errch reports errors for components other than zk as well and if those fail (dns/http) then the server should probably die. if you want to treat zk separately then use a separate error channel.

sepiroth887:
:+1: updated

main.go
```diff
@@ -40,16 +40,17 @@ func main() {
 	// initialize resolver
 	config := records.SetConfig(*cjson)
 	res := resolver.New(Version, config)
-	errch := make(chan error)
+	genErrch := make(chan error)
```
tsenart: `errch`

main.go
```diff
@@ -40,16 +40,17 @@ func main() {
...
+	zkErrch := make(chan error)
```
tsenart: `zkerrs`

main.go
```diff
@@ -82,8 +83,10 @@ func main() {
 			logging.VeryVerbose.Printf("new masters detected: %v", masters)
 			res.SetMasters(masters)
 			res.Reload()
-		case err := <-errch:
-			logging.Error.Fatal(err)
+		case err := <-genErrch:
+			log.Fatal(err)
+		case zkErr := <-zkErrch:
+			logging.VeryVerbose.Println(zkErr)
```
tsenart: `Verbose`

main.go
sepiroth887: doesnt this confuse things a bit? what about generrs?

tsenart:
Thanks! Can you squash the commits into one?

sepiroth887:
Yup. Though would it be easier to close and reopen ? Not 100% sure but it would require rewriting the history. Does the build system cope with a force update?

jdef:
you can force-push

tsenart:
Wait a minute. Thinking more about this, I don't think this approach is what we want. We'd like ZK to not timeout at all, I think, so that it can connect at any point in the process run-time.

karlkfi:
I don't understand this PR... 
Am I correct in reading that it disables the zk timeout and just logs once and keeps on going? Why have the timeout at all? I don't understand.

tsenart:
@karlkfi: Agreed, see my comment above.

karlkfi:
Not timing out is debatable. Perhaps the timeout can be optional?

Do we have a readiness endpoint on mesos-dns that fails until zk is available? It might be nice to distinguish between alive & ready. It's a pattern kubernetes adopted. NotAlive means it needs restarting. Alive but NotReady means still bootstrapping or that dependencies are NotReady.

tsenart:
Mesos-DNS offers multiple services which can function without Zookeeper. It's even allowed to configure it without Zookeeper altogether. The timeout is strictly not needed as long as the code which connects keeps retrying.

sepiroth887:
Would be easy to remove the timeout. It's just a Ticker anyways. I don't see a reason to keep it.

karlkfi:
Technically it's a Timer, not a Ticker. So it only happens once, as is. A Ticker would invoke repeatedly.

tsenart:
@sepiroth887: Yes, but in order to satisfy #284 you should make sure that the ZK detector keep retrying to connect.

sepiroth887:
gah. yea it is a Timer. sorry. regardless though it's not really needed.  The ZK detector already keeps retrying indefinitely.

Sending yet another commit shortly.

tsenart:
Since `config.ZkDetectionTimeout` isn't used anymore, please remove all references to it, including documentation.

sepiroth887:
rabbit hole :D on it.

main.go
```diff
@@ -78,12 +73,11 @@ func main() {
 		case <-reload.C:
 			res.Reload()
 		case masters := <-changed:
-			timeout.Stop()
 			logging.VeryVerbose.Printf("new masters detected: %v", masters)
 			res.SetMasters(masters)
 			res.Reload()
 		case err := <-errch:
-			logging.Error.Fatal(err)
+			log.Fatal(err)
```
tsenart: Keep using the Error logger please.

sepiroth887: this was unchanged. I'll update it though.

tsenart: How was it unchanged if git lists it as change?

sepiroth887: ugh yes. was keeping it in line with line 59 and 61. Want this changed as well to use the Error logger?

tsenart: I simply want this line to not change at all.

tsenart:
@sepiroth887: Please squash all your commits into one and the force-push to your branch.

sepiroth887:
done.

jdef:
For posterity: we added this initial ZK timeout feature because the old mesos-go zk detector implementation was buggy and would sometimes *never* connect. The problem with this is that even if the masters are initially configured statically in the config, the mesos-dns records that point to masters would get out of date if the master VMs were recycled. **This happened in production.**

The newer mesos-go zk detector should be much more stable. However, there's no guarantee that it's complete bug free.

I'm not thrilled that this PR has merged. I'd rather have seen a much higher timeout specified, perhaps something along the lines of O(hours)? That way an admin would at least see the mesos-dns service flapping and could investigate why.

tsenart:
@jdef: Couldn't that be solved simply by logging when new masters are detected? An admin would look and see that it's getting data from ZK. When those would be absent, we'd know something is off.

jdef:
My suspicion is that admins are watching a million things and that a
flapping service that's being monitored via some alert system is more
likely to get attention than a log line.

sepiroth887:
What about a healthy/unhealthy endpoint? 

What scared me most about the previous version was that the chance of mesos dns just exiting can have a major impact on a production environment. 
Even if it should never recover it would at least not prevent running apps from blowing to pieces. 

Monitoring the health of the dns could at least provide ops with the means to alert and take action.

sepiroth887:
We experienced this outage btw due to that issue. So naturally id like to mitigate that.

karlkfi:
I like the idea of a readiness endpoint that fails if not connected to mesos. Yes, mesos-dns functions without mesos, but I don't think that's actually a legitimate use case. If it can't talk to mesos the operator is gonna want to know.

jdef:
+1 for a readiness endpoint. should also integrate well with DCOS. I'm not
very familiar with how they currently monitor readiness of other critical
components.

sepiroth887:
@karlkfi sorry for blatantly selling your idea as mine :D 

one gotcha would be that this doesn't help if someone disables http on mesos-dns right? Unless you always bind the readiness endpoint to another port by default? but this also feels weird. :( 

Is there some form of DNS entry that could be hijacked to achive the same and support both DNS and HTTP based health?

jdef:
some k8s services have a dedicated endpoint/port that respond to "/healthz"
HTTP requests. Our readiness check could (and probably should) be
completely separate from the HTTP API.

cmaloney:
We do not have any infrastructure to monitor "Readiness" / "liveliness" of a base system service. We can monitor when a process is up and act on that with systemd.

Practically it would be good if Mesos-DNS just always continued operating serving old / cached results and causing an alert, but that requires a bunch of infrastructure we don't have currently.

tsenart:
@cmaloney: So, in the context of DCOS, would you say its preferable to have Mesos-DNS hard crash if it can't connect to Zookeeper at any point in time?

cmaloney:
The two options I see:

1. Hard Crash (very easy to detect and forward to a sysadmin)
2. Refuse all connections (prevents us from thinking it is up / working), and report via the systemd notify API (http://www.freedesktop.org/software/systemd/man/systemd-notify.html, http://www.freedesktop.org/software/systemd/man/sd-daemon.html) that it is in a degraded state. (Reasonably easy to notice, shows up in systemd logs).

We always configure it to listen to a localhost zookeeper though, so the only time it can't connect to zookeeper is if zookeeper is overloaded or not currently up. In the case where it is not up we'd rather use another machine, since using that particular

If the "watch zk" stuff was a separate process / daemon from the "serve records" then we could split this much better / make it monitorable as two different services which would be easier to detect / notify system administrators about, while still keeping the "old" state available so the cluster is less likely to fall over. There would also need to be work around not serving leader.mesos / master.mesos when there is no connection to zk (Which will cause certain things to fail properly), and work in generating /etc/resolv.conf so that we detect when mesos-dns is in this "degraded" state and avoid it if there are other healthy nodes.

tsenart:
OK, until we have a decent solution to properly monitor base services in DCOS, I will revert this change set and instead make the timeout **optional** so that we cover both use cases.

karlkfi:
I'm not a big fan of either of those options, honestly.

The best multi-component systems to maintain are the ones that:
1. can come up in any order
2. make it obvious which piece is broken
3. recover automatically with the least amount of intervention

You don't want to have to trace failures down a cascade of broken services and then have to restart all of them in order. You want to be able to look at a dashboard and see that N systems are Alive but NotReady and 1 system is NotAlive. You go fix that one piece or config that's borked and all the others auto-recover.

Not only that, but every component should be watched for NotAlive and auto-restarted either by process or by container. If the kube-dns crashes every time zk is unreachable it will thrash every time there's a network outage. If that happens at the container level you won't even have a running container to shell into to debug.

tsenart:
@karlkfi: Agree.

cmaloney:
I think it's important to note here that the reason why we're serving DNS is that we're talking to dumb clients. From what I know of internals / reliance on DNS inside of DCOS (Universe, DCOS UI, Frameworks), Mesos-DNS soft failing is going to result in a lot worse error cases than hard failing.

If Mesos-DNS hard fails from a clients perspective it looks just like you lost network access to it. Clients which want to be resillient against network partitions from Mesos-DNS need to handle this failure scenario already.

If we add soft failure of Mesos-DNS, we break some "liveliness" guarantees of responses from it from the PoV of different clients. Two I'm going to focus on are leader.mesos and master.mesos. There is a lot of software in DCOS (every universe service) which assumes master.mesos resolves to all the masters for using as a zk url. There is some software in DCOS (The UI, dcos history service, dependence on there being a leading mesos master by slaves) which assumes leader.mesos points to the current leading Mesos master.

If Mesos-DNS keeps serving even though it doesn't have a current / up to date leader.mesos + master.mesos (what it gets from zookeeper), then clients which query those are going to get incorrect responsesfrom their point of view. For some clients this is acceptable as a transient error. For others, (the UI, dcos history service, dependence on there being a leading mesos master by slaves) they need to change behavior when they know they are using an old, cached value. If Mesos-DNS keeps serving the old value, they are unable to make this choice. From their perspective as consumers of DNS, they are getting a new authoritative value, and have no reason to suspect the value is possibly outdated. This will lead to that particular service not realizing it is in a degraded state. Services which depend on it, depending on that dependance, may also be in a degraded state but not know it.

If Mesos-DNS hard fails, the clients see that immediately, and can opt to use the old value if they are smart clients, if they are dumb clients they are guaranteed that they won't have their assumptions around the liveliness of the values reported invalidated.

By introducing "soft" failure as keeping up the existing service port listenging / respondign in Mesos-DNS there are more failure scenarios a client must be aware of in order to operate properly with certain guarantees around service discovery. This breaks the simple client implementation. That implementation is why we are doing DNS in the first place.

If we implement "soft" failure in Mesos-DNS as serving on a different fallback port (ex: Port 153 is "always available", port 53 serving is "only when we're all good"), the guarnatees aren't broken, as dumb clients will get the failure they need for correct operation. To use the fallback clients need to opt-in rather than opt out.

One failure line which would happen today with soft failure:

1. Mesos-DNS keeps serving even though it lost a zk connect
2. /etc/resolv.conf keeps using 127.0.0.1 as the resolver, because Mesos-DNS is still up
3. The DCOS UI nginx keeps resolving master.mesos, leader.mesos. It uses leader.mesos to grab "state-summary" from the leading master. Mesos serves this file from every single master always. If you get it from not the leading master, it's contents are frequently incorrect.
4. An administrator uses the DCOS UI to check on the state of their cluster. The cluster "looks" healthy, because the state-summary being returned says everything is healthy, even though on the current leading master this is not the case.
5. An administrator uses the DCOS UI to view the Web UI of a particular service inside their cluster. The DCOS UI looks up webui_url as reported by the framework and sent to the ui in "state-summary". The DCOS UI forwards the administators request to that particular server. That server is no longer supposed to be serving the WebUI for the framework. The server is still there, but the framework migrated it to a different host, and updated the webui_url in mesos. The administrator therefore gets old / incorrect information about the service which they believe to be correct / authoritative information.

tsenart:
@cmaloney: Thanks for the detailed insight into these DCOS failure modes. I hope #320 makes everyone happy.

PR #307, opened 2015-10-06
Transparent proxying of external queries

### LINKED ISSUE #301: Research and possibly remove recursion logic from external resolver (forwarder)

tsenart:
The original external queries code mixed concepts of a proxy (forwarder) and a recurser for no good apparent reason. As per https://tools.ietf.org/html/rfc5625, this change set simplifies things.

Related to #297, #298.
Closes #301.

### DIFF
--- a/exchanger/exchanger.go
+++ b/exchanger/exchanger.go
@@ -2,7 +2,6 @@ package exchanger
 
 import (
 	"log"
-	"net"
 	"time"
 
 	"github.com/mesosphere/mesos-dns/logging"
@@ -36,24 +35,6 @@ func Decorate(ex Exchanger, ds ...Decorator) Exchanger {
 	return decorated
 }
 
-// Pred is a predicate function type for dns.Msgs.
-type Pred func(*dns.Msg) bool
-
-// While returns an Exchanger which attempts the given Exchangers while the given
-// predicate function returns true for the returned dns.Msg, an error is returned,
-// or all Exchangers are attempted, in which case the return values of the last
-// one are returned.
-func While(p Pred, exs ...Exchanger) Exchanger {
-	return Func(func(m *dns.Msg, a string) (r *dns.Msg, rtt time.Duration, err error) {
-		for _, ex := range exs {
-			if r, rtt, err = ex.Exchange(m, a); err != nil || !p(r) {
-				break
-			}
-		}
-		return
-	})
-}
-
 // ErrorLogging returns a Decorator which logs an Exchanger's errors to the given
 // logger.
 func ErrorLogging(l *log.Logger) Decorator {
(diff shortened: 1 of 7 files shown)


exchanger/exchanger.go
```diff
@@ -70,50 +51,18 @@ func ErrorLogging(l *log.Logger) Decorator {
...
+func Instrumentation(total, success, failure logging.Counter) Decorator {
 	return func(ex Exchanger) Exchanger {
 		return Func(func(m *dns.Msg, a string) (r *dns.Msg, rtt time.Duration, err error) {
-			for i := 0; i <= max; i++ {
-				if r, rtt, err = ex.Exchange(m, a); err != nil {
-					break
-				} else if a = rec(r); a == "" {
-					break
+			defer func() {
+				if total.Inc(); err != nil {
```
jdef: this looks strange because the increment of success depends on a negative outcome of total.Inc(). unless I'm missing something, total.Inc() should really be on its own line.

exchanger/forwarder.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/miekg/dns"
+)
+
+// A Forwarder is a DNS message forwarder that transparently proxies messages
+// to DNS servers.
+type Forwarder func(*dns.Msg, string) (*dns.Msg, error)
+
+// Forward is an utility method that calls f itself.
+func (f Forwarder) Forward(m *dns.Msg, proto string) (*dns.Msg, error) {
```
jdef: don't understand the need for this utility func at all since it adds no value.

tsenart: There is no need, that's why it's an utility function. Code reads much better with it. If we define an interface, it would also implement it automatically.

exchanger/exchanger.go
tsenart: `total.Inc()` has no impact on the conditional. It runs unconditionally.

resolver/resolver.go
```diff
@@ -246,52 +245,28 @@ func shuffleAnswers(rng *rand.Rand, answers []dns.RR) []dns.RR {
...
-		logging.Error.Println(err)
-		logging.CurLog.NonMesosFailed.Inc()
-	} else {
-		// nxdomain
-		if len(m.Answer) == 0 {
-			logging.CurLog.NonMesosNXDomain.Inc()
-		} else {
-			logging.CurLog.NonMesosSuccess.Inc()
-		}
+		m = new(dns.Msg).SetRcode(r, rcode(err))
```
jdef: we use to log errors. why did we stop?

resolver/resolver.go
```diff
@@ -29,9 +29,7 @@ type Resolver struct {
 	rs      *records.RecordGenerator
 	rsLock  sync.RWMutex
 	rng     *rand.Rand
-
-	// pluggable external DNS resolution, mainly for unit testing
-	extResolver exchanger.Exchanger
+	exchanger.Forwarder
```
jdef: instead of exposing a public field, just declare this as `fwd` and get rid of the confusing utility func (as previously noted)

resolver/resolver.go
tsenart: We didn't. They're logged with the `ErrorLogging` decorator.

jdef:
a few minor comments. like the cleanness of forwarding vs. recursing :+1:

resolver/resolver.go
tsenart: I'm pro making this private, but you have to convince me of the **confusion** associated with the `Forward` method. :-)

jdef: Defining a func on a func that calls itself is usually done to implement an
interface. Without the interface it adds unnecessary complexity to the API
(unless, of course, it adds some business/app logic value -- which this
impl doesn't).

Forwarder is already a func type. So something like `resolver.fwd(...)`
reads just as easily as what you have now, makes Forwarder field private,
and keeps the exchanger API simpler. If you don't like `fwd` for a field
name then name it `forward` and then the code will read
`resolver.forward(...)` which basically maintains the readability + the
other pros I mentioned.

tsenart: Yeah, in this case that's fine since it's becoming privatised. But when your API consists of anonymous functions being passed around, it's useful to have a more descriptive utility method attached to it so that you can name the functions other things than verbs.

If you look closer, there is a language mismatch. `NewForwarder` returns a `Forwarder` which is stored in a field with a verb name `forward`. With the utility method, this goes away.

jdef: I see where you're going with this. Could also just call the func type
Forward (it totally makes sense since it's a func type, not an interface
type). Then you get readability and drop the added/unneeded complexity.

tsenart: `func NewForward(...) Forward` doesn't ring well :/

PR #308, opened 2015-10-06
Proposal: Generate records for staging tasks

gabrielhartmann:
In some cases it is desirable for a Task to depend on resolution of its
own DNS address before it reports itself as RUNNING.  It's currently
impossible because DNS addresses are only generated for RUNNING Tasks.
This leads to the undesirable scenario of having to lie to Mesos in
order to get a DNS address generated.

There may be unknown legacy reasons why this policy has been instituted, that I don't know about.  I also haven't tested this etc.  This PR is mostly to get feedback about what's going on here.  If this just wasn't a supported scenario before, and it's a reasonable change I'm happy to test and do whatever else is required.

### DIFF
--- a/records/state/state.go
+++ b/records/state/state.go
@@ -162,7 +162,8 @@ func statusIPs(st []Status, src func(*Status) []string) []string {
 	// https://github.com/apache/mesos/blob/0.24.0/src/slave/slave.cpp#L5226-L5238
 	ts, j := -1.0, -1
 	for i := range st {
-		if st[i].State == "TASK_RUNNING" && st[i].Timestamp > ts {
+		state := st[i].State
+		if (state == "TASK_RUNNING" || state == "TASK_STAGING") && st[i].Timestamp > ts {
 			ts, j = st[i].Timestamp, i
 		}
 	}


tsenart:
From: #310
> Mesos-DNS as a service discovery system should be health-aware. This doesn't mean that it can *guarantee* healthiness of the returned service instances, only that it does its best to direct clients to capable ones.

I don't see how this proposal is compatible with the above but please let me know if you see a way.

jdef:
It probably makes sense to expose an API somewhere that lists ALL tasks, regardless of whether they are healthy. Perhaps via the HTTP API when the caller specifies an extra param in the URL (`&status=all`?)

tsenart:
@jdef: Unhealthy tasks seem to be killed by Mesos: https://github.com/apache/mesos/blob/master/src/health-check/main.cpp#L108

jdef:
Yes. That's OK. Grace period exists for a reason. Some cluster-oriented
software needs to bootstrap with an initial list of nodes and can't wait
for them all to be *ready* (because in that case the cluster never comes
up). Allowing a client to scrape the list of services for *all* things,
regardless of health, supports this use case.

tsenart:
OK, thanks James, that makes sense. @gabrielhartmann: Please write tests for this change.

records/state/state.go
```diff
@@ -162,7 +162,8 @@ func statusIPs(st []Status, src func(*Status) []string) []string {
 	// https://github.com/apache/mesos/blob/0.24.0/src/slave/slave.cpp#L5226-L5238
 	ts, j := -1.0, -1
 	for i := range st {
-		if st[i].State == "TASK_RUNNING" && st[i].Timestamp > ts {
+		state := st[i].State
+		if (state == "TASK_RUNNING" || state == "TASK_STAGING") && st[i].Timestamp > ts {
```
tsenart:
```go
state := st[i].State[:4]
if (state == "RUNNING" || state == "STAGING") && st[i].Timestamp <= ts {
```

jdef: That said, there's probably also a use case that says: "show me only the
healthy services". I think we need to support both.

tsenart: Yeah, although Mesos-DNS can't guarantee the health of a record. How long do you think a task can be in staging state? What would be the implications of clients connecting to services in that state? Wouldn't they need to retry in case of failure anyway?

gabrielhartmann: "This doesn't mean that it can guarantee healthiness of the returned service instances, only that it does its best to direct clients to capable ones."

I think an appropriate interpretation of doing its best to direct clients to capable nodes would be the following:

Give the DNS address to the most recently updated RUNNING Task.  If there are not RUNNING Tasks, give it to the most recently update STAGING Task.  Otherwise, generate no DNS address.

I think it's a good point that a STAGING Task probably will not be able to respond to clients, and if a better alternative is available in a RUNNING Task, it should be provided.

jdef: TASK_STAGING could be a while if it's depending on a slow docker pull (see
below for ambiguity w/ respect to STAGING). We have a 5m timeout by default
for k8s-mesos and that's still not enough for some users large docker
images. To some degree the framework might want to play a role in
determining advertisement. For example, in DiscoveryInfo perhaps a
framework could apply a label to a Port such as
"mesos-dns/advertise-always" if those service instances that are unhealthy
should be advertised to the cluster anyway.

For the "bootstrap-clustered-systems" use case there's often a different
peer port vs. client port; port labels could help distinguish advertising
requirements here.

The default advertisement scheme for service instance ports for mesos-dns
could could assume "advertise only for healthy tasks".

Also, some tasks transition from STAGING to RUNNING. Others go from STAGING
to STARTING to RUNNING. There's no way that mesos-dns is going to be able
to reliably determine the meaning of these transitions across multiple
frameworks. E.g. a task in STAGING could be in the middle of a docker-pull.
Or it could be trying to connect to its peers in the cluster. Totally
depends on the framework in question.

gabrielhartmann: This sounds like the right way to go.  Perhaps Frameworks should provide a prioritized list of states.  In the HDFS case it would be [RUNNING, STAGING].  Meaning provide the latest RUNNING TASK if available, or fallback to STAGING, or provide nothing.

In any case, this more nuanced scheme is way outside the scope of this PR.  I personally don't have the cycles to implement and test such a scheme.  There's definitely more design work to be done too.  Did you want me to pull the discussion from here and put it in a proposal issue or something?

jdef:
@gabrielhartmann is this something that we still need?

gabrielhartmann:
I've been working around it for a long time now...  I guess that means it's not needed.

Issue #294, opened 2015-10-01
Proposal: Enumeration


tsenart:
In order to enable humans or machines to discover available services (not service endpoints), we should support enumeration in DNS (via PTR records) and HTTP.

tony-kerz:
+1

errordeveloper:
It'd be great to have a public API for this, however I wonder if there's a low-level way to obtain all records?

jdef:
Enumeration is supported via HTTP: https://github.com/mesosphere/mesos-dns/blob/master/docs/docs/http.md#get-v1enumerate

PR #309, opened 2015-10-07
Implement service enumeration

tony-kerz:
per https://github.com/mesosphere/mesos-dns/issues/294

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,11 @@ install it in order to build from source.
 
 ```shell
 $ go get github.com/tools/godep
+
+$ mkdir -p $GOPATH/src/github.com/mesosphere
+$ cd $GOPATH/src/github.com/mesosphere
+$ git clone git@github.com:{account-you-forked-from}/mesos-dns.git
+$ cd mesos-dns/
 $ godep go build ./...
 ```
 
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -34,6 +34,13 @@ type Resolver struct {
 	extResolver exchanger.Exchanger
 }
 
+type serviceRecord struct {
+	Service string `json:"service"`
+	Host    string `json:"host"`
+	IP      string `json:"ip"`
+	Port    string `json:"port"`
+}
+
 // New returns a Resolver with the given version and configuration.
 func New(version string, config records.Config) *Resolver {
 	r := &Resolver{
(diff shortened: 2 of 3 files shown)


tony-kerz:
@tsenart this is my first pass. i closed it after i realized that i am returning a list of {service, host, ip, port} which probably isn't what you want, as you specified [PTR](http://help.dnsmadeeasy.com/managed-dns/dns-record-types/ptr-record/) type info v [SRV](http://help.dnsmadeeasy.com/managed-dns/dns-record-types/srv-record/). 

can you provide an example of the json you would expect to see in a response?

best,
tony.

tsenart:
@tony-kerz: This being a DNS server, the primary enumeration mechanism I had in mind was the one described in https://tools.ietf.org/html/rfc6763#section-9. We could start with the HTTP API for such functionality but that feels backwards.

tony-kerz:
i see what you are saying, makes perfect sense in that context. appreciate the insight, thx!

jdef:
Not all client will be able to speak DNS. I'm fine starting out with the
HTTP approach to gauge usefulness of this feature. The whole reason that we
added the HTTP API was because many clients/libraries have difficulty
speaking anything but the most simple aspects of the DNS protocol.

tsenart:
@jdef, @tony-kerz: Is this particular feature, enumeration, targeted at programs or humans? If it is targeted at humans, as I think it is , then documentation of the DNS API ought to be sufficient. Any sysadmin can learn that `$ dig _services._dns-sd._udp.mesos. PTR` will return the available services.

tony-kerz:
in my particular case, i wanted to go after a service, but i wasn't sure what the service-name was, so i was just wishing for a mechanism to show available services. i was logged onto an aws-dcos-slave-node, and i have `curl` at my disposal to hit the mesos-dns-api, but to my knowledge `dig` isn't around by default.

jdef:
Are we really proceeding down the dns-sd path? I forwarded you an internal
email that went around a while back about dns-sd. There were some serious
doubts back then concerning its viability as discovery mechanism for mesos
clusters.

jdef:
That's precisely one of the problems that we were trying to target -- not
all clients have DNS tooling available. Almost all clients can speak HTTP.

tsenart:
Just to be clear, I'm *not* at all against exposing an HTTP endpoint for enumeration as I understand the limitations of available tools in most environments. However, I *do* want that the DNS and HTTP interfaces to have parity.

tsenart:
I'm sorry if that wasn't clear from my first comment.

sargun:
@tony-kerz -
I'd love for you to update your PR, and ideally split it it into two. Your notes about getting started with development are great, but I wish they were in their own PR.

sargun:
Record enumeration seems to fulfill most of these needs

PR #316, opened 2015-10-08
Remove bogus "NS request" log line

tsenart:


jdef:
about about changing it to verbose, or very-verbose? it's probably of interest for some people to see this for debugging purposes

tsenart:
@jdef: How is this of interest? I don't understand. It's very unstructured logging which just fills up and an artifact of the past: 7ece6ca9ae15f6316c23226010b08e1a9bfc8118

Issue #314, opened 2015-10-08
Load testing


tsenart:
In order to find obvious bottlenecks and get a good picture of performance characteristics.

tsenart:
With appropriately high `ulimits`, on the following machine:
```
  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro11,5
  Processor Name:	Intel Core i7
  Processor Speed:	2,8 GHz
  Number of Processors:	1
  Total Number of Cores:	4
  L2 Cache (per Core):	256 KB
  L3 Cache:	6 MB
  Memory:	16 GB
```

We get:
```
Statistics:

  Queries sent:         11647379
  Queries completed:    10889583 (93.49%)
  Queries lost:         692266 (5.94%)
  Queries interrupted:  65530 (0.56%)

  Response codes:       NOERROR 10889583 (100.00%)
  Average packet size:  request 47, response 170
  Run time (s):         177.587661
  Queries per second:   61319.479848

  Average Latency (s):  0.043051 (min 0.000049, max 0.172032)
  Latency StdDev (s):   0.011778
```

From:

```shell
$ cat config.json
{
  "masters": ["127.0.0.1:5050"],
  "refreshSeconds": 60,
  "ttl": 60,
  "domain": "mesos",
  "ns": "ns1",
  "port": 53,
  "resolvers": ["8.8.4.4"],
  "timeout": 5,
  "listener": "0.0.0.0",
  "SOARname": "ns1.mesos",
  "SOAMname": "root.ns1.mesos",
  "SOARefresh": 60,
  "SOARetry":   600,
  "SOAExpire":  86400,
  "SOAMinttl": 60,
  "dnson": true,
  "httpon": true,
  "httpport": 8123,
  "externalon": true,
  "recurseon": true
}
$ sudo ./mesos-dns -config=config.json &
$ cat in.txt
master0.mesos. ANY
chronoswithaspaceandmixedcase-2.0.1.mesos. ANY
_liquor-store._tcp.marathon.slave.mesos. ANY
_car-store._tcp.marathon.slave.mesos. SRV
chronos.marathon.slave.mesos. A
_reviewbot._tcp.marathon.mesos. SRV
nginx.marathon.mesos. A
$ dnsperf -d in.txt -n 10000000 -q 600000 -Q 600000 -S 5 -t 15
```

CPU profiling didn't reveal any serious bottleneck in the critical path.

PR #317, opened 2015-10-08
resolver: Fix data races

tsenart:
While performing the load test for #314, two data races manifested.
This commit fixes both of them.

### DIFF
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/emicklei/go-restful"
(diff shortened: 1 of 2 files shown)


resolver/resolver.go
```diff
@@ -38,7 +39,8 @@ func New(version string, config records.Config) *Resolver {
 		version: version,
 		config:  config,
 		rs:      &records.RecordGenerator{},
-		rng:     rand.New(rand.NewSource(time.Now().UnixNano())),
+		// See: https://github.com/golang/go/issues/3611
+		rng:     rand.New(&lockedSource{src: rand.NewSource(time.Now().UnixNano())}),
```
sttts: Are there multiple seed calls? Why is that a race? Feel very strange to lock a rand source. This should be a singleton, injected by the main function of the binary or by tests.

tsenart: I left a URL there for a reason :-)

tsenart: And why should this be a singleton? I avoid global state whenever possible.

sttts: Reading through https://github.com/golang/go/issues/3611 clarified what you are trying to do. With some explaining text in the commit message it would be fine. Placing a random link somewhere in the code is not nice.

sttts: With some clearer documentation in the commit and/or in the code why we need this, lgtm.

PR #320, opened 2015-10-08
Implement optional ZK detection timeout with resets on watch loss

tsenart:
As a follow up of the discussion in https://github.com/mesosphere/mesos-dns/pull/306

### DIFF
--- a/docs/docs/configuration-parameters.md
+++ b/docs/docs/configuration-parameters.md
@@ -35,6 +35,18 @@ The configuration file should include the following fields:
 
 `zk` is a link to the Zookeeper instances on the Mesos cluster. Its format is `zk://host1:port1,host2:port2/mesos/`, where the number of hosts can be one or more. The default port for Zookeeper is `2181`. Mesos-DNS will monitor the Zookeeper instances to detect the current leading master. 
 
+`zkDetectionTimeout` defines how long to wait (in seconds) for Zookeeper to report a new leading Mesos master.
+This timeout is activated on:
+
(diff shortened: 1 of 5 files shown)


docs/docs/configuration-parameters.md
```diff
@@ -35,6 +35,8 @@ The configuration file should include the following fields:
 
 `zk` is a link to the Zookeeper instances on the Mesos cluster. Its format is `zk://host1:port1,host2:port2/mesos/`, where the number of hosts can be one or more. The default port for Zookeeper is `2181`. Mesos-DNS will monitor the Zookeeper instances to detect the current leading master. 
 
+`zkDetectionTimeout` defines how long in seconds to wait for Zookeeper to return the active Mesos master addresses. Default is 30 and 0 means no timeout.
```
jdef: I'd be little more explicit here: defines how long to wait (sec) for Zookeeper to report a new leading Mesos master. This timeout is taken into consideration at:
* mesos-dns startup time, where it plays the role of the "initial leader detection timeout"
* mesos failure time, if/when the master processes are recycled and there is no leader for some period of time
* ZK failure time, if/when mesos-dns loses connection to the ZK cluster
* network failure time, see above

If a non-zero timeout is specified and the timeout threshold is exceeded before a new leading mesos master is reported by the ZK-based master detector, the system will abort.

jdef:
minor comment, otherwise lgtm

sepiroth887:
Love it! Much cleaner :)

PR #330, opened 2015-10-19
Remove answers to fit 512 byte when truncating

### LINKED ISSUE #326: possible malformed response with large responses
```
[root@ip-10-0-1-234 ~]# tcpdump -vvv -s 0 -l -n port 53
20:17:51.513349 IP (tos 0x0, ttl 64, id 56497, offset 0, flags [DF], proto UDP (17), length 74)
    10.0.1.234.37336 > 10.0.4.245.domain: [bad udp cksum 0x1b26 -> 0x84d3!] 34960+ SRV? _etcd-server._tcp.etcd.mesos. (46)
20:17:51.514077 IP (tos 0x0, ttl 64, id 25692, offset 0, flags [DF], proto UDP (17), length 740)
    10.0.4.245.domain > 10.0.1.234.37336: [udp sum ok] 34960*| q: SRV? _etcd-server._tcp.etcd.mesos. 9/0/9 _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-19237-s0.etcd.mesos.:1025 0 0, _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-43841-s3.etcd.mesos.:1026 0 0, _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-43841-s3.etcd.mesos.:1025 0 0, _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-55660-s1.etcd.mesos.:1025 0 0, _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-19237-s0.etcd.mesos.:1027 0 0, _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-55660-s1.etcd.mesos.:1027 0 0, _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-19237-s0.etcd.mesos.:1026 0 0, _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-55660-s1.etcd.mesos.:1026 0 0, _etcd-server._tcp.etcd.mesos. [1m] SRV etcd-server-43841-s3.etcd.mesos.:1027 0 0 ar: etcd-server-19237-s0.etcd.mesos. [1m] A 10.0.1.236, etcd-server-19237-s0.etcd.mesos. [1m] A 10.0.1.236, etcd-server-19237-s0.etcd.mesos. [1m] A 10.0.1.236, etcd-server-55660-s1.etcd.mesos. [1m] A 10.0.1.235, etcd-server-55660-s1.etcd.mesos. [1m] A 10.0.1.235, etcd-server-55660-s1.etcd.mesos. [1m] A 10.0.1.235, etcd-server-43841-s3.etcd.mesos. [1m] A 10.0.1.234, etcd-server-43841-s3.etcd.mesos. [1m] A 10.0.1.234, etcd-server-43841-s3.etcd.mesos. [1m] A 10.0.1.234 (712)
```
In particular, note the `length 740`.  This manifests as SRV record lookup failures in go.  Dig works, however, so it's not clear if we are mishandling large responses or if go's srv lookup code is buggy.  Go is able to resolve small responses, when I only query for _leader._tcp.mesos.  I'm working around this in etcd-mesos by providing discoveryinfo that will limit the size of the response.

discordianfish:
This should fix #326. This is roughly based on https://github.com/skynetservices/skydns/blob/4c00898b5ed8769af24f8dde381386f3969c2275/server/server.go#L175

### DIFF
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -395,6 +395,7 @@ func (res *Resolver) handleEmpty(rs *records.RecordGenerator, name string, m, r 
 // compressing the message first and truncating it accordingly.
 func reply(w dns.ResponseWriter, m *dns.Msg) {
 	m.Compress = true // https://github.com/mesosphere/mesos-dns/issues/{170,173,174}
+
 	if err := w.WriteMsg(truncate(m, isUDP(w))); err != nil {
 		logging.Error.Println(err)
 	}
@@ -405,11 +406,40 @@ func isUDP(w dns.ResponseWriter) bool {
 	return strings.HasPrefix(w.RemoteAddr().Network(), "udp")
 }
 
-// truncate sets the TC bit in the given dns.Msg if its length exceeds the
-// permitted length of the given transmission channel.
+// truncate removes answers until the given dns.Msg fits the permitted
+// length of the given transmission channel and sets the TC bit.
 // See https://tools.ietf.org/html/rfc1035#section-4.2.1
 func truncate(m *dns.Msg, udp bool) *dns.Msg {
-	m.Truncated = udp && m.Len() > dns.MinMsgSize
+	max := dns.MinMsgSize
+	if !udp {
+		max = dns.MaxMsgSize
+	} else if opt := m.IsEdns0(); opt != nil {
+		max = int(opt.UDPSize())
+	}
+
+	m.Truncated = m.Len() > max
+	if !m.Truncated {
+		return m
+	}
+
+	m.Extra = nil // Drop all extra records first
+	if m.Len() < max {
+		return m
+	}
+	answers := m.Answer[:]
+	left, right := 0, len(m.Answer)
+	for {
+		if left == right {
+			break
+		}
+		mid := (left + right) / 2
+		m.Answer = answers[:mid]
+		if m.Len() < max {
+			left = mid + 1
+			continue
+		}
+		right = mid
+	}
 	return m
 }
 
--- a/resolver/resolver_test.go
+++ b/resolver/resolver_test.go
@@ -1,8 +1,10 @@
 package resolver
 
 import (
+	"encoding/binary"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"math/rand"
 	"net"
@@ -74,7 +76,24 @@ func TestShuffleAnswers(t *testing.T) {
 }
 
 func TestHandlers(t *testing.T) {
-	res := fakeDNS(t)
+	if err := runHandlers(); err != nil {
+		t.Error(err)
+	}
+}
+
+func BenchmarkHandlers(b *testing.B) {
+	for n := 0; n < b.N; n++ {
+		if err := runHandlers(); err != nil {
+			b.Error(err)
+		}
+	}
+}
+
+func runHandlers() error {
+	res, err := fakeDNS()
+	if err != nil {
+		return err
+	}
 	res.fwd = func(m *dns.Msg, net string) (*dns.Msg, error) {
 		rr1, err := res.formatA("google.com.", "1.1.1.1")
 		if err != nil {
@@ -208,15 +227,18 @@ func TestHandlers(t *testing.T) {
 		var rw ResponseRecorder
 		tt.HandlerFunc(&rw, tt.Msg)
 		if got, want := rw.Msg, tt.Msg; !reflect.DeepEqual(got, want) {
-			t.Logf("Test #%d\n%v\n", i, pretty.Sprint(tt.Msg.Question))
-			t.Error(pretty.Compare(got, want))
+			return fmt.Errorf("Test #%d\n%v\n%s\n", i, pretty.Sprint(tt.Msg.Question), pretty.Compare(got, want))
 		}
 	}
+	return nil
 }
 
 func TestHTTP(t *testing.T) {
 	// setup DNS server (just http)
-	res := fakeDNS(t)
+	res, err := fakeDNS()
+	if err != nil {
+		t.Fatal(err)
+	}
 	res.version = "0.1.1"
 
 	res.configureHTTP()
@@ -273,7 +295,7 @@ func TestHTTP(t *testing.T) {
 	}
 }
 
-func fakeDNS(t *testing.T) *Resolver {
+func fakeDNS() (*Resolver, error) {
 	config := records.NewConfig()
 	config.Masters = []string{"144.76.157.37:5050"}
 	config.RecurseOn = false
@@ -284,21 +306,21 @@ func fakeDNS(t *testing.T) *Resolver {
 
 	b, err := ioutil.ReadFile("../factories/fake.json")
 	if err != nil {
-		t.Fatal(err)
+		return nil, err
 	}
 
 	var sj state.State
 	err = json.Unmarshal(b, &sj)
 	if err != nil {
-		t.Fatal(err)
+		return nil, err
 	}
 
 	spec := labels.RFC952
 	err = res.rs.InsertState(sj, "mesos", "mesos-dns.mesos.", "127.0.0.1", res.config.Masters, res.config.IPSources, spec)
 	if err != nil {
-		t.Fatal(err)
+		return nil, err
 	}
-	return res
+	return res, nil
 }
 
 func onError(abort <-chan struct{}, errCh <-chan error, f func(error)) <-chan struct{} {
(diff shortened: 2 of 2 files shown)


resolver/resolver.go
```diff
@@ -407,7 +407,37 @@ func isUDP(w dns.ResponseWriter) bool {
 // permitted length of the given transmission channel.
 // See https://tools.ietf.org/html/rfc1035#section-4.2.1
```
tsenart: We need to update this comment to reflect what the function is now doing.

resolver/resolver.go
```diff
@@ -407,7 +407,37 @@ func isUDP(w dns.ResponseWriter) bool {
...
+	switch udp {
+	case true:
+		if opt := m.IsEdns0(); opt != nil {
+			bufsize = int(opt.UDPSize())
+		}
+		if bufsize < dns.MinMsgSize {
+			bufsize = dns.MinMsgSize
+		}
+	case false:
+		bufsize = dns.MaxMsgSize - 1
```
tsenart: Why the ` - 1`?

resolver/resolver.go
```diff
@@ -407,7 +407,37 @@ func isUDP(w dns.ResponseWriter) bool {
 // permitted length of the given transmission channel.
 // See https://tools.ietf.org/html/rfc1035#section-4.2.1
 func truncate(m *dns.Msg, udp bool) *dns.Msg {
-	m.Truncated = udp && m.Len() > dns.MinMsgSize
+	bufsize := 0
+	switch udp {
```
tsenart:
```go
var max int
if max = dns.MinMsgSize; !udp {
    max = dns.MaxMsgSize
} else if opt := m.IsEdns0(); opt != nil {
    max = int(opt.UDPSize())
}

if m.Truncated = m.Len() > max; !m.Truncated {
    return m
}

for m.Len() > max {
   // ...
}
```

resolver/resolver.go
```diff
@@ -407,7 +407,37 @@ func isUDP(w dns.ResponseWriter) bool {
...
+			case len(m.Answer) > 0:
+				m.Answer = m.Answer[:len(m.Answer)-1]
+			case len(m.Ns) > 0:
+				m.Ns = m.Ns[:len(m.Ns)-1]
+			default:
+				panic("Couldn't truncate message to fit dns.MinMsgSize")
+			}
+			i++
+		}
+		logging.VeryVerbose.Printf("Message truncated, removed %d answers", i)
```
tsenart: I'd rather not mix logging with this business logic. We can do this outside this function.

resolver/resolver.go
```diff
@@ -407,7 +407,37 @@ func isUDP(w dns.ResponseWriter) bool {
...
+		for m.Len() > bufsize {
+			switch {
+			case len(m.Extra) > 0:
+				m.Extra = m.Extra[:len(m.Extra)-1]
+			case len(m.Answer) > 0:
+				m.Answer = m.Answer[:len(m.Answer)-1]
+			case len(m.Ns) > 0:
+				m.Ns = m.Ns[:len(m.Ns)-1]
+			default:
+				panic("Couldn't truncate message to fit dns.MinMsgSize")
```
tsenart: No panics please. Let's make the function return an error instead an handle it in the caller. Makes it easier to reason and test.

tsenart:
This change set needs tests. It'd be also good to benchmark how it performs. I'm curious :-)

discordianfish:
@tsenart I started working on tests but wanted to spin up a test cluster so I don't have to hand craft the state.json. Will do that tomorrow.

tsenart:
@discordianfish: You shouldn't have to do so. I'd like unit tests for this function.

tsenart:
@discordianfish: If you want, we can pair on it this week.

discordianfish:
I've added a unit test for the truncate function and some benchmarks. Probably not ideal, so happy for suggestions to improve it. Still need to build better intuition around what/how to test exactly.

Benchmark results:
- Without my changes
```
BenchmarkHandlers	    2000	    927056 ns/op
BenchmarkTruncate	2000000000	         0.30 ns/op
```

- With my changes
```
BenchmarkHandlers	    2000	   1022113 ns/op
BenchmarkTruncate	2000000000	         0.36 ns/op
```

resolver/resolver.go
discordianfish: I have no idea tbh, but it was introduced in skydns by @miekg and I assume he knows what he is doing :). @miekg: Care to comment? I agree that it's a bit unexpected.

resolver/resolver.go
discordianfish: Like this? It's pretty much what you suggested but without putting the assigments in the if clause. IMO that should be only done if you want the variable to be in the scope of the condition only.
(I also think that small, non-edns udp responses will always get truncated because max == 0 after line 7 in your suggestion, right?)

resolver/resolver.go
miekg: [ Quoting <notifications@github.com> in "Re: [mesos-dns] Remove answers to f..." ]
>> @@ -407,7 +407,37 @@ func isUDP(w dns.ResponseWriter) bool {
>>  // permitted length of the given transmission channel.
>>  // See https://tools.ietf.org/html/rfc1035#section-4.2.1
>>  func truncate(m *dns.Msg, udp bool) *dns.Msg {
>> -	m.Truncated = udp && m.Len() > dns.MinMsgSize
>> +	bufsize := 0
>> +	switch udp {
>> +	case true:
>> +		if opt := m.IsEdns0(); opt != nil {
>> +			bufsize = int(opt.UDPSize())
>> +		}
>> +		if bufsize < dns.MinMsgSize {
>> +			bufsize = dns.MinMsgSize
>> +		}
>> +	case false:
>> +		bufsize = dns.MaxMsgSize - 1
>
>I have no idea tbh, but it was introduced in skydns by @miekg and I assume he knows what he is doing :). @miekg: Care to comment? I agree that it's a bit unexpected.

Removing answers it not the right thing to do... I had some code that did this
for huge (>64K) answers being sent over TCP. But removing answers AND NOT
SETTING the truncated bit is *wrong*.

/Miek

--
Miek Gieben

discordianfish: @miekg I was referring to @tsenart's comment:

> Why the - 1?

I saw this in your code here: 
https://github.com/skynetservices/skydns/blob/4c00898b5ed8769af24f8dde381386f3969c2275/server/server.go#L205

Beside that, this code here sets the truncated bit when removing. But in the meanwhile I changed it based on @tsenart suggestions to make it more readable.

miekg: [ Quoting <notifications@github.com> in "Re: [mesos-dns] Remove answers to f..." ]
>> +	case false:
>> +		bufsize = dns.MaxMsgSize - 1
>
>@miekg I was referring to @tsenart's comment:
>
>> Why the - 1?
>
>I saw this in your code here:
>https://github.com/skynetservices/skydns/blob/4c00898b5ed8769af24f8dde381386f3969c2275/server/server.go#L205

I think that as because dns.MaxMsgSize was once 2^16, now it is defined as
2^16-1, so the -1 can be removed here.

>Beside that, this code here sets the truncated bit when removing. But in the meanwhile I changed it based on @tsenart suggestions to make it more readable.

Ack.

/Miek

--
Miek Gieben

sargun:
Can you rebase on master? There is a fix for 01f9eb1cb29a39d68f163319a63e7bf154f4d91b in master.

jdef:
performance of `truncate` degrades significantly as the number of results increases. Instead of using a linear search for a fit, could we use something along the lines of a binary search (O(log n))?

discordianfish:
@jdef Not sure what could be possibly more efficient in the truncate function than just looping over the answers until they fit the given size. You have some more specific suggestions?
Another way to optimize that would be to check if answers fit when adding/generating them. But that requires some significant change to the code I'd say.

discordianfish:
@jdef Ok, I've discussed this with @s-urbaniak and I now know what you mean. Will work on that.

discordianfish:
@jdef Okay, I've change it to a binary search but now only truncate answers to keep it simple. That's what skydns is doing as well. Performance is much better now:

```
BenchmarkHandlers	    2000	   1049782 ns/op
BenchmarkTruncate	   30000	     54031 ns/op
```

jdef:
ok will take a look

resolver/resolver.go
```diff
@@ -405,12 +410,40 @@ func isUDP(w dns.ResponseWriter) bool {
...
+	}
+
+	m.Truncated = m.Len() > max
+	if !m.Truncated {
+		return m, nil
+	}
+
+	m.Extra = nil // Drop all extra records first
+
+	answers := make([]dns.RR, len(m.Answer))
```
jdef: you shouldn't need to copy the slice contents here, instead create a snapshot of the slice (cap,len) that refers to the same underlying storage:
```go
answers := m.Answer[:]
```

resolver/resolver.go
```diff
@@ -405,12 +410,40 @@ func isUDP(w dns.ResponseWriter) bool {
...
+	} else if opt := m.IsEdns0(); opt != nil {
+		max = int(opt.UDPSize())
+	}
+
+	m.Truncated = m.Len() > max
+	if !m.Truncated {
+		return m, nil
+	}
+
+	m.Extra = nil // Drop all extra records first
```
jdef: It's possible that this is enough to bring the size of the message <= max; should we check here before doing the binary search?

jdef:
love the performance improvement. left additional review comments.

resolver/resolver.go
discordianfish: Good point, fixed that now.

resolver/resolver.go
discordianfish: Sounds good, changed!

resolver/resolver_test.go
```diff
@@ -338,3 +360,38 @@ func TestMultiError(t *testing.T) {
...
+
+func TestTruncate(t *testing.T) {
+	tm := newTruncated()
+	if !tm.Truncated {
+		t.Fatal("Message not truncated")
+	}
+	if l := tm.Len(); l > 512 {
+		t.Fatalf("Message to large: %d bytes", l)
+	}
+}
```
jdef: It might be nice to test that adding one additional answer record via `genA(1)` brings the message length over 512.

jdef:
minor test case comment, otherwise lgtm

resolver/resolver_test.go
discordianfish: Good idea, added that.

jdef:
LGTM :shipit:

discordianfish:
@jdef Ok cool! Not sure what the process is, can I just merge it now? Or do we need another reviewer?

jdef:
Safe to merge at this point. @sargun is the owner of this repo now, but we talked yesterday about this PR and he wants this fix in the next release.

PR #340, opened 2015-11-11
Move repo to Go 1.5

sargun:
-Add Go1.5 to the Circle-CI build environment
-Add documentation about the go1.5 build to README
-Make Circle-CI output build artifacts with go1.5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ tmp/
 config.json
 mesos-dns
 *.gz
+VERSION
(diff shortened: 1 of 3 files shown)


jdef:
dcos-image will probably need changes in order to build using go1.5

sargun:
@jdef dcos-image is building using Go1.5. Going forward, I'd like DCOS to use the artifacts that come from Circle-CI: https://circleci.com/gh/mesosphere/mesos-dns/472#artifacts

cmaloney:
@sargun we should discuss that offline. There are reasons I don't want to use those artifacts.

Issue #223, opened 2015-08-07
Tune errcheck CI settings to ignore `defer Close()` on RO objects


tsenart:
> Yes, and we should totally use defer whenever we can use it. It's my fault I didn't explain this change well enough. So, here's the deal: One of the linters in our CI is `errcheck` which detects ignored errors in Go programs. Unfortunately, it creates tension with the common `defer r.Close()` Go idiom. Now, errors returned by `Close` methods can be safely ignored *as long as* what was closed was read-only. Whenever there are writes involved, errors returned by close should absolutely not be ignored.

> What is lacking, then, is the appropriate tooling to handle such intricacies intelligently. In this PR, I did the easiest thing that would silence the CI failure, but I knew that it wasn't the right thing to do. I will follow up

sargun:
I think being explicit about this, and using the new `util.IgnoreError`, e.g. `defer util.IgnoreError(r.Close())` versus blindly ignoring these, to the linter accidentally dropping errors.

tsenart:
I dread any package called `util`. It ends up sucking up all matter from the known universe. I like the idea of making it explicit, but let's create a package called `errors` and have `defer errors.Ignore(r.Close)`

sargun:
@tsenart Isn't there already a std golang package named errors? I wouldn't want to have to import that as another name to avoid the conflict. Any other name suggestions?

jdef:
yes there is a stdlib package called errors but that shouldn't preclude us
from using the same name if it makes sense.

sargun:
@jdef errors is imported, and used in generator.go, which also uses this util func. I'd say, if we know there's going to be an import name collision from day 1, let's try to avoid that.

jdef:
I'd be fine with an `errorutil` package

sargun:
@jdef :+1:

tsenart:
`errutil`

sargun:
See: https://github.com/mesosphere/mesos-dns/pull/352

PR #341, opened 2015-11-12
Add Timeout when fetching state.json

### LINKED ISSUE #243: Add timeouts to HTTP /state.json requests
Currently these requests could hang for indeterminate time. https://github.com/mesosphere/mesos-dns/blob/master/records/generator.go#L106

sargun:
Fixes #243

### DIFF
--- a/config.json.sample
+++ b/config.json.sample
@@ -1,6 +1,7 @@
 {
   "zk": "zk://127.0.0.1:2181/mesos",
   "masters": ["127.0.0.1:5050"],
+  "stateTimeoutSeconds": 300,
   "refreshSeconds": 60,
   "ttl": 60,
   "domain": "mesos",
--- a/records/config.go
+++ b/records/config.go
@@ -23,6 +23,8 @@ type Config struct {
 	// Timeout is the default connect/read/write timeout for outbound
 	// queries
 	Timeout int
+	// Timeout in seconds waiting for the master to return data from StateJson
+	StateTimeoutSeconds int
 	// Zookeeper Detection Timeout: how long in seconds to wait for Zookeeper to
 	// be initially responsive. Default is 30 and 0 means no timeout.
 	ZkDetectionTimeout int
@@ -67,26 +69,27 @@ type Config struct {
 // NewConfig return the default config of the resolver
 func NewConfig() Config {
 	return Config{
-		ZkDetectionTimeout: 30,
-		RefreshSeconds:     60,
-		TTL:                60,
-		Domain:             "mesos",
-		Port:               53,
-		Timeout:            5,
-		SOARname:           "root.ns1.mesos",
-		SOAMname:           "ns1.mesos",
-		SOARefresh:         60,
-		SOARetry:           600,
-		SOAExpire:          86400,
-		SOAMinttl:          60,
-		Resolvers:          []string{"8.8.8.8"},
-		Listener:           "0.0.0.0",
-		HTTPPort:           8123,
-		DNSOn:              true,
-		HTTPOn:             true,
-		ExternalOn:         true,
-		RecurseOn:          true,
-		IPSources:          []string{"netinfo", "mesos", "host"},
+		ZkDetectionTimeout:  30,
+		RefreshSeconds:      60,
+		TTL:                 60,
+		Domain:              "mesos",
+		Port:                53,
+		Timeout:             5,
+		StateTimeoutSeconds: 300,
+		SOARname:            "root.ns1.mesos",
+		SOAMname:            "ns1.mesos",
+		SOARefresh:          60,
+		SOARetry:            600,
+		SOAExpire:           86400,
+		SOAMinttl:           60,
+		Resolvers:           []string{"8.8.8.8"},
+		Listener:            "0.0.0.0",
+		HTTPPort:            8123,
+		DNSOn:               true,
+		HTTPOn:              true,
+		ExternalOn:          true,
+		RecurseOn:           true,
+		IPSources:           []string{"netinfo", "mesos", "host"},
 	}
 }
 
@@ -138,6 +141,7 @@ func SetConfig(cjson string) Config {
 	logging.Verbose.Println("   - DnsOn: ", c.DNSOn)
 	logging.Verbose.Println("   - TTL: ", c.TTL)
 	logging.Verbose.Println("   - Timeout: ", c.Timeout)
+	logging.Verbose.Println("   - StateTimeoutSeconds: ", c.StateTimeoutSeconds)
 	logging.Verbose.Println("   - Resolvers: " + strings.Join(c.Resolvers, ", "))
 	logging.Verbose.Println("   - ExternalOn: ", c.ExternalOn)
 	logging.Verbose.Println("   - SOAMname: " + c.SOAMname)
--- a/records/generator.go
+++ b/records/generator.go
@@ -13,10 +13,12 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/mesosphere/mesos-dns/logging"
 	"github.com/mesosphere/mesos-dns/records/labels"
 	"github.com/mesosphere/mesos-dns/records/state"
+	"github.com/mesosphere/mesos-dns/util"
 )
 
 // Map host/service name to DNS answer
@@ -27,9 +29,16 @@ type rrs map[string][]string
 // RecordGenerator contains DNS records and methods to access and manipulate
 // them. TODO(kozyraki): Refactor when discovery id is available.
 type RecordGenerator struct {
-	As       rrs
-	SRVs     rrs
-	SlaveIPs map[string]string
+	As         rrs
+	SRVs       rrs
+	SlaveIPs   map[string]string
+	httpClient http.Client
+}
+
+// NewRecordGenerator returns a RecordGenerator that's been configured with a timeout.
+func NewRecordGenerator(httpTimeout time.Duration) *RecordGenerator {
+	rg := &RecordGenerator{httpClient: http.Client{Timeout: httpTimeout}}
+	return rg
 }
 
 // ParseState retrieves and parses the Mesos master /state.json and converts it
@@ -73,8 +82,7 @@ func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
 			logging.Error.Println(err)
 		}
 
-		sj, _ = rg.loadWrap(ip, port)
-		if sj.Leader != "" {
+		if sj, err = rg.loadWrap(ip, port); err == nil && sj.Leader != "" {
 			return sj, nil
 		}
 		logging.Verbose.Println("Warning: Zookeeper is wrong about leader")
@@ -91,13 +99,11 @@ func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
 			logging.Error.Println(err)
 		}
 
-		sj, _ = rg.loadWrap(ip, port)
-		if sj.Leader == "" {
+		if sj, err = rg.loadWrap(ip, port); err == nil && sj.Leader == "" {
 			logging.VeryVerbose.Println("Warning: not a leader - trying next one")
 			if len(masters)-1 == i {
 				return sj, errors.New("no master")
 			}
-
 		} else {
 			return sj, nil
 		}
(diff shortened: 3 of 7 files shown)


config.json.sample
```diff
@@ -1,6 +1,7 @@
 {
   "zk": "zk://127.0.0.1:2181/mesos",
   "masters": ["127.0.0.1:5050"],
+  "StateTimeout": 300,
```
jdef: probably better named `stateTimeoutSeconds` (lowercase, and express the unit)

records/config.go
```diff
@@ -23,6 +23,8 @@ type Config struct {
 	// Timeout is the default connect/read/write timeout for outbound
 	// queries
 	Timeout int
+	// Timeout waiting for the master to return data from StateJson
```
jdef: indicate the unit here, please (seconds)

records/generator.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strconv"
 	"strings"
 
+	"time"
```
jdef: nit: get rid of the newline above "time"

records/generator.go
```diff
@@ -27,9 +29,17 @@ type rrs map[string][]string
...
-	SRVs     rrs
-	SlaveIPs map[string]string
+	As         rrs
+	SRVs       rrs
+	SlaveIPs   map[string]string
+	httpClient http.Client
+}
+
+// NewRecordGenerator returns a RecordGenerator that's been configured with a timeout.
+func NewRecordGenerator(httpTimeout time.Duration) *RecordGenerator {
```
jdef: why pass in the timeout here? probably better off passing in a http.Client since there's likely to be other things that will want configuration in the future (e.g. SSL settings)

records/generator.go
```diff
@@ -57,8 +67,8 @@ func (rg *RecordGenerator) ParseState(c Config, masters ...string) error {
 
 // Tries each master and looks for the leader
 // if no leader responds it errors
-func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
-	var sj state.State
+func (rg *RecordGenerator) findMaster(masters ...string) (*state.State, error) {
```
jdef: why the conversion of `state.State` to `*state.State`?
- seems outside the scope of the goal of this PR
- pass-by-value isn't terribly expensive here since the bulk of the data is in slices, and those are copied by reference (cheap)

records/generator.go
```diff
@@ -120,54 +128,56 @@ func (rg *RecordGenerator) loadFromMaster(ip string, port string) (sj state.Stat
...
 
-	client := &http.Client{}
-	resp, err := client.Do(req)
+	resp, err := rg.httpClient.Do(req)
 	if err != nil {
 		logging.Error.Println(err)
+		return nil, err
 	}
 
 	body, err := ioutil.ReadAll(resp.Body)
```
jdef: should really have a `defer resp.Body.close()` here, before the `ReadAll` (and get rid of the one on line 148)

records/generator.go
```diff
@@ -108,9 +115,10 @@ func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
 }
 
 // Loads state.json from mesos master
-func (rg *RecordGenerator) loadFromMaster(ip string, port string) (sj state.State) {
+func (rg *RecordGenerator) loadFromMaster(ip string, port string) (*state.State, error) {
```
jdef: if you're going to return error's here, why bother logging them in this func? leave that up to the caller (and if the caller doesn't log (but also just passes the error back) then it's the grand-caller's job to log, so on and so forth).

records/generator.go
```diff
@@ -198,7 +208,7 @@ func hostToIP4(hostname string) (string, bool) {
 }
 
 // InsertState transforms a StateJSON into RecordGenerator RRs
-func (rg *RecordGenerator) InsertState(sj state.State, domain, ns, listener string, masters, ipSources []string, spec labels.Func) error {
+func (rg *RecordGenerator) InsertState(sj *state.State, domain, ns, listener string, masters, ipSources []string, spec labels.Func) error {
```
jdef: ditto .. no need for the `*state.State` change everywhere

records/generator_test.go
```diff
@@ -7,6 +7,12 @@ import (
 	"testing"
 	"testing/quick"
 
+	"net"
```
jdef: nit: these imports should be merged with those above

resolver/resolver.go
```diff
@@ -151,7 +154,7 @@ func (res *Resolver) Reload() {
 		atomic.StoreUint32(&res.config.SOASerial, timestamp)
 		res.rs = &t
 	} else {
-		logging.VeryVerbose.Println("Warning: master not found; keeping old DNS state")
+		logging.VeryVerbose.Printf("Warning: Error generating records: %v; keeping old DNS state", err)
```
jdef: why do we log an error condition as VeryVerbose?

records/generator.go
jdef: e.g.
```go
defer resp.Body.Close()
body, err := ioutil.ReadAll(resp.Body)
```

jdef:
initial review complete

records/generator.go
sargun: It's primarily to be able to provide a nil (Optional). I'd rather mesos-dns fail in an obvious way if somehow error checking fails, as opposed to returning an empty state, and the entire zone emptying out.

records/generator.go
sargun: See above.

records/generator.go
sargun: I'm still investigating cases where we might want to blow away the httpclient entirely (stale connections, etc..). Originally, I had it where if HTTP timeout was encountered, it'd blow away the old client. I'm not sure how legitimate this concern is, hence I didn't proceed.

config.json.sample
sargun: Okay. I'm going to set the field as StateTimeoutSeconds in JSON, rather than using a JSON Tag.

records/generator.go
sargun: Got rid of the error logging.

records/config.go
```diff
@@ -67,26 +69,27 @@ type Config struct {
 // NewConfig return the default config of the resolver
 func NewConfig() Config {
 	return Config{
```
sargun: Gofmt shifted everything over due to a longer field

records/generator.go
jdef: returning error objects upon an error condition is idiomatic Go. if err == nil then there's no error. having two possible error states to check (error object, or state == nil) is confusing and not intuitive.

sargun: I think my reasoning here was state == nil iff err != nil. If someone is checking the err condition, and bailing out, there should be no need for them to ever check whether state == nil. I see the benefit as if the value is somehow dereferenced, because error checking went awry, it'll crash. From my understanding, mesos-dns is typically run under a supervisor, which will restart it. I see this as a better alternative than loading in an incorrect state. I see this as a variant of crash-only computing.

If you look at the behaviour in the Go standard library, it appears to be the same: https://play.golang.org/p/aUUoHoNXnM. 

Although, I've noticed in our project, we've followed the semantic of preferring to load the zero value. If we'd like to continue that, I don't think it's detrimental.

config.json.sample
```diff
@@ -1,6 +1,7 @@
 {
   "zk": "zk://127.0.0.1:2181/mesos",
   "masters": ["127.0.0.1:5050"],
+  "stateTimeoutSeconds": 300,
```
sargun: One of the questions I have is why we didn't use strings, and time.Duration / time.ParseDuration? @jdef do you know?

records/generator.go
jdef: It really depends on the API. For example, the io.Reader api tells you to
check for data, regardless of whether an error is returned.

We don't need complicated API semantics here because we don't have
different types of errors to handle in different ways. If err != nil then I
think we shouldn't trust the returned state -- end of story.

I don't hate the change to *state.State -- I just don't think it's
necessary.

config.json.sample
jdef: I can't think of a good reason, probably because there isn't one.

records/generator.go
sargun: Understood.

config.json.sample
sargun: I created a related bug: mesos-dns #342

records/generator.go
```diff
@@ -152,22 +163,21 @@ func (rg *RecordGenerator) loadWrap(ip string, port string) (state.State, error)
...
-
+	sj, err = rg.loadFromMaster(ip, port)
+	if err != nil {
+		return state.State{}, err
+	}
 	if rip := leaderIP(sj.Leader); rip != ip {
 		logging.VeryVerbose.Println("Warning: master changed to " + ip)
-		sj = rg.loadFromMaster(rip, port)
+		sj, err = rg.loadFromMaster(rip, port)
+		if err != nil {
```
jdef: why do you need this `if`? can't you just `return sj, err`?

resolver/resolver.go
```diff
@@ -140,7 +142,8 @@ func (res *Resolver) SetMasters(masters []string) {
 // Reload triggers a new state load from the configured mesos masters.
 // This method is not goroutine-safe.
 func (res *Resolver) Reload() {
-	t := records.RecordGenerator{}
+	var t records.RecordGenerator
```
jdef: this is weird. just say `t := records.NewRecordGenerator(...)`

records/generator.go
```diff
@@ -120,28 +128,31 @@ func (rg *RecordGenerator) loadFromMaster(ip string, port string) (sj state.Stat
...
 
-	client := &http.Client{}
-	resp, err := client.Do(req)
+	resp, err := rg.httpClient.Do(req)
 	if err != nil {
 		logging.Error.Println(err)
+		return state.State{}, err
 	}
 
+	defer util.IgnoreError(resp.Body.Close)
```
jdef: this is an interesting idea. we normally don't care about close errors. not sure why we need to be explicit here with IgnoreError()

util/util.go
```diff
@@ -52,3 +52,22 @@ func logPanic(r interface{}) {
...
+// IgnoreError Calls an ErrorFunction, and ignores the result.
+// This allows us to be more explicit when there is no error
+// handling to be done, for example in defers
+func IgnoreError(f ErrorFunction) {
+	_ = f()
+}
+
+// PanicOnError Calls an ErrorFunction, and panics if the result
+// isn't nil.
+func PanicOnError(f ErrorFunction) {
```
jdef: nothing uses this, except for a test case. are there bigger plans for incorporating panics?

records/generator.go
jdef: I think it's recommended to reuse http.Client instances if possible. People that haven't done so have ended up with connection leaks

sargun: We weren't reusing the HTTP client before. Is there any reason that was the case? I personally don't know, but, if you're confident it's safe to do it, I'll make the change to have it passed in.

jdef: hm .. let's re-use it in a separate PR

records/generator.go
sargun: This is mostly a matter of dealing with the linters. Related: https://github.com/mesosphere/mesos-dns/issues/223 -- I don't think it's a good idea to ignore those errors completely. I think that being explicit about ignoring errors is valuable, and as laid out in the issue, we'd want to handle R/W objects differently.

records/generator.go
sargun: Will do - it's a simple change, but I think it should be looked at.

records/generator_test.go
```diff
@@ -164,6 +169,7 @@ func testRecordGenerator(t *testing.T, spec labels.Func, ipSources []string) Rec
 	if err != nil {
 		t.Fatal(err)
 	} else if err = json.Unmarshal(b, &sj); err != nil {
+		t.Log("err: ", err)
```
jdef: this is strange: we're already logging the error w/ t.Fatal(), why the additional t.Log()?

records/generator_test.go
```diff
@@ -267,3 +273,26 @@ func TestHashString(t *testing.T) {
...
+		notify := w.(http.CloseNotifier).CloseNotify()
+		<-notify
+	}
+	server := httptest.NewServer(http.HandlerFunc(sleepForeverHandler))
+	defer server.Close()
+
+	rg := NewRecordGenerator(500 * time.Millisecond)
+	host, port, err := net.SplitHostPort(server.Listener.Addr().String())
+	_, err = rg.loadFromMaster(host, port)
+	netErr := (err).(*url.Error).Err.(net.Error)
```
jdef: if might make for a more readable failure if there some additional assertions here:
- `if err == nil { f.Fatalf("expected error because ...") }`
- `urlErr, ok := ...; if !ok { t.Fatalf("expected url.Error instead of %#+v", err) }`

... etc

jdef:
nits: otherwise lgtm

records/generator_test.go
sargun: I have no idea how this snuck in. Getting rid of it.

jdef:
:shipit:

PR #347, opened 2015-11-13
Move to use Go's built in coverage command

sargun:
- Go coverage has been pulled in to go 1.5 (https://golang.org/cmd/cover/)
  - Remove fetching Go's coverage tool from code.google.com (deprecated)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ config.json
 mesos-dns
 *.gz
 VERSION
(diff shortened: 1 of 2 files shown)


sargun:
I don't see any obvious issues, but since we moved the repo to Go 1.5, we shouldn't need to pull in Google's external coverage tools.

Go cover still seems to work.

PR #353, opened 2015-11-16
fix docs typo

akitty:


sargun:
@akitty Can you actually open a new PR, reverting your change and fixing the whitespace at the end of the documentation, as well fixing your commit message to be more verbose?

PR #355, opened 2015-11-17
Use Mesos 0.26 version of NetworkInfo

Symmetric:
In MESOS-3788, the NetworkInfo message was updated to clarify
the structure of how IPs map onto interfaces.

Update mesos-dns to use the new NetworkInfo.

### DIFF
--- a/records/state/state.go
+++ b/records/state/state.go
@@ -59,9 +59,17 @@ type ContainerStatus struct {
 	NetworkInfos []NetworkInfo `json:"network_infos,omitempty"`
 }
 
-// NetworkInfo holds a network address resolution as defined in the
-// /state.json Mesos HTTP endpoint.
+// NetworkInfo holds the network configuration for a single interface
+// as defined in the /state.json Mesos HTTP endpoint.
 type NetworkInfo struct {
+	IPAddresses []IPAddress `json:"ip_addresses,omitempty"`
+	// back-compat with 0.25 IPAddress format
+	IPAddress string `json:"ip_address,omitempty"`
+}
+
+// IPAddress holds a single IP address configured on an interface,
+// as defined in the /state.json Mesos HTTP endpoint.
+type IPAddress struct {
 	IPAddress string `json:"ip_address,omitempty"`
 }
 
@@ -123,12 +131,22 @@ var sources = map[string]func(*Task) []string{
 func hostIPs(t *Task) []string { return []string{t.SlaveIP} }
 
 // networkInfoIPs returns IP addresses from a given Task's
-// []Status.ContainerStatus.[]NetworkInfos.IPAddress
+// []Status.ContainerStatus.[]NetworkInfos.[]IPAddresses.IPAddress
 func networkInfoIPs(t *Task) []string {
 	return statusIPs(t.Statuses, func(s *Status) []string {
 		ips := make([]string, len(s.ContainerStatus.NetworkInfos))
-		for i, netinfo := range s.ContainerStatus.NetworkInfos {
-			ips[i] = netinfo.IPAddress
+		for _, netinfo := range s.ContainerStatus.NetworkInfos {
+			if len(netinfo.IPAddresses) > 0 {
+				// In v0.26, we use the IPAddresses field.
+				for _, ipAddress := range netinfo.IPAddresses {
+					ips = append(ips, ipAddress.IPAddress)
+				}
+			} else {
+				// Fall back to v0.25 syntax of single IPAddress if that's being used.
+				if netinfo.IPAddress != "" {
+					ips = append(ips, netinfo.IPAddress)
+				}
+			}
 		}
 		return ips
 	})
--- a/records/state/state_test.go
+++ b/records/state/state_test.go
@@ -51,19 +51,44 @@ func TestTask_IPs(t *testing.T) {
 		{nil, []string{}, nil},
 		{nil, []string{"host"}, nil},
 		{ // no IPs for the given sources
-			Task: task(statuses(status(state("TASK_RUNNING"), netinfo("1.2.3.4")))),
+			Task: task(statuses(status(state("TASK_RUNNING"), netinfos(netinfo("1.2.3.4"))))),
 			srcs: []string{"host", "mesos"},
 			want: nil,
 		},
 		{ // unknown IP sources are ignored
-			Task: task(statuses(status(state("TASK_RUNNING"), netinfo("1.2.3.4")))),
+			Task: task(statuses(status(state("TASK_RUNNING"), netinfos(netinfo("1.2.3.4"))))),
 			srcs: []string{"foo", "netinfo", "bar"},
 			want: ips("1.2.3.4"),
+		},
+		{ // multiple IPs on a NetworkInfo
+			Task: task(statuses(status(state("TASK_RUNNING"), netinfos(netinfo("1.2.3.4"), netinfo("2.3.4.5"))))),
+			srcs: []string{"netinfo"},
+			want: ips("1.2.3.4", "2.3.4.5"),
+		},
+		{ // multiple NetworkInfos each with one IP
+			Task: task(statuses(status(state("TASK_RUNNING"), netinfos(netinfo("1.2.3.4", "2.3.4.5"))))),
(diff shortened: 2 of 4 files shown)


Symmetric:
@sargun rebased. Thanks!

sargun:
Looks good. The only thing I might want to do is grab a state.json that looks like it was produced by Mesos 0.26 for factories/fake.json. What do you think, @jdef?

jdef:
only if you're going to write unit tests to use it. at some point the
labels-for-ipaddress stuff will be removed from core and at that point
we'll *have* to migrate the unit tests. i'm fine leaving things as they are
for now, as long as someone on the mesos-dns team can test manually (which
seems possible using the docker containerizer) against v0.26-rc{whatever}.

sargun:
@jdef Any reason why we can't keep it backwards compatible? I think upgrading for $CUSTOMER would be really difficult otherwise - if they upgrade mesos master, they lose all their DNS records that use Docker. If they upgrade mesos-dns first, they also lose all of the records. 

It seems like it would be easy enough to leave the field in there, and write a note saying TODO: deprecate.

jdef:
DCOS releases are tied to a particular version of mesos+mesos-dns. so they
are, in essence, released together. so i'm not fully convinced it's worth
the overhead to add full backwards compat into this PR.

one could make the argument that some CM.x release will target some
pre-mesos-0.26 version and will want changes in the latest mesos-dns
release (with the exception of this PR). and in that case .. if we don't
want to maintain multiple branches then we'd be better off adding full
backwards compat and going through a deprecation cycle.

That said, CM.3 was just released. If CM.4 is going to target mesos-0.26
then I'm leaning towards accepting this PR as is. If CM.4 is going to
target mesos-0.25 then we probably need full backwards compat.

Somewhat related: many people in the community are basically adjusting
mesos-dns's config'd ipsources now because our default doesn't work with
mesos-0.25 and default docker IPs :(  An unfortunate oversight on our part.
0.26 will not fix it. I suspect that most people are going to shut off this
ipsource because of this problem. Those that aren't are likely using our
DCOS releases.

cmaloney:
CM.4 will be Mesos 0.25

jdef:
OK. Our options are: (a) don't upgrade mesos-dns for the CM.4 release, or;
(b) patch in full backwards compat as part of this PR so that we can
release an updated mesos-dns as part of CM.4, or; (c) leave the PR as is,
cut a special branch that supports v0.25 w/ select bug fixes for inclusion
in CM.4

sargun:
The thing is that even during the upgrade process to CM.5, there will be interruption. If the operator upgrades one master, they'll upgrade to Mesos 0.26, and mesos-dns v-next at the same time. When another mesos-master that is running Mesos 0.25 takes over, mesos-dns will go fetch state.json from that master, and be unable to interpret the state.json correctly due to it assuming that it'll be in the 0.26 format. 

If this short period of interruption of service is okay, while the Mesos masters are being upgraded in a rolling fashion, this is a-okay, but otherwise, I think it will prove somewhat problematic. 

Also, to @cmaloney's point, CM.4 is targeting Mesos 0.25, and I'd rather cut a special branch for CM.4.

jdef:
Not convinced that short-term service unavailability upgrades poses a
problem for $CUSTOMER. Should check w/ the DCOS customer-facing folks.

sargun:
@jdef Finding that out now.

sargun:
So, I talked to the customer-facing folks, and for the customers that are using NetworkInfo already, we need to maintain backward compatibility. @jdef We should fix this patch to be backwards compatible, and tests with 0.25 NetworkInfos, and 0.26 NetworkInfos.

Symmetric:
@sargun I've added back-compat support, and a few tests checking that it works. PTAL.

Symmetric:
(I'll squash once you've reviewed)

records/state/state.go
```diff
@@ -123,12 +131,22 @@ var sources = map[string]func(*Task) []string{
 func hostIPs(t *Task) []string { return []string{t.SlaveIP} }
 
 // networkInfoIPs returns IP addresses from a given Task's
-// []Status.ContainerStatus.[]NetworkInfos.IPAddress
+// []Status.ContainerStatus.[]NetworkInfos.[]IPAddresses.IPAddress
 func networkInfoIPs(t *Task) []string {
 	return statusIPs(t.Statuses, func(s *Status) []string {
 		ips := make([]string, len(s.ContainerStatus.NetworkInfos))
```
sargun: Can you add comments here explaining why this exists? And why we do this check?

sargun:
Looks good to me. The only thing I wish we could do is integration-style tests with a state.json extracted from a Mesos with NetworkInfo from 0.25, and 0.26.

sargun:
@jdef Do you want to take a look, and then we can spin up the squash, merge, cut, release, etc..

records/state/state.go
```diff
@@ -123,12 +131,22 @@ var sources = map[string]func(*Task) []string{
...
-		for i, netinfo := range s.ContainerStatus.NetworkInfos {
-			ips[i] = netinfo.IPAddress
+		for _, netinfo := range s.ContainerStatus.NetworkInfos {
+			if len(netinfo.IPAddresses) > 0 {
```
jdef: the `if len(..) > 0 {} else {}` is redundant. simplify to:
```go
for _, ipAddress := range netinfo.IPAddresses {
...
}
if netinfo.IPAddress != "" {
...
}
```

records/state/state_test.go
```diff
@@ -167,15 +192,35 @@ func state(st string) statusOpt {
...
+
+// NetworkInfo using v0.25 syntax for storing a single IP.
+func oldnetinfo(ip string) NetworkInfo {
+	netinfo := NetworkInfo{}
+	netinfo.IPAddress = ip
+	return netinfo
+}
+
+// NetworkInfo using both 0.25 and 0.26 syntax for IPs.
+func oldnewnetinfo(oldip string, newip string) NetworkInfo {
```
jdef: why? does mesos actually generate this? i thought there was no deprecation cycle between `ipaddress` and `ipaddresses` as far as mesos was concerned -- that it was a clean break.

jdef:
minor comments

records/state/state.go
Symmetric: I think that without the if/else, we'd pick the 0.25 syntax IP if both
versions were present for some reason - this way we pick the 0.26 version
in that case.

records/state/state_test.go
Symmetric: My motivation here was to check that we gracefully handle malformed input,
since either syntax is accepted, the server could provide both if there is
an error.

Seems that it doesn't cost much to verify that, but happy to remove it if
you prefer to keep this code simple.

records/state/state.go
jdef: understood. i was under the impression that no version of mesos would ever generate network-info that would contain both.

that said, this is really a minor comment - intent was to further simplify the code. if we're really worried about a botched mesos impl then we can keep this as is

records/state/state_test.go
jdef: see comments above in `networkInfoIPs()`

records/state/state.go
Symmetric: I'll defer to your judgement here; my instinct is to code defensively, but
I am not very familiar with the broader mesosphere ecosystem (i.e. what
could be generating the state.json) so if you think it's safe to simplify
then I'm happy to do so.

jdef: Hm, just re-reviewed the mesos-0.26-rc1 code:
https://github.com/apache/mesos/blob/0.26.0-rc1/include/mesos/mesos.proto#L1383

I didn't realize they were going through a deprecation cycle -- last I looked it was a clean break. Let's be defensive, keep the code as-is.

jdef:
@sargun lgtm

sargun:
@Symmetric Please squash, and I'll merge.

sargun:
@Symmetric Since this is no longer a breaking change, we can land it on master. Sorry for the switcharoo. I just wanted to keep it on a separate branch to avoid breaking master.

Symmetric:
Merged in https://github.com/mesosphere/mesos-dns/pull/360

Issue #365, opened 2015-11-24
We return NXDOMAIN incorrectly


sargun:
Version: Mesos-DNS 0.5.0

Dig directly against Mesos-DNS: 
```
sargun@ubuntu:/etc/bind$ dig @127.0.0.2 _externalapp._tcp.marathon.mesos.sfo.widgets.foo.; dig @127.0.0.2 _externalapp._tcp.marathon.mesos.sfo.widgets.foo. -t SRV

; <<>> DiG 9.9.5-11ubuntu1-Ubuntu <<>> @127.0.0.2 _externalapp._tcp.marathon.mesos.sfo.widgets.foo.
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 488
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0

;; QUESTION SECTION:
;_externalapp._tcp.marathon.mesos.sfo.widgets.foo. IN A

;; AUTHORITY SECTION:
_externalapp._tcp.marathon.mesos.sfo.widgets.foo. 5 IN SOA ns1.mesos.sfo.widgets.foo. admin.lol. 1448325464 60 600 86400 5

;; Query time: 1 msec
;; SERVER: 127.0.0.2#53(127.0.0.2)
;; WHEN: Mon Nov 23 16:37:47 PST 2015
;; MSG SIZE  rcvd: 115


; <<>> DiG 9.9.5-11ubuntu1-Ubuntu <<>> @127.0.0.2 _externalapp._tcp.marathon.mesos.sfo.widgets.foo. -t SRV
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 8084
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; QUESTION SECTION:
;_externalapp._tcp.marathon.mesos.sfo.widgets.foo. IN SRV

;; ANSWER SECTION:
_externalapp._tcp.marathon.mesos.sfo.widgets.foo. 5 IN SRV 0 0 31170 externalapp-4ikab-s0.marathon.slave.mesos.sfo.widgets.foo.

;; ADDITIONAL SECTION:
[...]

PR #366, opened 2015-11-24
Fix errant NXDOMAIN errors & return notimplemented on AAAA

sargun:
Related to #365. This is a work around. Although the pattern for all the other records is to add a handle$RECORDTYPE, unless we implement AAAA records, it'll always fall through to this function, which would have marked it with RcodeNameError. I think that RcodeSuccess is the most appriopriate response based upon:  https://tools.ietf.org/html/rfc4074.
I also looked at using RcodeNotImplemented, but it sounds like it's misbehaving according to that doc. 

This means that if someone searches for AAAA, they'll _always_ getting a positive cache for the record, and then they can do subsequent queries
CC: @jdef

### DIFF
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -378,7 +378,22 @@ func (res *Resolver) handleEmpty(rs *records.RecordGenerator, name string, m, r 
 	}
 
 	m.Rcode = dns.RcodeNameError
-	if qType == dns.TypeAAAA && len(rs.SRVs[name])+len(rs.As[name]) > 0 {
+
+	// Because we don't implement AAAA records, AAAA queries will always
+	// go via this path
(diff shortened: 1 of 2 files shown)


resolver/resolver.go
```diff
@@ -378,7 +378,8 @@ func (res *Resolver) handleEmpty(rs *records.RecordGenerator, name string, m, r
 	}
 
 	m.Rcode = dns.RcodeNameError
-	if qType == dns.TypeAAAA && len(rs.SRVs[name])+len(rs.As[name]) > 0 {
+
+	if (qType == dns.TypeAAAA) || (len(rs.SRVs[name])+len(rs.As[name]) > 0) {
```
probably worth a comment explaining why this behavior is the way that it is. including a link to the relevant RFC

sargun:
@jdef Added comments.

jdef:
lgtm

PR #367, opened 2015-11-24
Add some more process around releases

sargun:


sargun:
Just for notes -
I know that we can push the releases directly from Circle-CI to Github, but I very much do not want to do this _yet_, until I figure out a way to avoid accidental releases from private branches.

sargun:
PTAL / @jdef

jdef:
lgtm

PR #370, opened 2015-11-24
Add IPSources into config.json.sample, omitting NetInfo

sargun:


docs/docs/configuration-parameters.md
```diff
@@ -93,7 +93,7 @@ The configured name will always be converted to a FQDN by ensuring it ends with
 `enforceRFC952` will enforce an older, more strict set of rules for DNS labels. For details, see the [RFC-952](https://tools.ietf.org/html/rfc952). The default value is `false`.
 
 `IPSources` defines a fallback list of IP sources for task records,
-sorted by priority. The default value is: `["netinfo", "mesos", "host"]`
```
jdef: the default **coded** value hasn't changed (it's different than what's in config.json.sample, but that's not the same thing as "the default")

docs/docs/index.md
```diff
@@ -31,6 +31,11 @@ Note that the `hostname` field refers to the hostname used by the slave when it
 curl http://master_hostname:5050/master/state.json | python -mjson.tool
 ```
 
+### Mesos-DNS with Docker
+If you choose to use Mesos-DNS with Docker, with a version of Mesos after 0.25, be aware that there are some caveats. By default the Docker executor publishes the IP of the Docker container into the NetworkInfo field. Unfortunately, unless you're running some kind of SDN solution, bridged, or host networking with Docker, this can prove to make the containers unreachable.
+
+The default configuration that Mesos-DNS ships with in config.json.sample omits `netinfo` from the sources. The default options if you omit this field from the configuration includes `netinfo`. If you have trouble with Docker, ensure you check the IPSources field to omit netinfo.
```
jdef: :+1:

docs/docs/configuration-parameters.md
jdef: basically, leave "The default value is ..." sentence as-is here

jdef:
lgtm

PR #380, opened 2015-12-14
make EnforceRFC952=false the default

lloesche:
I've been asked several times by several people from different companies why mesos-dns has a 24 character limit. I'm wondering myself.

This PR makes RFC1123 (1989) the default instead of RFC952 (1985) thereby increasing the limit to 63 characters (which is what all modern DNS servers use).


jdef:
I'm confused. This should already be the default. Not sure why this PR is needed.

lloesche:
@jdef hm you're right. Let me go and find out why my DCOS cluster was restricted to 24 characters then. The option wasn't set to true in the config.

lloesche:
@jdef I think I know what's going on. The mesos-dns binary I'm using was from March 18th when the option didn't exist. I'll close the PR and make a new one that only adds the option to the sample config.

PR #381, opened 2015-12-14
add EnforceRFC952 to sample config

lloesche:
Option is available but currently absent from the sample config. People who get mesos-dns as a binary package are likely to check the sample config first for available options. So I think it makes sense to have all possible options in there.


jdef:
lgtm

PR #382, opened 2015-12-16
Fixes #377 - Apply label schema to task.DiscoveryInfo.Name

### LINKED ISSUE #377: Labeling schema should also be applied to task.DiscoveryInfo.Name
See:

https://github.com/mesosphere/mesos-dns/blob/e65fcabd5e5c0f7e21d65fd927914514e6d99a5f/records/generator.go#L380-L391

mesos-dns will apply `spec` on `task.Name`, but not on `task.DiscoveryInfo.Name`, this leads to inconsistent naming if a task specifies `DiscoveryInfo.Name`.

gkleiman:
It is necessary to treat task.DiscoveryInfo.Name in the same way as
task.Name, otherwise mesos-dns will use a different naming schema for
tasks that include DiscoveryInfo.

### DIFF
--- a/factories/fake.json
+++ b/factories/fake.json
@@ -223,7 +223,7 @@
                     "executor_id": "",
                     "framework_id": "20140703-014514-3041283216-5050-5348-0000",
                     "id": "liquor-store.b8db9f73-562f-11e4-a088-c20493233aa5",
-                    "name": "liquor-store",
+                    "name": "liquor.store",
                     "resources": {
                         "cpus": 1,
                         "disk": 0,
@@ -283,7 +283,7 @@
                     "discovery" : {
                         "version" : "1.0",
                         "visibility" : "CLUSTER",
-                        "name" : "liquor-store",
+                        "name" : "liquor.store",
                         "ports" : {
                             "ports" : [
                                 {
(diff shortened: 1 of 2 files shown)


sargun:
Thank you for your contribution. I’m not sure how I feel about ‘breaking’ the naming scheme in a point release. What do you think about materializing both DNS names with a comment to deprecate?

gkleiman:
@sargun Thank you for the review! I think that it makes sense to materialize both names until 1.0, but I am afraid that I can't work on this PR anymore.

Please feel free to take over the branch, or to wait until mesos-dns 1.0 to merge it "as is".

jdef:
I think this change makes a lot of sense (for consistency), at least until name generation is overhauled. I don't see a name generation rewrite being completed in the short term so I vote for getting this change in sooner than later.

jdef:
@sargun what's the ETA on this?

sargun:
The next bout of work on Mesos-DNS is going to happen a bit after Mesos 0.27 drops.

jdef:
it might also be nice to "fix" the schemas so that they don't translate `.` into `-` (or maybe we can generate both records for a release or two, to give people time to change over?)

Issue #61, opened 2015-01-22
SRV for multi-port applications


alberts:
Hey, Went to the meetup in SF tonight. Thanks for the presentation. IIRC (I could be wrong), in @guenter's demo he looked up a SRV of the form _foo.bar._tcp.mesos and it successfully resolved, but all the docs in this repo seem to contain examples that conform to RFC 2782, so maybe that was some kind of typo or old version of Mesos DNS.

Anyway, I was wondering if you've given any thought to how multi-port applications will be exposed via SRV? It would seem that a naming convention where the port becomes the service part, e.g., _port[012]._tcp.foo.bar.mesos could work.

Also, have you given any thought to how you might populate priorities? It might be nice to be able to tag tasks running in DC1 with priority 1 and tasks running in DC2 with priority 10, so that if all the DC1 nodes/tasks are unreachable (even if they are still in DNS), applications can know to switch to DC2. This is probably a bit of an edge case but it would be interesting to hear your thoughts.

We're keen to deploy Mesos DNS ASAP, hence the questions. Thanks for open sourcing this.

kozyraki:
Thanks for the feedback Albert. Let's separate issues here:

***Apps with multiple ports***:  If you have an app that listens to a number of ports (e.g., 8181 and 8282), we can create multiple SRV records. So, if you lookup the service name (e.g., _app._tcp.domain) you will get multiple answers, same IP for both, but different ports (8181 and 8282). I assume this is useful for people who know the service name but don't know the actual ports used (may not be the default ones). Let me know if this is useful for you. If someone knows the ports and wants this as part of the service name, then it is a different discussion (see below). 

***Naming scheme***: The current naming scheme in mesos-dns is simple and limited. My hope is that the 0.22.0 release of mesos will include the updates I made for Jira MESOS-2150. This will allow frameworks to pass much more interesting information to Mesos about how they want their task to be discovered (names, env/loc/ variables, general labels). Once we have this, we can support much more interesting names for both A and SRV records. My suspicion is that there is no single way to do this and we need to support some way for users to select. Assuming MESOS-2150, what kind of naming scheme(s) would work best for you? Give us some guidance.

mindscratch:
+1 for "Apps with multiple ports"

alberts:
[...]

PR #386, opened 2016-01-15
generate additional SRV records if port name is specified

jdef:
related to #61 

summary: if s task provides a discoveryInfo w/ named ports then mesos-dns will generate additional SRV records for those ports: `_{di-port-name}._{task-name}._{protocol}.{framework}.{domain}.`

### DIFF
--- a/records/chains.go
+++ b/records/chains.go
@@ -1,0 +1,80 @@
+package records
+
+import (
+	"github.com/mesosphere/mesos-dns/records/labels"
+)
+
+// chain is a generation func that consumes record-like strings and does
+// something with them
+type chain func(...string)
+
+const (
+	protocolNone = "" // for readability
+	domainNone   = "" // for readability
+)
+
+// withProtocol appends `._{protocol}.{framework}` to records. if protocol is "" then
+// the protocols "tcp" and "udp" are assumed.
+func withProtocol(protocol, framework string, spec labels.Func, gen chain) chain {
+	return func(records ...string) {
+		protocol = spec(protocol)
+		if protocol != protocolNone {
+			for i := range records {
+				records[i] += "._" + protocol + "." + framework
+			}
+		} else {
+			records = append(records, records...)
+			for i, j := 0, len(records)/2; j < len(records); {
+				records[i] += "._tcp." + framework
+				records[j] += "._udp." + framework
+				i++
+				j++
+			}
+		}
+		gen(records...)
+	}
+}
+
+// withSubdomains appends `.{subdomain}` (for each subdomain spec'd) to records.
+// the empty subdomain "" indicates to generate records w/o a subdomain fragment.
+func withSubdomains(subdomains []string, gen chain) chain {
+	if len(subdomains) == 0 {
+		return gen
+	}
+	return func(records ...string) {
+		var (
+			recordLen = len(records)
+			tmp       = make([]string, recordLen*len(subdomains))
+			offset    = 0
+		)
+		for s := range subdomains {
+			if subdomains[s] == domainNone {
+				copy(tmp[offset:], records)
+			} else {
+				for i := range records {
+					tmp[offset+i] = records[i] + "." + subdomains[s]
+				}
+			}
+			offset += recordLen
+		}
+		gen(tmp...)
+	}
+}
+
+// withNamedPort prepends a `_{discoveryInfo port name}.` to records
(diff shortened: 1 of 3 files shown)


sargun:
Technically RFC non-compliant, but RFCs be damned, productivity ftw.

jdef:
considering `_{portname}._sub._{taskname}._{protocol}` since it's just as expressive but probably more compat with existing tooling that understands dns-sd (not that we're super concerned about implementing dns-sd)

jdef:
xref https://tools.ietf.org/html/rfc6763

sargun:
I prefer the current patch. It's something we're already doing, and I think preserving the SRV-style records follows the principal of least surprise. If we choose to support dns-sd, I'd rather create both records.

jdef:
understood. i still modified the current patch to simply prepend a
_portname if there is one. so it's possible to get
_portname._taskname._protocol .. but still think that's ok

records/chains.go
```diff
@@ -0,0 +1,80 @@
...
+// the protocols "tcp" and "udp" are assumed.
+func withProtocol(protocol, framework string, spec labels.Func, gen chain) chain {
+	return func(records ...string) {
+		protocol = spec(protocol)
+		if protocol != protocolNone {
+			for i := range records {
+				records[i] += "._" + protocol + "." + framework
+			}
+		} else {
+			records = append(records, records...)
```
sargun: Is there any reason to reuse records, and do this in place? Does Golang actually care that much or optimize that much do this in place? Can we just create another slice, and populate that? I think that would make this code much easier to read.

records/chains.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+// withSubdomains appends `.{subdomain}` (for each subdomain spec'd) to records.
+// the empty subdomain "" indicates to generate records w/o a subdomain fragment.
+func withSubdomains(subdomains []string, gen chain) chain {
+	if len(subdomains) == 0 {
+		return gen
+	}
+	return func(records ...string) {
```
sargun: Same with above

records/chains.go
```diff
@@ -0,0 +1,80 @@
...
+		var (
+			recordLen = len(records)
+			tmp       = make([]string, recordLen*len(subdomains))
+			offset    = 0
+		)
+		for s := range subdomains {
```
sargun: and reason not to do `for _, domain := range subdomains`?

records/chains.go
jdef: this *can* be more efficient if the underlying capacity of the slice is
large enough to hold twice the record count. in this particular instance it
depends how golang is allocating the initial slice for varargs parameter.
experiments on playground suggest that it allocates a slice w/ a capacity
equivalent to the # of variadic params.

even so, I'm loath to changing this line: it's concise and idiomatic.

records/chains.go
jdef: it probably doesn't matter for this case. in general, i'm trying to be more conservative with `range` statements over slices that use two iteration vars (index, value) because i've recently been bitten several times by the fact that `value` ends up being a copy of what's in the slice. so while it probably doesn't matter here, i don't see any reason to change the code - it's still very readable

sargun:
So, I like FP, but the lambda passing doesn't seem as idiomatic (in go) that I hoped for. I found the code really hard to read, but I don't think that's a matter of code-quality, I think that it's just a matter of statements like:
```

// withNamedPort prepends a `_{discoveryInfo port name}.` to records
func withNamedPort(portName string, spec labels.Func, gen chain) chain {
	portName = spec(portName)
	if portName == "" {
		return gen
	}
	return func(records ...string) {
		// generate without port-name prefix
		gen(records...)

		// generate with port-name prefix
		for i := range records {
			records[i] = "_" + portName + "." + records[i]
		}
		gen(records...)
	}
}
```
-- and since it's passing around lambdas that all have the same type signature / name, and doing all of the manipulations of data types in place, it makes it difficult to follow. I'm not saying we should change it, I'm just wondering if we have a better way to structure this flow. 

I'm curious as to whether using goroutines as generators, and channels to wire them all up might be a better option. I know that's less efficient, but as a general pattern I like it, and since you're only processing one item at a time, it should be memory efficient, and not require interesting scheduling.

sargun:
lgtm.

jdef:
thanks @sargun. bombs away

PR #392, opened 2016-02-03
Update tutorial to be general. Closes #388

### LINKED ISSUE #388: Google Compute Engine cluster provisioning tool is no longer active
The Google cluster provisioning tool mentioned here is deprecated and soon to be turned off: http://mesosphere.github.io/mesos-dns/docs/tutorial-gce.html

This tutorial should be rewritten to no longer mention it.

ssk2:
Remove mentions of Mesosphere for GCP and GCP specific things. This tutorial now relies on the user having a Mesos and Marathon cluster set up using the guide on open-docs. 

cc @joel-hamill


docs/docs/tutorial.md
```diff
@@ -1,36 +1,26 @@
...
-</p>
-
-[Google Cloud Platform](https://cloud.google.com) blocks port `53` by default. To unblock traffic for port `53` follow these [directions](http://stackoverflow.com/questions/21065922/how-to-open-a-specific-port-such-as-9090-in-google-compute-engine). Note that you need to unblock port `53` for both `tcp` and `udp` traffic. At the end, if you examine the firewall rules for your cluster you have a rule such as the one listed below, which opens port `53` for `udp` and `tcp` traffic from hosts in the `10.0.0.0` subnet. 
-
-<p class="text-center">
-  <img src="{{ site.baseurl}}/img/example-firewall-gce.png" width="350"  alt="">
-</p>
+Have the IP addresses of your cluster to hand, since we'll be installing and running Mesos DNS by sshing into your nodes. Make sure that port `53` is unblocked for both `tcp` and `udp` traffic between your nodes. To unblock traffic for port `53` on Google Compute Engine, follow these [directions](http://stackoverflow.com/questions/21065922/how-to-open-a-specific-port-such-as-9090-in-google-compute-engine).
 
 ### Step 2: Build and install Mesos-DNS
```
jdef: We should probably change this section to avoid building Mesos-DNS since we ship stable binaries -- people should just use those instead of building their own.

docs/docs/tutorial.md
```diff
@@ -91,33 +83,33 @@ $ more mesos-dns.json
...
+curl -X POST -H "Content-Type: application/json" http://1.2.3.4:8080/v2/apps -d@mesos-dns.json
 ```
 
-This command instructs Marathon to launch Mesos-DNS on node `10.14.245.208`. The `-v` option allows you to capture detailed warning/error logs for Mesos-DNS that may be useful for debugging. For example, Mesos-DNS will now periodically print in `stdout` information about all the A and SRV records it servers. This can be useful if you are not sure about how tasks and frameworks are named in your setup. You can access the `stdout` and `stderr` for Mesos-DNS through the Mesos webUI, accessible through `http://10.41.40.151:5050` in this example. 
+This command instructs Marathon to launch Mesos-DNS on node `2.3.4.5`. The `-v` option allows you to capture detailed warning/error logs for Mesos-DNS that may be useful for debugging. For example, Mesos-DNS will now periodically print in `stdout` information about all the A and SRV records it servers. This can be useful if you are not sure about how tasks and frameworks are named in your setup. You can access the `stdout` and `stderr` for Mesos-DNS through the Mesos webUI, accessible through `http://1.2.3.4:5050` in this example.
```
jdef: Most of this paragraph doesn't make sense, starting with "The `-v` option...". It starts making more sense at "You can access..."

docs/docs/tutorial.md
```diff
@@ -239,7 +231,7 @@ Commercial support is available at
 Now, let's use Mesos-DNS to communicate with nginx. We will still use the master node:
 
 ```
-ssh jclouds@10.41.40.151
+ssh 1.2.3.4
 ```
 
 First, let's do a DNS lookup for nginx, using the expected name `nginx.marathon-0.7.6.mesos`. The version number of Marathon is there because it registed with Mesos using name `marathon-0.7.6`. We could have avoided this by launching Marathon using ` --framework_name marathon`:
```
jdef: does marathon still register it's name with a version number like this?

ssk2:
@jdef - updated!

Issue #390, opened 2016-02-03
Cleanup insertRR


sargun:
insertRR currently will generate a massive amount of CPU if there are a large number of tasks. Part of this is due to the exists function, which has a naive loop per slave to check for colliding records. We need to do performance testing, and improvement for high-task count clusters.

jdef:
it would be great if we could somehow decide which tasks were not worth
adding to the recordset at all. maybe if discovery-info.visibility ==
"none" ?

cmaloney:
On a Mesos cluster with 8k tasks, runtime.memeqbody starts taking ~22% in a `perf top`, and `github.com/mesosphere/mesos-dns/records.(*RecordGenerator).exists` takes ~9%. The cluster had 10 agents, 800 tasks per agent. @sargun has some ideas / did some digging and found a lot more calls to the runtime.memeqbody than needed. Sent you the state.json

jdef:
thinking that it should be easy to covert the records container from a
slice to a `map[string]struct{}`. we're not really concerned about order
any way. maps are just as easy to iterate over and queries for exact
matches is much faster

jdef:
fixed by #395

PR #395, opened 2016-02-11
improve performance of record generation

jdef:
also, eliminate redundant A records in SRV additional-records section

#390

/cc @sargun @cmaloney

### DIFF
--- a/records/generator.go
+++ b/records/generator.go
@@ -25,7 +25,53 @@ import (
 // Map host/service name to DNS answer
 // REFACTOR - when discoveryinfo is integrated
 // Will likely become map[string][]discoveryinfo
-type rrs map[string][]string
+type rrs map[string]map[string]struct{}
+
+func (r rrs) add(name, host string) bool {
+	if host == "" {
+		return false
+	}
+	v, ok := r[name]
+	if !ok {
+		v = make(map[string]struct{})
+		r[name] = v
+	} else {
+		// don't overwrite existing values
+		_, ok = v[host]
+		if ok {
+			return false
+		}
+	}
+	v[host] = struct{}{}
+	return true
+}
+
+func (r rrs) First(name string) (string, bool) {
+	for host := range r[name] {
+		return host, true
+	}
+	return "", false
+}
+
+type rrsKind string
+
+const (
+	// A record types
+	A rrsKind = "A"
+	// SRV record types
+	SRV = "SRV"
+)
+
+func (kind rrsKind) rrs(rg *RecordGenerator) rrs {
+	switch kind {
+	case A:
+		return rg.As
+	case SRV:
+		return rg.SRVs
+	default:
+		return nil
+	}
+}
 
 // RecordGenerator contains DNS records and methods to access and manipulate
 // them. TODO(kozyraki): Refactor when discovery id is available.
@@ -253,10 +299,10 @@ func (rg *RecordGenerator) frameworkRecords(sj state.State, domain string, spec 
 		host, port := f.HostPort()
 		if address, ok := hostToIP4(host); ok {
 			a := fname + "." + domain + "."
-			rg.insertRR(a, address, "A")
+			rg.insertRR(a, address, A)
 			if port != "" {
-				srv := net.JoinHostPort(a, port)
-				rg.insertRR("_framework._tcp."+a, srv, "SRV")
+				srvAddress := net.JoinHostPort(a, port)
+				rg.insertRR("_framework._tcp."+a, srvAddress, SRV)
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


jdef:
circle ci failure:
```
gometalinter --cyclo-over=12 --tests --deadline=120s ./...

records/generator.go:59:2:warning: exported const A should have comment (or a comment on this block) or be unexported (golint) gometalinter --cyclo-over=12 --tests --deadline=120s ./... returned exit code 1
```

records/generator_test.go
```diff
@@ -24,6 +24,24 @@ func init() {
...
 
+func (rg *RecordGenerator) exists(name, host, rtype string) bool {
+	if rtype == "A" {
+		if val, ok := rg.As[name]; ok {
+			// check if A record already exists
+			// identical tasks on same slave
+			_, ok := val[host]
+			return ok
+		}
+	} else {
```
sargun: should we check if the rtype here is SRV?

records/generator_test.go
```diff
@@ -244,8 +262,16 @@ func TestInsertState(t *testing.T) {
 		{rgDocker.As, "nginx.marathon.mesos.", []string{"1.2.3.11"}},
 		{rgDocker.As, "car-store.marathon.slave.mesos.", []string{"1.2.3.11"}},
 	} {
-		if got := tt.rrs[tt.name]; !reflect.DeepEqual(got, tt.want) {
-			t.Errorf("test #%d: %q: got: %q, want: %q", i, tt.name, got, tt.want)
+		// convert want into a map[string]struct{} for simpler comparison
+		want := map[string]struct{}{}
```
sargun: This seems confusing to me, although not explicitly part of this, you're touching near here - an explanation how why this code looks like this would be really handy.

sargun: Specifics:
-Why are we converting the datastructure?
-What are we actually testing for?

sargun:
A couple comments + golinter being mad at you, but looks fine otherwise.

records/generator_test.go
jdef: sure we could. I think I just ported this from the non-test code

records/generator_test.go
jdef: we're converting want into a string set so that we can use `reflect.DeepEqual`

jdef: we're converting want into a string set so that we can use `reflect.DeepEqual`

sargun: Sweet.

records/generator.go
```diff
@@ -529,58 +575,30 @@ func (rg *RecordGenerator) setFromLocal(host string, ns string) {
...
-	enumRecord := EnumerableRecord{Name: name, Host: host, Rtype: rtype}
-	enumTask.Records = append(enumTask.Records, enumRecord)
-	return rg.insertRR(name, host, rtype)
-}
-func (rg *RecordGenerator) insertRR(name, host, rtype string) bool {
-	if host == "" || rg.exists(name, host, rtype) {
-		return false
+func (rg *RecordGenerator) insertTaskRR(name, host string, kind rrsKind, enumTask *EnumerableTask) bool {
+	if rg.insertRR(name, host, kind) {
+		enumRecord := EnumerableRecord{Name: name, Host: host, Rtype: string(kind)}
```
jdef: I "fixed" the logic here: only add an enumerable record if the insertRR succeeds

jdef: @sargun

sargun: :+1:

records/generator.go
```diff
@@ -253,10 +299,10 @@ func (rg *RecordGenerator) frameworkRecords(sj state.State, domain string, spec
 		host, port := f.HostPort()
 		if address, ok := hostToIP4(host); ok {
 			a := fname + "." + domain + "."
-			rg.insertRR(a, address, "A")
+			rg.insertRR(a, address, A)
 			if port != "" {
 				srv := net.JoinHostPort(a, port)
-				rg.insertRR("_framework._tcp."+a, srv, "SRV")
+				rg.insertRR("_framework._tcp."+a, srv, SRV)
```
sargun: This is a little weird with srv, SRV -- I like above where we did address, maybe srvAddress?

records/generator_test.go
```diff
@@ -24,6 +24,25 @@ func init() {
 	logging.SetupLogs()
 }
 
+func (rg *RecordGenerator) exists(name, host, rtype string) bool {
```
sargun: Since we added the enum for rtype, is there a reason not to use the enum here?

resolver/resolver_test.go
```diff
@@ -235,13 +229,66 @@ func runHandlers() error {
...
 		tt.HandlerFunc(&rw, tt.Msg)
-		if got, want := rw.Msg, tt.Msg; !reflect.DeepEqual(got, want) {
+		if got, want := rw.Msg, tt.Msg; !(Msg{got}).equivalent(Msg{want}) {
 			return fmt.Errorf("Test #%d\n%v\n%s\n", i, pretty.Sprint(tt.Msg.Question), pretty.Compare(got, want))
 		}
 	}
 	return nil
 }
 
+type Msg struct{ *dns.Msg }
```
sargun: :+1:

sargun:
Circle-CI seems a little bit unhappy right now. Let's rerun things later and hope it feels better.

records/generator_test.go
jdef: laziness. I can fit it

sargun:
Rad. lgtm.

jdef:
great. will squash and merge

PR #399, opened 2016-02-25
Fix #377 - Use spec() on DiscoveryInfo records

### LINKED ISSUE #377: Labeling schema should also be applied to task.DiscoveryInfo.Name

sargun:


sargun:
@jdef

records/generator.go
```diff
@@ -363,78 +363,100 @@ func (rg *RecordGenerator) listenerRecord(listener string, ns string) {
...
+		slaveIDTail(task.SlaveID),
+		task.IP(ipSources...),
+		task.SlaveIP,
+	}
 
-			rg.insertRR(arec+tail, ctx.taskIP, "A")
-			rg.insertRR(canonical+tail, ctx.taskIP, "A")
+	// use DiscoveryInfo name if defined instead of task name
+	if task.HasDiscoveryInfo() {
+		// LEGACY TODO: REMOVE
```
jdef: this doesn't look right (this wasn't there in the old code)

jdef: compare #382

records/generator_test.go
```diff
@@ -196,6 +196,8 @@ func TestInsertState(t *testing.T) {
 		name string
 		want []string
 	}{
+		{rg.As, "big-dog.marathon.mesos.", []string{"10.3.0.1"}},
```
jdef: where are the SRV's?

jdef:
This looks broken to me. Some additional context in the description/header of this PR would be very nice.

records/generator_test.go
sargun: Added the SRV records to the test.

records/generator.go
sargun: Basically what I did break up the `tasksRecords` function into three functions -

`taskRecord` - Meant to operator on a single task. Before there was the usage of anonymous struct, ctx, for passing around informaiton about how a record should be built up, so I defined it as an independent type. Now, tasksRecord passed the task + framework to tasksRecord to taskRecord iif it was RUNNING, and we had an IP for the slave it was running on.

records/generator.go
```diff
@@ -363,78 +363,100 @@ func (rg *RecordGenerator) listenerRecord(listener string, ns string) {
...
-		fname := labels.DomainFrag(f.Name, labels.Sep, spec)
 
-		// insert taks records
-		tail := "." + domain + "."
 		for _, task := range f.Tasks {
 			var ok bool
 			task.SlaveIP, ok = rg.SlaveIPs[task.SlaveID]
 
-			// skip not running or not discoverable tasks
-			if !ok || (task.State != "TASK_RUNNING") {
```
sargun: Previously we would bail, and now we check for both.

records/generator.go
```diff
@@ -363,78 +363,100 @@ func (rg *RecordGenerator) listenerRecord(listener string, ns string) {
...
+	slaveID,
+	taskIP,
+	slaveIP string
+}
 
-			// use DiscoveryInfo name if defined instead of task name
-			if task.HasDiscoveryInfo() {
-				ctx.taskName = task.DiscoveryInfo.Name
-			}
+func (rg *RecordGenerator) taskRecord(task state.Task, f state.Framework, domain string, spec labels.Func, ipSources []string) {
```
sargun: Task Record just defines the context, and passes it to taskContextRecord.

records/generator.go
```diff
@@ -363,78 +363,100 @@ func (rg *RecordGenerator) listenerRecord(listener string, ns string) {
...
-				slaveTarget := slaveHost + ":" + port
+}
+func (rg *RecordGenerator) taskContextRecord(ctx context, task state.Task, f state.Framework, domain string, spec labels.Func) {
+	fname := labels.DomainFrag(f.Name, labels.Sep, spec)
 
-				if !task.HasDiscoveryInfo() {
-					rg.insertRR(tcpName+tail, slaveTarget, "SRV")
-					rg.insertRR(udpName+tail, slaveTarget, "SRV")
-				}
+	tail := "." + domain + "."
```
sargun: Github Diff formatting seems to have brokens with a combination of whitespace changes + reorganization of functions.

records/generator_test.go
```diff
@@ -208,6 +210,14 @@ func TestInsertState(t *testing.T) {
 		{rg.As, "slave.mesos.", []string{"1.2.3.10", "1.2.3.11", "1.2.3.12"}},
 		{rg.As, "some-box.chronoswithaspaceandmixe.mesos.", []string{"1.2.3.11"}}, // ensure we translate the framework name as well
 		{rg.As, "marathon.mesos.", []string{"1.2.3.11"}},
+		{rg.SRVs, "_big-dog._tcp.marathon.mesos.", []string{
+			"big-dog-4dfjd-0.marathon.mesos.:80",
+			"big-dog-4dfjd-0.marathon.mesos.:443",
+		}},
+		{rg.SRVs, "_big-dog2._tcp.marathon.mesos.", []string{
```
jdef: I think this is the (unintended?) behavior change. the "LEGACY" block you added causes this to be created. Prior to this PR the `big-dog2` task name is replaced by the `big-dog` discovery info name.

What's the justification for this change in behavior?

sargun:
This was the solve the problem of both having the old behaviour when discoveryInfo was found as part of a Mesos task record (to put the record, denormalized) into DNS. What I've done is made to sure it adds both the normalized, and denormalized records. The statement in question is here:
```

			// use DiscoveryInfo name if defined instead of task name
			if task.HasDiscoveryInfo() {
				ctx.taskName = task.DiscoveryInfo.Name
			}
```
It should be:

```

			// use DiscoveryInfo name if defined instead of task name
			if task.HasDiscoveryInfo() {
				ctx.taskName = spec(task.DiscoveryInfo.Name)
			}
```

Unfortunately, this could break people's existing installations, so I split out the function into the first part which is just initialization, and loops. The second part, which is setting up the context, and the third part which is creating the records. If discovery info is detected, I do call the record generation component, before and after spec has been called on DiscoveryInfo.

records/generator.go
jdef: if you're going for backwards compat, then you're missing a line here in the LEGACY block, prior to invoking taskContextRecord: `ctx.taskName = task.DiscoveryInfo.Name`

records/generator_test.go
sargun: The idea is to continue to provide the existing behaviour + new behaviour:

```

		// LEGACY TODO: REMOVE
		rg.taskContextRecord(ctx, task, f, domain, spec)
		// LEGACY, TODO: REMOVE

		ctx.taskName = spec(task.DiscoveryInfo.Name)
		rg.taskContextRecord(ctx, task, f, domain, spec)
```

So, when the Marathon team adds a bunch of DiscoveryInfo entries, the old ones still exist.

jdef: I think we're talking past each other. In my mind when there's discovery
info if you want to preserve the old behavior then you need ctx.taskName =
task.DiscoveryInfo.Name before taskContextRecord. for the new behavior, set
ctx.taskName=spec(task.DiscoveryInfo.Name) and call taskContextRecord again.

sargun: Ah, okay, I think I misunderstood the original ask.

records/generator.go
sargun: Changed it to look more like described.

jdef: great. can we set some deadline for when we're allowed to remove the legacy block?

jdef:
can we incorporate the changes from the fake JSON in #382 and see how tests are impacted?

sargun:
Completed.

jdef:
lgtm.

when can we remove the legacy block?

sargun:
Whenever someone looks at this ticket after June: https://github.com/mesosphere/mesos-dns/issues/400

PR #404, opened 2016-03-01
Add basic enumeration of records to mesos-dns

sargun:


resolver/resolver.go
```diff
@@ -82,6 +82,14 @@ func exchangers(timeout time.Duration, protos ...string) map[string]exchanger.Ex
 	return exs
 }
 
+// Returns the currents set of enumerable data from the resolver
+// this data only makes sense to be read-only
+func (res *Resolver) enumerableData() records.EnumerationData {
+	res.rsLock.Lock()
```
jdef: stop referring to res.rs directly, get rid of the lock. this entire func is then reduced to `return res.records().EnumData`

jdef: at which point, is this func still needed?

records/generator.go
```diff
@@ -33,9 +33,36 @@ type RecordGenerator struct {
 	As         rrs
 	SRVs       rrs
 	SlaveIPs   map[string]string
+	EnumData   EnumerationData
 	httpClient http.Client
 }
 
+// EnumerableRecord is the lowest level object, and should map 1:1 with DNS records
```
jdef: wondering if we should start an `api` package for types like this that we expose?

sargun: I think that's a good idea to talk about in a different issue. (https://github.com/mesosphere/mesos-dns/issues/405). 

I'd rather go through the exercise of protobufing them, or using another schema language rather than forcing downstreams to pull in Go.

sargun:
Added the enumeration data.

records/config.go
```diff
@@ -64,6 +64,8 @@ type Config struct {
 	ExternalOn bool
 	// EnforceRFC952 will enforce an older, more strict set of rules for DNS labels
 	EnforceRFC952 bool
+	// Enumeration enabled via the API enumeration endpoint
+	EnumerationEnabled bool
```
jdef: nit: could we call it "EnumerationOn" for consistency?

resolver/resolver.go
```diff
@@ -498,6 +501,15 @@ func (res *Resolver) RestConfig(req *restful.Request, resp *restful.Response) {
 	}
 }
 
+// RestEnumerate handles HTTP requests of the enumeration data
+func (res *Resolver) RestEnumerate(req *restful.Request, resp *restful.Response) {
+
+	enumData := res.records().EnumData
```

resolver/resolver.go
```diff
@@ -466,6 +466,9 @@ func (res *Resolver) configureHTTP() {
 	ws.Route(ws.GET("/v1/hosts/{host}").To(res.RestHost))
 	ws.Route(ws.GET("/v1/hosts/{host}/ports").To(res.RestPorts))
 	ws.Route(ws.GET("/v1/services/{service}").To(res.RestService))
+	if res.config.EnumerationEnabled {
```

jdef:
nit: otherwise lgtm

PR #411, opened 2016-03-07
added BenchmarkTaskRecord_withoutDiscoveryInfo benchmark

jdef:
Tested before and after commit 0bf67246bdc605c352617e4ba07e55d3bbfcf118 that introduced functional record generation. No difference in record generation time. Captured sample micro-benchmark time below.
```
vagrant@node-1:~/mesos-dns-workspace/src/github.com/mesosphere/mesos-dns$ !go
go test -run='^$' -bench='^BenchmarkTaskRecord.*$' ./records/...
PASS
BenchmarkTaskRecord_withoutDiscoveryInfo          200000             10750 ns/op
...
```

### DIFF
--- a/records/generator_perf_test.go
+++ b/records/generator_perf_test.go
@@ -4,6 +4,9 @@ import (
 	"math/rand"
 	"strconv"
 	"testing"
+
+	"github.com/mesosphere/mesos-dns/records/labels"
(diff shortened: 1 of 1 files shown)


sargun:
LGTM. I like this.

PR #412, opened 2016-03-20
Make `state.json` fetching more robust

vespian:


records/generator.go
```diff
@@ -150,41 +150,44 @@ func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
 
 	// Check if ZK leader is correct
 	if leader != "" {
-		logging.VeryVerbose.Println("Zookeeper says the leader is: ", leader)
+		logging.VeryVerbose.Println("The leader is: ", leader)
 		ip, port, err := getProto(leader)
 		if err != nil {
 			logging.Error.Println(err)
+		} else {
+			if sj, err = rg.loadWrap(ip, port); err == nil && sj.Leader != "" {
```
jdef: shouldn't these conditions (`err==nil` and `sj.Leader != ""`) be tested separately, as they are below? the logic is so similar, it's actually begging to be consolidated.

records/generator.go
```diff
@@ -241,12 +244,21 @@ func (rg *RecordGenerator) loadWrap(ip string, port string) (state.State, error)
 	if err != nil {
 		return state.State{}, err
 	}
-	if rip := leaderIP(sj.Leader); rip != ip {
-		logging.VeryVerbose.Println("Warning: master changed to " + ip)
-		sj, err = rg.loadFromMaster(rip, port)
-		return sj, err
+	if sj.Leader != "" {
+		// FIXME(prozlach): Still, we are assuming that sj.Leader is a correct
```
jdef: `leaderIP` parses this format: `master@ip:port`. did the format change?

vespian: The format did not change, but if it changes due to i.e. some error condition, this will Bork again.

My point was that it would be nice to have state.json validated before we start processing it so that assumptions like that are caught early and can be quickly reported to Mesos team in case when they become invalid.

records/generator.go
vespian: True :) I just did not want to change the logic here too much.

Considering the changes in `loadWrap`, `sj.Leader != ""` will always be True as long as `err==nill`.

records/generator.go
```diff
@@ -150,41 +150,38 @@ func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
 
 	// Check if ZK leader is correct
 	if leader != "" {
-		logging.VeryVerbose.Println("Zookeeper says the leader is: ", leader)
+		logging.VeryVerbose.Println("The leader is: ", leader)
```
sargun: This should probably still be Zookeeper

vespian: :+1:

records/generator.go
```diff
@@ -241,12 +238,21 @@ func (rg *RecordGenerator) loadWrap(ip string, port string) (state.State, error)
...
+		if rip != ip {
+			logging.VeryVerbose.Println("Warning: master changed to " + ip)
+			sj, err = rg.loadFromMaster(rip, port)
+			return sj, err
+		}
+		return sj, nil
 	}
-	return sj, nil
+	msg := "Fetched state.json does not contain leader information"
+	logging.Error.Println(msg)
```
jdef: i wouldn't bother logging this here, let the caller log it if it's important

records/generator.go
```diff
@@ -150,41 +150,38 @@ func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
...
 
-		if sj, err = rg.loadWrap(ip, port); err == nil && sj.Leader == "" {
-			logging.VeryVerbose.Println("Warning: not a leader - trying next one")
-			if len(masters)-1 == i {
-				return sj, errors.New("no master")
-			}
-		} else {
-			return sj, nil
+		if sj, err = rg.loadWrap(ip, port); err != nil {
+			logging.Error.Println("Failed to fetch state.json - trying next one")
```
jdef: include the err content here

records/generator.go
```diff
@@ -150,41 +150,38 @@ func (rg *RecordGenerator) findMaster(masters ...string) (state.State, error) {
...
-		logging.VeryVerbose.Println("Zookeeper says the leader is: ", leader)
+		logging.VeryVerbose.Println("Zookeeper says, the leader is: ", leader)
 		ip, port, err := getProto(leader)
 		if err != nil {
 			logging.Error.Println(err)
+		} else {
+			if sj, err = rg.loadWrap(ip, port); err == nil {
+				return sj, nil
+			}
+			logging.Error.Println("Failed to fetch state.json from leader")
```
jdef: include the err content here

jdef:
sorry this has been sitting so long in review. just completed another pass. @sargun anything you want to add?

records/generator.go
sargun: Can we add a test with the "corrupt" state.json?

sargun:
It looks good. It'd be nice to add a test, but I realize that negative tests are difficult. Otherwise looks good.

vespian:
@sargun @jdef thanks for your comments, I will amend the code soonish  :)

sargun:
@vespian I'd like to get these changes in before cutting another release, which I'd like to do soon.

vespian:
ACK, will do my best.

sargun:
Closed

PR #429, opened 2016-05-18
Prevent /v1/enumerate from returning frameworks: null, or tasks: null

sargun:


records/generator.go
```diff
@@ -436,7 +442,10 @@ func (rg *RecordGenerator) listenerRecord(listener string, ns string) {
 
 func (rg *RecordGenerator) taskRecords(sj state.State, domain string, spec labels.Func, ipSources []string) {
 	for _, f := range sj.Frameworks {
-		enumerableFramework := &EnumerableFramework{Name: f.Name}
+		enumerableFramework := &EnumerableFramework{
+			Name: f.Name,
+			Tasks: []*EnumerableTask{},
```
jdef: can't you just use a json `omitempty` tag for `Tasks` in the struct where it's defined?

records/generator.go
```diff
@@ -111,7 +111,13 @@ type EnumerationData struct {
 
 // NewRecordGenerator returns a RecordGenerator that's been configured with a timeout.
 func NewRecordGenerator(httpTimeout time.Duration) *RecordGenerator {
-	rg := &RecordGenerator{httpClient: http.Client{Timeout: httpTimeout}}
+	enumData := EnumerationData{
+		Frameworks: []*EnumerableFramework{},
```
jdef: can't you just use a json `omitempty` tag for `Frameworks` in the struct where it's defined?

records/generator.go
sargun: We don't want to entirely omit it, we want it to be [] when there are no tasks.

records/generator.go
sargun: Nope, it should be [] on no frameworks, not null or missing entirely

jdef:
lgtm

PR #430, opened 2016-05-18
First cut at adding AXFR API to Mesos-DNS

sargun:
@jdef This is the first cut of adding an "AXFR" API to Mesos-DNS. This is an API to enable things like Spartan to mirror the Mesos-DNS zone, or manipulate it in an outside DNS server

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -1,0 +1,26 @@
+package models
+
+// AXFRResourceRecordSet is a representation of record name -> string
+type AXFRResourceRecordSet map[string][]string
+
+// Even though it would be nice to have the model map to something better than a string
+// (a la an actual SRV, or A struct).
+// This is the internal structure of how mesos-dns works today and the transformation of string -> DNS Struct
+// happens on actual query time. Why this logic happens at query time? Who knows.
+
+// AXFRRecords are the As, and SRVs that actually make up the Mesos-DNS zone
+type AXFRRecords struct {
+	As   AXFRResourceRecordSet
+	SRVs AXFRResourceRecordSet
+}
+
+// AXFR is a rough representation of a "transfer" of the Mesos-DNS data
+type AXFR struct {
+	TTL            int32  // DNS TTL according to config
+	Serial         uint32 // Current DNS zone version / serial number
+	RefreshSeconds int    // How often we try to poll Mesos for updates -- minimum downstream poll interval
+	Mname          string // primary name server
+	Rname          string // email of admin esponsible
+	Domain         string // Domain: name of the domain used (default "mesos", ie .mesos domain)
+	Records        AXFRRecords
+}
--- a/records/generator.go
+++ b/records/generator.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/mesosphere/mesos-dns/errorutil"
 	"github.com/mesosphere/mesos-dns/logging"
+	"github.com/mesosphere/mesos-dns/models"
 	"github.com/mesosphere/mesos-dns/records/labels"
 	"github.com/mesosphere/mesos-dns/records/state"
 	"github.com/tv42/zbase32"
@@ -25,6 +26,8 @@ import (
 // Map host/service name to DNS answer
 // REFACTOR - when discoveryinfo is integrated
 // Will likely become map[string][]discoveryinfo
+// Effectively we're (ab)using the map type as a set
+// It used to have the type: rrs map[string][]string
 type rrs map[string]map[string]struct{}
 
 func (r rrs) add(name, host string) bool {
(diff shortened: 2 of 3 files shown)


models/models.go
```diff
@@ -0,0 +1,26 @@
...
+// AXFRRecords are the As, and SRVs that actually make up the Mesos-DNS zone
+type AXFRRecords struct {
+	As   AXFRResourceRecordSet
+	SRVs AXFRResourceRecordSet
+}
+
+// AXFR is a rough representation of a "transfer" of the Mesos-DNS data
+type AXFR struct {
+	TTL            uint32
+	Serial         uint32
```
jdef: What is `Serial`? comments for each of the public fields would help :)

records/generator.go
```diff
@@ -25,6 +26,8 @@ import (
 // Map host/service name to DNS answer
 // REFACTOR - when discoveryinfo is integrated
 // Will likely become map[string][]discoveryinfo
+// Effectively we're (ab)using the map type as a set
+// It used to have the type: rrs map[string][]string
```
jdef: ... but values were converted to a "set" for higher efficiency

records/generator.go
```diff
@@ -53,6 +56,18 @@ func (r rrs) First(name string) (string, bool) {
 	return "", false
 }
 
+// Transform the record set into something exportable via the REST API
+func (r rrs) ToAXFRResourceRecordSet() models.AXFRResourceRecordSet {
+	ret := make(models.AXFRResourceRecordSet)
+	for host, values := range r {
+		ret[host] = []string{}
```
jdef: use `make([]string,0,len(values))` for efficiency

models/models.go
```diff
@@ -0,0 +1,26 @@
...
+	RefreshSeconds int
```
jdef: * everything else is a `uint32` here, why not this?
* why have both `RefreshSeconds` and `TTL`?

resolver/resolver.go
```diff
@@ -517,6 +519,29 @@ func (res *Resolver) RestEnumerate(req *restful.Request, resp *restful.Response)
...
+func (res *Resolver) RestAXFR(req *restful.Request, resp *restful.Response) {
+	records := res.records()
+
+	AXFRRecords := models.AXFRRecords{
+		SRVs: records.SRVs.ToAXFRResourceRecordSet(),
+		As:   records.As.ToAXFRResourceRecordSet(),
+	}
+	AXFR := models.AXFR{
+		Records:        AXFRRecords,
+		Serial:         atomic.LoadUint32(&res.config.SOASerial),
```
jdef: I see .. why not call `Serial` by its real name?

jdef: ugh nevermind, just a comment would be nice for public field

records/generator.go
```diff
@@ -53,6 +56,18 @@ func (r rrs) First(name string) (string, bool) {
 	return "", false
 }
 
+// Transform the record set into something exportable via the REST API
+func (r rrs) ToAXFRResourceRecordSet() models.AXFRResourceRecordSet {
+	ret := make(models.AXFRResourceRecordSet)
```
jdef: use `make(models.AXFRResourceRecordSet,len(r))`

resolver/resolver.go
```diff
@@ -517,6 +519,29 @@ func (res *Resolver) RestEnumerate(req *restful.Request, resp *restful.Response)
...
+		Mname:          res.config.SOAMname,
+		Rname:          res.config.SOARname,
+		TTL:            uint32(res.config.TTL),
+		RefreshSeconds: res.config.RefreshSeconds,
```
jdef: something like this is really non-dns-standard. should we come up with a way to more cleanly separate dns from non-dns standard things?

records/generator.go
sargun: I thought the values were converted to a set for higher efficiency during building? Is it that much overhead to transform it here?

models/models.go
sargun: I'll add the comment -- it's the version # for the zone. Borrowed from formatSOA. I was thinking of just calling formatSOA directly, and embedding the dns.SOA in the response. Do you think that might be better?

models/models.go
sargun: Taken from the config datastructure:
```
	SOASerial  uint32 // initial version number (incremented on refresh)
	TTL int32
	RefreshSeconds int
```
I'll make TTL an int32.

sargun: RefreshSeconds is a hint of refreshseconds from the config. When someone polls this endpoint, it hints to them to never poll it more often than refreshseconds. TTL tells someone that when they generate the zone, this is the TTL they should put on the records.

TTL should probably be something like 5 seconds if refreshseconds is ~60s or less, to get a decent interval for refreshing.

sargun: My counter-question is why are they different settings in the config file?

resolver/resolver.go
sargun: Any suggestions?

records/generator.go
jdef: i think you're right, important during the "build" phase (trying to avoid duplicate entries)

jdef:
lgtm after squash

models/models.go
jdef: don't remember

resolver/resolver.go
jdef: maybe wrapped in a `Mesos` or `Config` struct? don't have many good ideas here

PR #436, opened 2016-05-21
make HTTP listener address configurable

### LINKED ISSUE #435: HTTP frontend does not respect the `listener` option
Although the `listener` option [seems to be respected][1] when setting up a DNS server, it is [not taken into account][2] when setting up the HTTP frontend.

Either `listener` should be taken into account when it comes to the API, or a new `httplistener` option should be introduced.

At the moment, my workaround is to simply `"httpon": false`, which does have the downside of not having the HTTP frontend available.

[1]: https://github.com/mesosphere/mesos-dns/blob/135f7ac65a4baa14b8e5b7a69b9dc5cf47c1e869/resolver/resolver.go#L118
[2]: https://github.com/mesosphere/mesos-dns/blob/135f7ac65a4baa14b8e5b7a69b9dc5cf47c1e869/resolver/resolver.go#L490

jdef:
fixes #435

### DIFF
--- a/records/config.go
+++ b/records/config.go
@@ -53,8 +53,10 @@ type Config struct {
 	Domain string
 	// File is the location of the config.json file
 	File string
-	// ListenAddr is the server listener address
+	// Listen is the server DNS listener IP address
 	Listener string
+	// HTTPListen is the server HTTP listener IP address
+	HTTPListener string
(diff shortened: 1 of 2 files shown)


sargun:
👍

PR #441, opened 2016-06-09
Add support for zone-specific resolvers

timcharper:
This is a work-in-progress pull request.

This patch adds the ability to configure zone-specific upstream forwarders, which can help simplify the deployment of mesos-dns when other multiple dns providers are used (IE weave).

An example of the configuration could look like this:

```
{
  "zk": "zk://10.141.141.10:2181,10.141.141.11:2181,10.141.141.12:2181/mesos",
  "masters": ["10.141.141.10:5050","10.141.141.11:5050","10.141.141.12:5050"],
  "refreshSeconds": 60,
  "ttl": 60,
  "domain": "mesos",
  "ns": "ns1",
  "port": 53,
  "zoneResolvers": {"weave.local": ["172.17.0.1"]},
  "resolvers": ["8.8.8.8", "4.2.2.1"],
  "timeout": 5,
  "listener": "10.141.141.10",
  "SOAMname": "root.ns1.mesos",
  "SOARname": "ns1.mesos",
  "SOARefresh": 60,
  "SOARetry":   600,
  "SOAExpire":  86400,
  "SOAMinttl": 60,
  "dnson": true,
  "httpon": true,
  "httpport": 8123,
  "externalon": true,
  "recurseon": true
}
```

TODO (prior to merge):

- (done?) Mesosphere should agree that this is a good idea
- (done) ~~The config should perhaps be generalized such that general resolvers and zone resolvers can be specified in the same data structure. `Resolvers` could be deprecated and backwards-compatibility retained.~~ (kept Resolvers, enforced no "" ZoneResolver)
- (done) Tests should be added.
- (done) Documentation should be updated.

Please let me know what you think of this change. Having this feature reduces the complexity of our DNS deployment significantly; if this change is violently rejected, we'll probably have to deploy a bind server in front of mesos-DNS, which adds a degree of complexity to our infrastructure.

### DIFF
--- a/records/config.go
+++ b/records/config.go
@@ -48,6 +48,8 @@ type Config struct {
 	SOARname   string // email of admin esponsible
 	// Mesos master(s): a list of IP:port pairs for one or more Mesos masters
 	Masters []string
+	// DNS server: IP address of the DNS server for forwarded accesses
+	ZoneResolvers map[string][]string
 	// DNS server: a list of IP addresses or IP:port pairs for DNS servers for forwarded accesses
 	Resolvers []string
 	// IPSources is the prioritized list of task IP sources
@@ -105,6 +107,7 @@ func NewConfig() Config {
 		SOARetry:            600,
 		SOAExpire:           86400,
 		SOAMinttl:           60,
+		ZoneResolvers:       map[string][]string{},
 		Resolvers:           []string{"8.8.8.8"},
 		Listener:            "0.0.0.0",
 		HTTPListener:        "0.0.0.0",
@@ -146,6 +149,11 @@ func SetConfig(cjson string) Config {
 	}
 
 	c.Domain = strings.ToLower(c.Domain)
+
+	err = validateDomainName(c.Domain)
+	if err != nil {
+		logging.Error.Fatalf("%s is not a valid domain name", c.Domain)
+	}
 
 	c.initSOA()
 
@@ -190,6 +198,9 @@ func (c *Config) initResolvers() {
 		if err := validateResolvers(c.Resolvers); err != nil {
 			logging.Error.Fatal(err)
 		}
+		if err := validateZoneResolvers(c.ZoneResolvers, c.Domain); err != nil {
+			logging.Error.Fatal(err)
+		}
 	}
 }
 
@@ -202,6 +213,10 @@ func (c *Config) initSOA() {
 
 func (c Config) log() {
 	// print configuration file
+	zoneResolversJSON, err := json.Marshal(c.ZoneResolvers)
+	if err != nil {
+		zoneResolversJSON = []byte(fmt.Sprintf("error: %v", err))
+	}
 	logging.Verbose.Println("Mesos-DNS configuration:")
 	logging.Verbose.Println("   - Masters: " + strings.Join(c.Masters, ", "))
 	logging.Verbose.Println("   - Zookeeper: ", c.Zk)
@@ -215,6 +230,8 @@ func (c Config) log() {
 	logging.Verbose.Println("   - TTL: ", c.TTL)
 	logging.Verbose.Println("   - Timeout: ", c.Timeout)
 	logging.Verbose.Println("   - StateTimeoutSeconds: ", c.StateTimeoutSeconds)
+
+	logging.Verbose.Println("   - ZoneResolvers: " + string(zoneResolversJSON))
 	logging.Verbose.Println("   - Resolvers: " + strings.Join(c.Resolvers, ", "))
 	logging.Verbose.Println("   - ExternalOn: ", c.ExternalOn)
 	logging.Verbose.Println("   - SOAMname: " + c.SOAMname)
--- a/records/validation.go
+++ b/records/validation.go
@@ -3,8 +3,12 @@ package records
 import (
 	"fmt"
 	"net"
+	"regexp"
 	"strconv"
+	"strings"
 )
+
+var dnsValidationRegex = regexp.MustCompile(`^[a-zA-Z0-9-]+(\.[a-zA-Z0-9-]+)*$`)
 
 func validateEnabledServices(c *Config) error {
 	if !c.DNSOn && !c.HTTPOn {
@@ -49,6 +53,50 @@ func validateResolvers(rs []string) error {
 		return fmt.Errorf("Error validating resolvers: %v", err)
 	}
 	return nil
+}
+
+func validateDomainName(domain string) error {
+	if !dnsValidationRegex.MatchString(domain) {
+		return fmt.Errorf("Invalid domain name: %s", domain)
+	}
+	return nil
+}
+
+func validateZoneResolvers(zrs map[string][]string, mesosDomain string) (
+	err error) {
+
+	allDomains := make([]string, 0, len(zrs)+1)
+
+	for domain, rs := range zrs {
+		if len(rs) == 0 {
+			return fmt.Errorf("ZoneResolver %v is empty", domain)
+		}
+		err = validateDomainName(domain)
+		if err != nil {
+			return err
+		}
+
+		err = validateResolvers(rs)
+		if err != nil {
+			return
+		}
+		if domain == mesosDomain {
+			return fmt.Errorf("Can't specify ZoneResolver for Mesos domain (%v)",
+				mesosDomain)
+		}
+		allDomains = append(allDomains, "."+domain)
+	}
+	allDomains = append(allDomains, "."+mesosDomain)
+	for _, a := range allDomains {
+		for _, b := range allDomains {
+			if (a != b) &&
+				strings.HasSuffix(a, b) {
+				return fmt.Errorf("Ambiguous zone resolvers: %v is masked by %v",
+					a, b)
+			}
+		}
+	}
+	return
 }
 
 func normalizeResolver(hostPort string) (string, error) {
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -31,8 +31,9 @@ type Resolver struct {
 	rs               *records.RecordGenerator
 	rsLock           sync.RWMutex
 	rng              *rand.Rand
-	fwd              exchanger.Forwarder
 	generatorOptions []records.Option
+	zoneFwds         map[string]exchanger.Forwarder // map of zone -> forwarder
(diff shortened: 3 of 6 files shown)


sargun:
Are you a DCOS user or a Mesos user?

timcharper:
@sargun presently a Mesos user

timcharper:
I will fix the tests once it is decided that this is a change worth including.

sargun:
We have a DNS Proxy that runs on every machine in DC/OS. I wonder if you could look into running it on your infrastructure. https://github.com/dcos/spartan

timcharper:
Hmm, Spartan seems to have some pretty heavy dependencies (erlang). Also, it depends on Exhibitor, which I don't have deployed.

I'd prefer to not have to deploy another service, if possible. I'd be inclined to go with deploying a few bind servers if this suggestion is rejected.

sargun:
Erlang is pretty light weight. A release of Spartan is in the area of ~50MB, and  Spartan doesn't require Exhibitor.  You can disable the ZK record creator. 

You can also have a static master list: https://github.com/dcos/spartan/blob/master/src/spartan_config_loader_server.erl

* https://github.com/dcos/spartan/blob/master/src/spartan_config_loader_server.erl#L115
* https://github.com/dcos/spartan/blob/master/src/spartan_config_loader_server.erl#L118-L119
* https://github.com/dcos/spartan/blob/master/src/spartan_config_loader_server.erl#L126-L129

The reason behind Spartan is also to deal with DNS server availability issues -- If one Mesos-DNS goes down, it causes significant outages, because 1 second DNS timeouts are typically problematic. I'm not sure how you'll circumvent this.

timcharper:
We've installed DC/OS in a new cluster and I'm not sure if it's going to fit our needs. Spartan is getting in the way of Weave, which is something on which we presently rely and something for which there seems to be no substitute in DC/OS.

Additionally, zone-specific forwarding doesn't appear to be presently supported by Spartan:

https://github.com/dcos/spartan/blob/020357146ae0411177e353472608ab20a32c53bb/src/spartan_router.erl#L71

Can you consider re-opening this pull request? I think there's a good chance we'll stick with just mesos-dns, since it's enough for our needs, and the change I propose is really quite small.

jdef:
@sargun

timcharper:
I'll get the tests passing again.

timcharper:
@jdef the tests are passing again and I've added documentation.

exchanger/exchanger.go
```diff
@@ -26,6 +26,17 @@ func (f Func) Exchange(m *dns.Msg, addr string) (*dns.Msg, time.Duration, error)
 // A Decorator adds a layer of behaviour to a given Exchanger.
 type Decorator func(Exchanger) Exchanger
 
+var IgnoreErrTruncated Decorator = func(ex Exchanger) Exchanger {
```
sargun: Once you rebase, I think this should go away.

resolver/resolver.go
```diff
@@ -100,8 +108,17 @@ func (res *Resolver) records() *records.RecordGenerator {
 func (res *Resolver) LaunchDNS() <-chan error {
 	// Handers for Mesos requests
 	dns.HandleFunc(res.config.Domain+".", panicRecover(res.HandleMesos))
-	// Handler for nonMesos requests
-	dns.HandleFunc(".", panicRecover(res.HandleNonMesos))
+	// Handlers for nonMesos requests
+	for zone, fwd := range res.fwds {
```
sargun: What happens if someone has something like:
* sfo.ca
* sjc.ca
* .ca

resolver/resolver.go
```diff
@@ -57,11 +57,18 @@ func New(version string, config records.Config) *Resolver {
...
-	if !config.ExternalOn {
+	if config.ExternalOn {
+		r.fwds[""] = exchanger.NewForwarder(rs, exchangers(timeout, "udp", "tcp"))
+		for zone, resolvers := range config.ZoneResolvers {
+			r.fwds[zone] = exchanger.NewForwarder(resolvers, exchangers(timeout, "udp", "tcp"))
+		}
+
+	} else {
 		rs = rs[:0]
+		r.fwds[""] = exchanger.NewForwarder(rs, exchangers(timeout, "udp", "tcp"))
```
sargun: Move this out of the conditional?

sargun:
Any idea how much of a performance hit this is?

resolver/resolver.go
jdef: I think @sargun is asking for validation of the zone (probably near config-parsing time)

jdef: also probably also want to validate that no zone matches `res.config.Domain`

timcharper: We could sort them by length descending to ensure that the most specific matcher wins. I'm fine with validation that there are no overlaps, also.

timcharper:
@sargun if the feature isn't used, then the performance hit is 0. Otherwise, there is a a linear cost per query per zone resolver. I haven't measured the cost but I believe it is trivial.

resolver/resolver.go
timcharper: I added the validation to prevent overlapping domains.

timcharper:
I'm going to say that tests are adequate at this point; zone-specific routing -> handler function is a feature implemented by mkieg's DNS; I think testing that behavior in this project would be inappropriate.

records/validation.go
```diff
@@ -51,6 +52,42 @@ func validateResolvers(rs []string) error {
...
+		if domain == "" {
+			return fmt.Errorf("To specify the default ZoneResolver, use Resolvers")
+		}
+
+		err = validateResolvers(rs)
+		if err != nil {
+			return
+		}
+		if domain == mesosDomain {
+			return fmt.Errorf("Can't specify ZoneResolver for mesos domain (%v)",
```
sargun: "mesos" -> "Mesos"

timcharper:
as mentioned in #466 we're going to hold off the merge of this issue until the stable release goes out, after which point I'll rebase to the stable release and deploy the patch as is to our test environments, and ultimately production. I will report success of said effort here.

timcharper:
@sargun want to add a milestone for it?

sargun:
@timcharper Let's get this queued up next. Do you want to rebase, and then we can code review / merge?

timcharper:
@sargun sure! I rebased and squashed.

sargun:
@jdef What do you think?

records/config.go
```diff
@@ -202,6 +208,10 @@ func (c *Config) initSOA() {
 
 func (c Config) log() {
 	// print configuration file
+	zoneResolversJSON, err := json.Marshal(c.ZoneResolvers)
+	if err != nil {
+		zoneResolversJSON = []byte("error")
```
jdef: nit: it might be helpful to actually see the error message here. probably not likely that we'd every see one

records/validation.go
```diff
@@ -51,6 +52,42 @@ func validateResolvers(rs []string) error {
...
+func validateZoneResolvers(zrs map[string][]string, mesosDomain string) (
+	err error) {
+
+	allDomains := make([]string, 0, len(zrs)+1)
+
+	for domain, rs := range zrs {
+		if len(rs) == 0 {
+			return fmt.Errorf("ZoneResolver %v is empty", domain)
+		}
+		if domain == "" {
```
jdef: might be nice to have additional validation for `domain`

resolver/resolver.go
```diff
@@ -57,11 +57,16 @@ func New(version string, config records.Config) *Resolver {
...
-	if !config.ExternalOn {
+	if config.ExternalOn {
+		for zone, resolvers := range config.ZoneResolvers {
+			r.fwds[zone] = exchanger.NewForwarder(resolvers, exchangers(timeout, "udp", "tcp"))
+		}
+	} else {
 		rs = rs[:0]
 	}
-	r.fwd = exchanger.NewForwarder(rs, exchangers(timeout, "udp", "tcp"))
+	r.fwds[""] = exchanger.NewForwarder(rs, exchangers(timeout, "udp", "tcp"))
```
jdef: dislike magic constant `""` here. can we extract and call it something like `DefaultZone` or `NoZone` or whatever the semantic is?

resolver/resolver.go
```diff
@@ -101,8 +106,17 @@ func (res *Resolver) records() *records.RecordGenerator {
...
+	for zone, fwd := range res.fwds {
+		if zone != "" {
+			dns.HandleFunc(
+				zone+".",
+				panicRecover(res.HandleNonMesos(fwd)))
+		}
+	}
+	dns.HandleFunc(
+		".",
+		panicRecover(res.HandleNonMesos(res.fwds[""])))
```
jdef: ditto, magic constant

timcharper:
Okay @jdef ! All of the changes have been made. I've tested the build in my local mesos environment and have confirmed that the feature works, and the domain validation works, also.

jdef:
thanks @timcharper ! lgtm @sargun

resolver/resolver.go
```diff
@@ -31,8 +31,9 @@ type Resolver struct {
 	rs               *records.RecordGenerator
 	rsLock           sync.RWMutex
 	rng              *rand.Rand
-	fwd              exchanger.Forwarder
 	generatorOptions []records.Option
+	zoneFwds         map[string]exchanger.Forwarder
```
sargun: Mind commenting on what "string" is here?

timcharper: the string key is the tld for the specified forwarder, and is used later, https://github.com/mesosphere/mesos-dns/pull/441/files#diff-c6a2b1bb9d7feeb56f43abf7caadeed1R112

It's not necessary that this data structure be a map, but there's very little disadvantage to it being one IMO.

timcharper: I should clarify and say zone, and not tld, since one is not necessarily the other.

sargun: Yeah, I know what it it is, but I think it's valuable having a comment in the code explaining this.

timcharper: oh, lol; okay, sure.

sargun: See: Things that make me sad about Go.

sargun:
Squash, and I will merge!

timcharper:
Okay, @sargun ! All squashed!

PR #442, opened 2016-06-20
Reuse HTTP client instance

sargun:


sargun:
@kaysoky Mind running a build of this, and seeing if it solves the problem?

records/generator.go
```diff
@@ -226,6 +225,7 @@ func (rg *RecordGenerator) loadFromMaster(ip string, port string) (state.State,
 	}
 
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", "Mesos-DNS")
```
jdef: :+1:

resolver/resolver.go
```diff
@@ -25,27 +25,47 @@ import (
...
+	version    string
+	config     records.Config
+	rs         *records.RecordGenerator
+	rsLock     sync.RWMutex
+	rng        *rand.Rand
+	fwd        exchanger.Forwarder
+	httpClient *http.Client
+}
+
+func newHTTPClient(httpTimeout time.Duration) *http.Client {
```
jdef: probably need to sync w/ @alberts

sargun: Talked to @spahl. He thinks our changes make sense, and he's going to wait till we merge to master (soon), and then he can rebase atop us.

sargun:
@jdef Looks good otherwise? The only thing I was thinking of is adding the PID to the user agent. What do you think?

jdef:
@sargun LGTM; no pid needed (too leaky)

PR #444, opened 2016-06-21
Switch to govendor

sargun:


jdef:
any particular reason that you're not bundling in the vendored code? I saw that with rexray builds were much more stable on TC when all the code deps were vendored and committed as part of the code base vs. resolved/downloaded at build time. Also made for faster builds

sargun:
I find it really ugly. It's hard to track down what's going on when someone changes a dep and maybe 1 or 2 lines. Nonetheless, it wouldn't be difficult to add the code back in. You're right about stable builds, but in exchange for unparseable diffs, I do not know.

sargun:
Nonetheless, we can always add deps. I really wouldn't want to add deps in this PR, just for clarity (and sanity) sake.

jdef:
as long as changes to vendor/ code are grouped into a single commit
separate from other changes then diff's are pretty easy to review

sargun:
How about this:
(1) Merge to the vetshadow changes fixed
(2) Merge to vendor
(3) Merge to add deps to repo

As three separate PRs?

jdef:
sounds good to me

sargun:
@jdef Mergeable?

jdef:
lgtm once it passes CI

PR #446, opened 2016-06-21
Add vendored code itself

sargun:


circle.yml
```diff
@@ -11,24 +11,27 @@ machine:
     - go version
   services:
 
+
 dependencies:
-  pre:
+  cache_directories:
+    - "~/gopath"
```
jdef: might this inadvertently cache the mesos-dns source code?

PR #454, opened 2016-08-03
Update README badge.

ssk2:


sargun:
I still think we should leave Circle-CI there. There was a brief conversation of moving to Jenkinsfiles, and I think that'll help with the build usability. Once we do that, let's reconsider.

PR #462, opened 2016-08-23
Simplify gometalinter ci test

drewkerrigan:
* Fixed several unhandled errors warnings
* Modified gometalinter check in circle.yml to ignore intentional unsafe package usage as well as an SSL warnings.


sargun:
Can we re-enable the tests? And figure out how to solve them rather than just dropping them all? Quite a few of the other linter checks make quite a lot of sense to keep on.

jdef:
agree with @sargun ; `disable-all` is a pretty heavy hammer. let's either fix the problems, or put specific disable patterns in there. for example:
```
gometalinter \
  --vendor --disable=gotype \
  --exclude='cyclomatic complexity .* of function .dnsCharTable.* is high' \
  --concurrency=6 --cyclo-over=12 --tests --deadline=300s ./...
```

drewkerrigan:
Yes I agree that's a better approach, I'll modify the check

drewkerrigan:
CC: @alberts

jdef:
Would be really nice if we could put all the metalinter excludes in one place, vs having a copy in Jenkinsfile and in the circleci file. If we want to tackle that in a separate PR that's fine, just want to see it in the backlog somewhere if we're not tackling it here

jdef:
lgtm once tests pass

sargun:
@drewkerrigan Can you squash your commits and have a single commit with the summary of the changes, and then I'll merge.

sargun:
LGTM.

PR #465, opened 2016-08-25
Rename iam secret to private_key

drewkerrigan:


drewkerrigan:
@jdef @sargun please review when you have a chance.

sargun:
Gotta update: https://github.com/mesosphere/mesos-dns/blob/master/docs/docs/configuration-parameters.md

Other than that, looks okay

drewkerrigan:
@sargun looks like the example in https://github.com/mesosphere/mesos-dns/blob/master/docs/docs/configuration-parameters.md uses `basic` authentication which is still using "principal" and "secret". The only thing that this PR changes is the `iam` authentication mechanism (which now uses "uid", "private_key", and "login_endpoint").

jdef:
lgtm. please loop in the dcos security team on this if not already because this will impact them

PR #467, opened 2016-08-26
Properly proxy truncated responses

### LINKED ISSUE #457: Forwarded truncated DNS responses get extra truncated
With the targeted version of miekg dns client, if the DNS server for which mesos-dns is forwarding requests returns a response with the truncated bit set, then the dns client returns a response with effectively no answers. The truncation is happening here:

https://github.com/mesosphere/mesos-dns/blob/3564f1eb20a5d77e068d38b6fcebe60fbb1b7ef3/vendor/github.com/miekg/dns/msg.go#L1703

This effectively means that if a Forwarded DNS server responds with truncated messages, then mesos-dns returns no answers.

Ideally, mesos-dns should forward on the response as much as possible, indicating that the response was truncated.

timcharper:
Fixes #457


timcharper:
Truncation state is being reset here:

```
dns/resolver/resolver.go

 459 | func truncate(m *dns.Msg, udp bool) *dns.Msg {
 460 | 	max := dns.MinMsgSize
 461 | 	if !udp {
 462 | 		max = dns.MaxMsgSize
 463 | 	} else if opt := m.IsEdns0(); opt != nil {
 464 | 		max = int(opt.UDPSize())
 465 | 	}
 466 | 
 467 | 	m.Truncated = m.Len() > max
 468 | 	if !m.Truncated {
 469 | 		return m
 470 | 	}
 471 | 
 472 | 	m.Extra = nil // Drop all extra records first
 473 | 	if m.Len() < max {
 474 | 		return m
 475 | 	}
 476 | 	answers := m.Answer[:]
 477 | 	left, right := 0, len(m.Answer)
 478 | 	for {
 479 | 		if left == right {
 480 | 			break
 481 | 		}
 482 | 		mid := (left + right) / 2
 483 | 		m.Answer = answers[:mid]
 484 | 		if m.Len() < max {
 485 | 			left = mid + 1
 486 | 			continue
 487 | 		}
 488 | 		right = mid
 489 | 	}
 490 | 	return m
 491 | }
 492 | 
```

timcharper:
I've confirmed, using tcpdump / wireshark, that the last patch causes `mesos-dns` to appropriately forward the truncated state.

timcharper:
Looks like the failure may be unrelated to my change

drewkerrigan:
@timcharper rebasing should help - some changes to circle.yml and other fixes have been committed since you forked

timcharper:
@drewkerrigan thanks; I have rebased

sargun:
Thank you for splitting commits!

sargun:
Do you have an idea on how to simply get truncated replies out of dnsmasq?

timcharper:
@sargun the way I replicate the issue is using weave DNS. I run 14 containers or so with the same weave host name and it serves them truncated

timcharper:
(Weave DNS does not compress responses so it is easier to replicate)

timcharper:
http://unix.stackexchange.com/questions/52933/how-to-assign-multiple-fixed-ip-address-for-one-domain ?

sargun:
@timcharper I'll recreate this issue over the coming week. Would anyone be interested if we had an integration test that propped up a dns server?

timcharper:
🎉

Issue #459, opened 2016-08-22
Support non-standard ports for upstreams


sargun:
We should be able to have a colon in our upstreams and support non-standard ports.

drewkerrigan:
@sargun can this be closed?

PR #468, opened 2016-08-27
Support non-standard ports for upstreams

drewkerrigan:
* Modifies master / resolver validation to reuse more code
* Master / resolver validation now mutate source lists with normalized values for more efficient usage
* Implements #459


drewkerrigan:
@jdef yes your original comment makes more sense now :-).

I assume you'd prefer that approach to having mutation occur with the validation? Cc: @sargun

jdef:
Right, a validate() func shouldn't mutate.

exchanger/forwarder.go
```diff
@@ -23,13 +23,22 @@ func (f Forwarder) Forward(m *dns.Msg, proto string) (*dns.Msg, error) {
 // If no addresses or no matching protocol exchanger exist, a *ForwardError will
 // be returned.
 func NewForwarder(addrs []string, exs map[string]Exchanger) Forwarder {
+	normalized := make([]string, len(addrs))
+	for i, a := range addrs {
```
sargun: What is "A"

exchanger/forwarder.go
```diff
@@ -23,13 +23,22 @@ func (f Forwarder) Forward(m *dns.Msg, proto string) (*dns.Msg, error) {
...
+		normalized[i] = net.JoinHostPort(host, port)
+	}
 	return func(m *dns.Msg, proto string) (r *dns.Msg, err error) {
 		ex, ok := exs[proto]
 		if !ok || len(addrs) == 0 {
 			return nil, &ForwardError{Addrs: addrs, Proto: proto}
 		}
-		for _, a := range addrs {
-			if r, _, err = ex.Exchange(m, net.JoinHostPort(a, "53")); err == nil {
+		for _, a := range normalized {
```
sargun: What is A?

exchanger/forwarder.go
```diff
@@ -23,13 +23,22 @@ func (f Forwarder) Forward(m *dns.Msg, proto string) (*dns.Msg, error) {
 // If no addresses or no matching protocol exchanger exist, a *ForwardError will
 // be returned.
 func NewForwarder(addrs []string, exs map[string]Exchanger) Forwarder {
+	normalized := make([]string, len(addrs))
```
sargun: A comment on what normalized will be populated with would be nice.

records/validation.go
```diff
@@ -10,60 +10,71 @@ func validateEnabledServices(c *Config) error {
 		return fmt.Errorf("Either DNS or HTTP server should be on")
 	}
 	if len(c.Masters) == 0 && c.Zk == "" {
-		return fmt.Errorf("specify mesos masters or zookeeper in config.json")
+		return fmt.Errorf("Specify mesos masters or zookeeper in config.json")
```
sargun: While we're at it, Mesos, and Zookeeper are proper nouns as well.

records/validation.go
```diff
@@ -10,60 +10,71 @@ func validateEnabledServices(c *Config) error {
...
 	}
 	return nil
 }
 
-// validateMasters checks that each master in the list is a properly formatted host:ip pair.
+// validateMasters checks that each master in the list is a properly formatted host|IP:port pair.
```
sargun: Spell out "or"?

records/validation.go
```diff
@@ -10,60 +10,71 @@ func validateEnabledServices(c *Config) error {
...
+			return fmt.Errorf("Duplicate host specified: %v", normalized)
 		}
-		ips[ipstr] = struct{}{}
+		valid[normalized] = struct{}{}
 	}
 	return nil
 }
 
+func normalizeValidateHostPort(hostPort string, ipRequired bool, defaultPort string, portRequired bool) (string, error) {
+	host, port, err := net.SplitHostPort(hostPort)
```
sargun: Do we want to check port:

```
package main

import (
	"fmt"
	"net"
)

func main() {
	_, _, err := net.SplitHostPort("1.2.3.4:-1")
	fmt.Printf("%v\n", err)
}
```
returns:
```
<nil>
```

records/validation.go
```diff
@@ -10,60 +10,71 @@ func validateEnabledServices(c *Config) error {
...
+	if ip == nil {
+		if ipRequired {
+			return "", fmt.Errorf("Illegal ip specified: %v", host)
+		}
+	} else {
+		//TODO(jdef) distinguish between intended hostnames and invalid ip addresses
+		host = ip.String()
+	}
+
+	return host + "_" + port, nil
```
sargun: Why _? Not :? Or https://golang.org/pkg/net/#JoinHostPort

drewkerrigan:
@sargun agree with your comments, they've been addressed

records/validation.go
```diff
@@ -3,67 +3,84 @@ package records
...
 func validateResolvers(rs []string) error {
-	if len(rs) == 0 {
+	if err := validateHostPorts(rs, true, "53", false); err != nil {
+		return fmt.Errorf("Error validating resolvers: %v", err)
+	}
+	return nil
+}
+
+func validateHostPorts(hostPorts []string, ipRequired bool, defaultPort string, portRequired bool) error {
+	if len(hostPorts) == 0 {
```
sargun: Just curious any reason why this short-circuit was here in the first place? I mean, I can't imagine `valid := make(map[string]struct{}, len(hostPorts))` would be so expensive to make this worth it?

CC: @jdef

records/validation.go
```diff
@@ -3,67 +3,84 @@ package records
...
-		if _, found := ips[ipstr]; found {
-			return fmt.Errorf("duplicate resolver IP specified: %v", r)
+		if _, found := valid[normalized]; found {
+			return fmt.Errorf("Duplicate host specified: %v", normalized)
 		}
-		ips[ipstr] = struct{}{}
+		valid[normalized] = struct{}{}
 	}
 	return nil
 }
```
sargun: So, I had to do a couple reads of this function to get what it's going (I still don't really know why it does what it does, but I predict there is a good reason).

Given it's a bit odd, it may make sense to write a godoc explaining it, and the parameters. Alternatively, it might make sense to break it up for each of the required cases? I'm not sure.

sargun: Maybe it makes sense to check the ip required, and portRequired checks first?

records/validation.go
```diff
@@ -3,67 +3,84 @@ package records
...
+		if ipRequired {
+			return "", fmt.Errorf("Illegal ip specified: %v", host)
+		}
+	} else {
+		//TODO(jdef) distinguish between intended hostnames and invalid ip addresses
+		host = ip.String()
+	}
+
+	port, err := strconv.Atoi(portStr)
+	if err != nil || port <= 0 {
```
sargun: Also > 65k

records/validation.go
```diff
@@ -3,67 +3,84 @@ package records
...
+		if _, found := valid[normalized]; found {
+			return fmt.Errorf("Duplicate host specified: %v", normalized)
 		}
-		ips[ipstr] = struct{}{}
+		valid[normalized] = struct{}{}
 	}
 	return nil
 }
 
+func normalizeValidateHostPort(hostPort string, ipRequired bool, defaultPort string, portRequired bool) (string, error) {
```
sargun: I wonder if you could write it where you have a factory function that curries a function for portRequired, and ipRequired?

records/validation.go
jdef: FWIW the length check isn't very important because these validation routines are only called once, at startup time.

sargun: Yeah, it seems a bit weird. Let's get rid of it.

drewkerrigan:
@sargun @jdef I've addressed your above comments, please check out the latest changes.

jdef:
lgtm

drewkerrigan:
@sargun Please take another look when you get a chance - this should make it to master before https://github.com/dcos/dcos/pull/606 gets merged ideally.

exchanger/forwarder.go
```diff
@@ -23,13 +23,23 @@ func (f Forwarder) Forward(m *dns.Msg, proto string) (*dns.Msg, error) {
 // If no addresses or no matching protocol exchanger exist, a *ForwardError will
 // be returned.
 func NewForwarder(addrs []string, exs map[string]Exchanger) Forwarder {
+	// List of IP:port pairs from addrs with or without ports
+	normalized := make([]string, len(addrs))
+	for i, addr := range addrs {
+		host, port, err := net.SplitHostPort(addr)
+		if err != nil {
+			host = addr
+			port = "53"
```
sargun: This feels kinda weird not to be a constant somewhere, but it's okay.

sargun:
One nit, but otherwise, looks good.

exchanger/forwarder.go
drewkerrigan: I agree, but to do that correctly I think we'd want to make a common package or something like that, because the `records/validation.go` would also want to see it. It seems a bit overkill to introduce something like that for this. Thoughts?

sargun: Scratch it.

sargun:
Can you squash?

PR #470, opened 2016-09-06
Update vendored go-zookeeper dependency

vespian:
This includes many bugfixes, including ones related to authentication:
https://github.com/samuel/go-zookeeper/pull/127
https://github.com/samuel/go-zookeeper/pull/139


vespian:
@sargun @drewkerrigan Could you please have a look ? Thanks!

sargun:
LGTM. CC: @jdef

vendor/github.com/samuel/go-zookeeper/zk/dnshostprovider.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"net"
+	"sync"
+)
+
+// DNSHostProvider is the default HostProvider. It currently matches
+// the Java StaticHostProvider, resolving hosts from DNS once during
+// the call to Init.  It could be easily extended to re-query DNS
+// periodically or if there is trouble connecting.
+type DNSHostProvider struct {
```
jdef: this is now the default, and is a significant change from what's in place already. the old code would resolve DNS names for ZK servers upon each connect. now, by default, they're resolved ONCE per invocation of `Connect`. This is a **very** significant change that will impact people who aren't using virtual IPs for their ZK servers.

vespian: Please correct me if I got it wrong:
This change will only affect people who use DNS to dynamically define ZK ensemble members, right ?

jdef: That appears to be the case

vespian: OK, so please decide with @sargun and @drewkerrigan whether we want to support these users or not - I am totally fine with both (== closing/merging the PR). 

Thanks for doing the review !

sargun: I wish we could have the behaviour that if ZK fails, we recreate the client.

jdef: I suppose nothing stops us from writing our own HostProvider implementation
that panic's or fatal's if there's problems. Is that what you're aiming for?

sargun: Well, not only that, but it looks like the Zookeeper client has a little bit of internal state. If we go into a disconnect state, does it make sense to try to completely rebuild the client after some interval to get rid of any potential bad state?

cmaloney: There's a reason why we made Mesos-DNS forcibly die in DC/OS if it can't connect to zk within a relatively short window. Killing / restarting the process always kills bad state. Also makes it so things like load balancers / spartan don't forward traffic to DNS nodes which are serving stale data which doesn't put up any other bad health signs. I'm a big fan of keeping the behavior simple / just kill Mesos-DNS and let it be restarted if things don't heal fairly quickly.

sargun: The only problem with the fail-stop model is that it prevents the cluster's current state from being served if only the ZKs fail. I don't know what we should prefer.

cmaloney: DNS clients tend to be dumb, we have no idea stale data is being served, there is no alerting / monitoring of any of this. It's a silent failure that shows up in bad behavior. Failing hard at least makes a nice, recognizable impact of where to look, and shows up in the monitoring we have today.

PR #476, opened 2016-09-29
godep: Move from govendor to godep

nlsun:
The decision to move from godep earlier was because it did not have
working `vendor/`support at the time.

Remotely related to https://github.com/dcos/dcos/pull/752


nlsun:
Ready for review

sargun:
Can you please update the circle.yml as well and build instructions?

nlsun:
Updated circle.yml. I don't think build instructions have changed though.

PR #480, opened 2016-10-12
docs: Document the `enumerate` endpoint

nlsun:


docs/docs/http.md
```diff
@@ -72,3 +73,64 @@ curl http://10.190.238.173:8123/v1/services/_nginx._tcp.marathon.mesos.
...
+## `GET /v1/enumerate`
+
+Lists in JSON format all DNS information.
+
+```console
+curl http://master.mesos:8123/v1/enumerate
+{
+  "frameworks": [
+   {
+    "tasks": null,
```
jdef: too bad we don't omit this when empty

nlsun: Should I change this? I just copied this out of the DC/OS docs.

jdef: probably fine to have copied. omitting the `null` fields is something that should be configured at in the Golang source.

sargun: I thought we fixed this and it returns []?

nlsun: I went ahead and ran the command on a live cluster and updated it accordingly.

jdef: :+1:

jdef:
lgtm

Issue #286, opened 2015-09-26
Proposal: Support PTR records / Generate off of A


sepiroth887:
Reasoning:

Various ssl implementations (notably java) have issues when trying to connect to ssl enabled servers when no PTR zones can be found. 

This can happen regardless whether an actual A record reverse pointer exists or not. Its mostly the zone that matters.

Having PTR records generally can speed up ssl negotiation for other implementations that try a reverse lookup. 

Also iirc Hbase/hdfs requires ptr entries as well. 

Solution:

Add a PTR map to store reverse PTR records and generate them for every A record from masters. 
maybe the dns library already supports this. 

Also generate zones on say the /18 of every Ipv4 to allow gracefull failure of ptr lookups. 

Thoughts?

sepiroth887:
Actually just read the related and closed issue. Having multiple entries for single dns entry is a bad bad idea. 

However id love to see the zone implemented such that stupid implementations dont break. Its not nice i know but there are plenty out there that should be supported regardless imo

tsenart:
Reference: https://github.com/mesosphere/mesos-dns/issues/67

tsenart:
@sepiroth887: We should keep in mind that Mesos-DNS is primarily *not* a DNS zone server.

sepiroth887:
I am aware of that. However if the intend of mesos-dns is to act as a primary dns and a service discovery mechanism for applications then i do think it should support that feature. 

If i do have to keep a dynamically configuring dns server to fullfill that requirement then there is really no reason to use mesos-dns in such a scenario as the dynamic piece would already have to talk to mesos master/zk to work. :( 

I am hoping that eventually mesosdns will become the only required app for dns resolution inside a mesos cluster.

jdef:
I'm not opposed at all to PTR records for all of the A's that we generate.

sepiroth887:
i think the primary reason PTRs themselves could be bad is it may confuse apps looking for certain entries. For ssl it could break the validation if the response FQDN is not part of the cert ( and there is basically a n-1/n chance of that happening if you run n tasks on the same IP)

Not sure how this is solved today with other DNS servers though, I assume CNames?

jdef:
sounds like PTR's for A records created w/ container IP's (vs. host IPs) might still be useful. it'll take frameworks a bit to catch up with ip-per-container (and not all mesos deployments will want to support it). could be a nice way to test rollout: enabled by default for containers that have their own IPs.

[...]

Issue #67, opened 2015-01-23
Support Reverse Lookup


kensipe:
mesos-dns does not support reverse lookup.   The following command should return the service name that is registered at that node:

$ dig @$servicehost -x 10.240.214.151

** where $servicehost is the IP of the mesos-dns service.

tarnfeld:
Presumably you would need to return multiple records to describe the many services that might be running on this node? I'm not sure if it's valid for a reverse lookup to return multiple `PTR` record types.

I believe the RFC doesn't explicitly prohibit the use of multiple PTR records for reverse lookups, but there's plenty of [material](https://supernoc.rogerstelecom.net/pdfs/multiple-ptrs.pdf) on the net about why you shouldn't do it.

kozyraki:
I don't particularly like the idea of supporting reverse lookup (through PTR records). In addition to the issues Tom lists above, it is a security concern and an anti-pattern as allows one to query about all tasks running on a specific machine. Other than the Mesos master, that already has this info, I am not sure who else needs it.

tarnfeld:
IMO if you want to embed this information into the DNS system it's worth reading more into the utilities the DNS-SD RFC (6763) spec provides.

mindscratch:
With mesos-dns in place, I'm not able to use cURL and I suspect http libraries would also have issues if I tried to access "http://chronos.marathon.mesos". 

Programmatically, if want to perform a GET request to chronos.marathon.mesos, would I first need to do a DNS query to get the IP address, and then perform the GET?

mindscratch:
@kozyraki my use case is mostly having services that use static configuration files. It would be nice to put in the configuration file "<service>.marathon.mesos". If marathon moves that service to a different node, no configuration changes are required. However, many client applications are not "home grown" so I can't alter how they look up that host. Most likely, those applications expect DNS to resolve properly versus doing a DNS query and then interacting with the service. Does that help?

mindscratch:
@kozyraki with the opening of [this ticket on marathon](https://github.com/mesosphere/marathon/issues/1112), does that mean reverse lookup support will be added to mesos-dns eventually?

kozyraki:
@mindscratch 
The problem with curl is either is a misconfiguration or a bug from our side. You don't need inverse DNS to translate chronos.marathon.mesos to an IP address. Here are some suggestions of things to check:
[...]

PR #481, opened 2016-10-13
Serve PTR records for unambiguous tasks

christian-marie:
Hi, we (a team at Commonwealth Bank) are trying to roll out a HDFS cluster on top of Mesos. This seems to require reverse DNS to be properly configured for tasks, and this seemed the most convenient place to add PTR records.

This patch is currently only unit tested, I'll be integration testing tomorrow.

This may be of interest to #286 and #67 (closed).

### DIFF
--- a/records/generator.go
+++ b/records/generator.go
@@ -22,6 +22,7 @@ import (
 	"github.com/mesosphere/mesos-dns/records/labels"
 	"github.com/mesosphere/mesos-dns/records/state"
 	"github.com/mesosphere/mesos-dns/urls"
+	"github.com/miekg/dns"
 	"github.com/tv42/zbase32"
 )
 
@@ -75,6 +76,8 @@ type rrsKind string
 const (
 	// A record types
 	A rrsKind = "A"
+	// PTR record types
+	PTR = "PTR"
 	// SRV record types
 	SRV = "SRV"
 )
@@ -83,6 +86,8 @@ func (kind rrsKind) rrs(rg *RecordGenerator) rrs {
 	switch kind {
 	case A:
 		return rg.As
+	case PTR:
+		return rg.PTRs
 	case SRV:
 		return rg.SRVs
 	default:
@@ -95,6 +100,7 @@ func (kind rrsKind) rrs(rg *RecordGenerator) rrs {
 type RecordGenerator struct {
 	As            rrs
 	SRVs          rrs
+	PTRs          rrs
 	SlaveIPs      map[string]string
 	EnumData      EnumerationData
 	httpClient    httpcli.Doer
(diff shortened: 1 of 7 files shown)


jdef:
ok to test

christian-marie:
Update: Handlers needed to be set up for these reverse records (obviously).

Tested to work in practice now (very ad-hoc testing, redacted):

```
$ dig aaa.bb.cc.dd.in-addr.arpa. -tptr @localhost

; <<>> DiG 9.10.3-P4-Ubuntu <<>> aaa.bb.cc.dd.in-addr.arpa. -tptr @localhost
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 29836
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;aaa.bb.cc.dd.in-addr.arpa.	IN	PTR

;; ANSWER SECTION:
aaa.bb.cc.dd.in-addr.arpa. 15	IN	PTR	task-name.domain.

;; Query time: 0 msec
;; SERVER: ::1#53(::1)
;; WHEN: Mon Oct 17 05:08:54 UTC 2016
;; MSG SIZE  rcvd: 93
```

christian-marie:
@jdef I'm pretty happy with this now, does it seem workable?

jdef:
last time @sargun that I discussed PTR records there was some concern about mesos-dns being responsible for ALL PTR-resolution and that it would break in enterprise environments. We really didn't have a strong plan for delegation of PTR resolution. There's at least one RFC that tries to tackle this. And a quick google search turns up existing documentation for other DNS apps w/ respect to PTR delegation, e.g. http://www.zytrax.com/books/dns/ch9/reverse.html . I suspect that a more robust solution would be needed long term if we intent to fully support PTR resolution in reasonably sized clusters within an enterprise context.

I have no problem with mesos-dns supporting PTR resolver functionality, as implemented, as long as it defaults to "off" (seems to be the case for this PR) and comes with a strong warning in the docs that this is a highly experimental feature and enabling it could break anything but the most simple mesos clusters.

christian-marie:
Ok. I'll add some experimental flags and docs today.

christian-marie:
Rebased, WIP removed.

christian-marie:
P.S. Some docs added.

sargun:
Is there a SOA for the PTR zone?

christian-marie:
Not sure what you mean. If you can modify the test to demonstrate, that would hep.

christian-marie:
Good day sirs

records/generator.go
```diff
@@ -554,6 +561,10 @@ func (rg *RecordGenerator) taskContextRecord(ctx context, task state.Task, f sta
 	rg.insertTaskRR(arec+tail, ctx.taskIP, A, enumTask)
 	rg.insertTaskRR(canonical+tail, ctx.taskIP, A, enumTask)
 
+	// insert PTR records, these are only served when ReverseDnsOn is true,
+	// which sets up a handler for the reveverse DNS zones
+	rg.insertTaskRR(arec+tail, ctx.taskIP, PTR, enumTask)
```
drewkerrigan: @christian-marie Perhaps we should only insert this type when configured to do so? Wouldn't want unusable data taking up space.

christian-marie: Don't really care about performance or this patch anymore.

jdef:
closing this out since the OP is no longer interested in this and we haven't needed it yet for DC/OS. please re-open if this is being closed in error.

PR #483, opened 2016-11-05
Added client side certificate and key pair.

tillt:
- adds the ability to provide client side SSL certificate/key-pair for the TLS configuration
- adds configuration parsing for `CertFile`
- adds configuration parsing for `KeyFile`

### DIFF
--- a/records/config.go
+++ b/records/config.go
@@ -1,6 +1,7 @@
 package records
 
 import (
+	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


records/config.go
```diff
@@ -165,6 +173,14 @@ func SetConfig(cjson string) Config {
 		c.caPool = pool
 	}
 
+	if c.CertFile != "" {
```
jdef: it's probably an error if one of these new config params is set, but not the other. any value in having a smarter check here for sanity?

tillt: Jeez, this ```records/config.go:134::warning: cyclomatic complexity 16 of function SetConfig() is high (> 12) (gocyclo)``` is getting me crazy - guess I should leave this stuff to the pro's.

jdef:
lgtm. PTAL @sargun

sargun:
LGTM. I'll merge once you rebase.

tillt:
@sargun is there more I need to do?

PR #485, opened 2016-11-16
Fix truncation errors

sargun:


sargun:
@jdef Mind reviewing? I'm curious as to whether I need the sampling at all, and I can't just take the first bit of the slice, because of the shuffle.

PR #488, opened 2017-01-24
resolver: make setting of the Truncate bit configurable

gpaul:
Setting the Truncate bit in the DNS message when it is truncated is now configurable.

The default behaviour is to set the Truncate bit when appropriate.

When this option is set to false the Truncate bit will be unset in all DNS responses.

This is non-compliant behaviour but we support it for cases where clients are also non-compliant.

### DIFF
--- a/records/config.go
+++ b/records/config.go
@@ -77,6 +77,11 @@ type Config struct {
 	// SetTruncateBit when `false` ensures responses never have the Truncate bit set even
 	// if they were truncated. When `true` any message that gets truncated will have the
 	// Truncate bit set.
+	// Compliant clients that receive truncated responses will retry the request
+	// over TCP. Some non-compliant clients simply cannot handle truncated responses at all.
+	// Disabling this option causes mesos-dns to behave in a non-compliant way. It exists
+	// only as a workaround for non-compliant clients and users who cannot tolerate the
+	// latency added by the standard TCP fallback.
 	SetTruncateBit bool
 	// Enumeration enabled via the API enumeration endpoint
 	EnumerationOn bool
--- a/records/config.go
+++ b/records/config.go
@@ -74,6 +74,10 @@ type Config struct {
 	ExternalOn bool
 	// EnforceRFC952 will enforce an older, more strict set of rules for DNS labels
 	EnforceRFC952 bool
+	// SetTruncateBit when `false` ensures responses never have the Truncate bit set even
+	// if they were truncated. When `true` any message that gets truncated will have the
+	// Truncate bit set.
+	SetTruncateBit bool
 	// Enumeration enabled via the API enumeration endpoint
 	EnumerationOn bool
 	// Communicate with Mesos using HTTPS if set to true
@@ -123,6 +127,7 @@ func NewConfig() Config {
 		DNSOn:               true,
 		HTTPOn:              true,
 		ExternalOn:          true,
+		SetTruncateBit:      true,
 		RecurseOn:           true,
 		IPSources:           []string{"netinfo", "mesos", "host"},
 		EnumerationOn:       true,
@@ -272,6 +277,7 @@ func (c Config) log() {
 	logging.Verbose.Println("   - HttpOn: ", c.HTTPOn)
 	logging.Verbose.Println("   - ConfigFile: ", c.File)
 	logging.Verbose.Println("   - EnforceRFC952: ", c.EnforceRFC952)
+	logging.Verbose.Println("   - SetTruncateBit: ", c.SetTruncateBit)
 	logging.Verbose.Println("   - IPSources: ", c.IPSources)
 	logging.Verbose.Println("   - EnumerationOn", c.EnumerationOn)
 	logging.Verbose.Println("   - MesosHTTPSOn", c.MesosHTTPSOn)
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -7,13 +7,14 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
-	"github.com/emicklei/go-restful"
+	restful "github.com/emicklei/go-restful"
 	_ "github.com/mesos/mesos-go/detector/zoo" // Registers the ZK detector
 	"github.com/mesosphere/mesos-dns/exchanger"
 	"github.com/mesosphere/mesos-dns/logging"
@@ -285,7 +286,7 @@ func (res *Resolver) HandleNonMesos(fwd exchanger.Forwarder) func(
 		} else if len(m.Answer) == 0 {
 			logging.CurLog.NonMesosNXDomain.Inc()
 		}
-		reply(w, m)
+		reply(w, m, res.config.SetTruncateBit)
 	}
 }
 
@@ -343,7 +344,7 @@ func (res *Resolver) HandleMesos(w dns.ResponseWriter, r *dns.Msg) {
 		logging.CurLog.MesosFailed.Inc()
 	}
 
-	reply(w, m)
+	reply(w, m, res.config.SetTruncateBit)
 }
 
 func (res *Resolver) handleSRV(rs *records.RecordGenerator, name string, m, r *dns.Msg) error {
@@ -441,10 +442,9 @@ func (res *Resolver) handleEmpty(rs *records.RecordGenerator, name string, m, r 
 
 // reply writes the given dns.Msg out to the given dns.ResponseWriter,
 // compressing the message first and truncating it accordingly.
-func reply(w dns.ResponseWriter, m *dns.Msg) {
+func reply(w dns.ResponseWriter, m *dns.Msg, setTruncateBit bool) {
 	m.Compress = true // https://github.com/mesosphere/mesos-dns/issues/{170,173,174}
-
-	if err := w.WriteMsg(truncate(m, isUDP(w))); err != nil {
+	if err := w.WriteMsg(truncate(m, isUDP(w), setTruncateBit)); err != nil {
 		logging.Error.Println(err)
 	}
 }
@@ -457,7 +457,12 @@ func isUDP(w dns.ResponseWriter) bool {
 // truncate removes answers until the given dns.Msg fits the permitted
 // length of the given transmission channel and sets the TC bit.
 // See https://tools.ietf.org/html/rfc1035#section-4.2.1
-func truncate(m *dns.Msg, udp bool) *dns.Msg {
+// If `setTruncateBit` is true, the Truncate bit will be set if the message
+// has been truncated already or gets truncated here.
+// If `setTruncateBit` is false, the message will have its Truncate bit
+// cleared if it is already set, and won't set it even if it does get
+// truncated.
+func truncate(m *dns.Msg, udp bool, setTruncateBit bool) *dns.Msg {
 	max := dns.MinMsgSize
 	if !udp {
 		max = dns.MaxMsgSize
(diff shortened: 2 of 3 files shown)


resolver/resolver.go
```diff
@@ -467,28 +472,29 @@ func truncate(m *dns.Msg, udp bool) *dns.Msg {
...
-			continue
-		}
-		right = mid
+	search := func(n int) bool {
+		m.Answer = answers[:n]
+		return m.Len() > max
+	}
+	limit := sort.Search(len(answers), search)
+	if limit > 0 {
+		m.Answer = answers[:limit-1]
```
jdef: pretty sure that this should be `answers[:limit]` to cause the item at index `limit` to be excluded and the item at `limit-1` to be included (which is what we really want, right?)

records/config.go
```diff
@@ -74,6 +74,10 @@ type Config struct {
 	ExternalOn bool
 	// EnforceRFC952 will enforce an older, more strict set of rules for DNS labels
 	EnforceRFC952 bool
+	// SetTruncateBit when `false` ensures responses never have the Truncate bit set even
+	// if they were truncated. When `true` any message that gets truncated will have the
+	// Truncate bit set.
```
jdef: Perhaps a comment here about setting the value to `false` results in non-compliance, but is required for some clients that are non-compliant?

resolver/resolver.go
```diff
@@ -467,28 +472,29 @@ func truncate(m *dns.Msg, udp bool) *dns.Msg {
...
+	} else {
+		m.Answer = answers[:0]
```
jdef: any particular reason that we do this instead of setting the value to `nil`?

jdef:
honestly, this feels a little hackish. can someone provide a real-world use case that justifies this? is there precedence for a configuration setting like this in other DNS server implementations?

resolver/resolver.go
gpaul: No - `sort.Search` returns the first value of `n` for which the bool check `m.Len() > max` is true. Looking at the search function it is clear that this is the value `n` such that `m.Answer[:n]` results in `m.Len() > max`.

Perhaps a better approach would be to search in the reverse direction, then we'd return the first `n` for which `m.Len() <= max`. I'll see if I can make this code simpler, but it certainly is correct as is.

resolver/resolver.go
gpaul: This doesn't blow away the array that backs `m.Answer`, which means `append(m.Answer , ...)` on the message returned from this function won't allocate a new backing array. 

This is a standard go idiom in high-performance environments. My impression is that mesos-dns is such an environment. Looking at this function in the greater context of the project this doesn't seem necessary as we never append answers to the message returned from truncate. Still, that is a global consideration that may change in the future so I'd rather write fast code here and not worry.

records/config.go
gpaul: Good idea - will do.

gpaul:
@jdef Yes - it is hackish, non-compliant and sad. 

>  can someone provide a real-world use case that justifies this?

The Truncate bit triggers compliant clients to resend the DNS request over TCP. This means there lurks a performance cliff that you drop off once you have enough records. This hits mesos-dns users when they start enough instances of some task.

This PR allows users with large clusters to avoid this cliff by forcing mesos-dns to clear the Truncate bit. This doesn't play nice with for example a round-robin DNS cache as it will only balance load over a subset of the available hosts. This and the general non-compliance is why we don't make this the default.

That's the performance argument. The non-compliance argument can be seen in libraries that fail to handle truncated DNS responses, eg. http://lists.uclibc.org/pipermail/uclibc-cvs/2007-August/024363.html

PM me for details of real-world cases where this caused problems.

> is there precedence for a configuration setting like this in other DNS server implementations?

A quick google shows https://github.com/hashicorp/consul/issues/375. 
Nothing new under the sun.

gpaul:
@jdef PTAL there are two new commits. We can squash them all together once everyone is happy.

alberts:
@mesosphere-ci Retest this please

resolver/resolver.go
```diff
@@ -467,29 +472,28 @@ func truncate(m *dns.Msg, udp bool) *dns.Msg {
...
-			break
-		}
-		mid := (left + right) / 2
-		m.Answer = answers[:mid]
-		if m.Len() < max {
-			left = mid + 1
-			continue
-		}
-		right = mid
+	answers := m.Answer
```
nlsun: `answers := m.Answer[:]` to take a copy.

resolver/resolver.go
```diff
@@ -467,29 +472,28 @@ func truncate(m *dns.Msg, udp bool) *dns.Msg {
...
+	search := func(n int) bool {
```
nlsun: I think you should pull this out into a separate function and unit test it. Especially for off by 1 errors.

resolver/resolver_test.go
```diff
@@ -417,27 +417,68 @@ func TestMultiError(t *testing.T) {
...
+func TestTruncateNoSetTruncateBit(t *testing.T) {
+	testTruncateNoSetTruncateBit(t, dns.MinMsgSize)
+}
+
+func TestTruncateEdns0NoSetTruncateBit(t *testing.T) {
+	testTruncateNoSetTruncateBit(t, 4096)
+}
+
+func testTruncateSetTruncateBit(t *testing.T, maxSize uint16) {
+	orig := *newMessage(maxSize)
```
nlsun: I don't think there's any point in this `orig` `tmp` thing. They aren't treated separately later on anyways.

resolver/resolver.go
```diff
@@ -457,7 +457,12 @@ func isUDP(w dns.ResponseWriter) bool {
 // truncate removes answers until the given dns.Msg fits the permitted
 // length of the given transmission channel and sets the TC bit.
 // See https://tools.ietf.org/html/rfc1035#section-4.2.1
-func truncate(m *dns.Msg, udp bool) *dns.Msg {
+// If `setTruncateBit` is true, the Truncate bit will be set if the message
+// has been truncated already or gets truncated here.
+// If `setTruncateBit` is false, the message will have its Truncate bit
+// cleared if it is already set, and won't set it even if it does get
+// truncated.
+func truncate(m *dns.Msg, udp bool, setTruncateBit bool) *dns.Msg {
```
nlsun: I know the original function had the same signature, but I feel like we shouldn't return the `*dns.Msg`. It's kinda misleading, as we mutate the passed in `m *dns.Msg` and then return the exact same pointer to it. Passing something in and getting returned another implies that the original is left untouched and a completely new thing is returned, which is not the case here.

This is also causing what I believe to be a bug in the unit tests, I have a suspicion that all the times when the input and output of `truncate` is compared, it will always trivially be true because they are the same. Instead, what should be done is to _start_ with 2 separate messages, created the exact same way, mutating one, and _then_ comparing them.

resolver/resolver.go
```diff
@@ -467,29 +472,28 @@ func truncate(m *dns.Msg, udp bool) *dns.Msg {
...
+		// to have it set set or whether we are going to truncate it here.
+		m.Truncated = false
+	}
 	if !furtherTruncation {
 		return m
 	}
-
 	m.Extra = nil // Drop all extra records first
 	if m.Len() < max {
+		// Dropping the extra records shrunk the message down sufficiently.
```
gpaul: I don't believe so - shrunk is the past participle of shrink (thanks google.) I shrink, yesterday I shrunk. I'll reword the comment anyway.

resolver/resolver.go
gpaul: Good idea

resolver/resolver.go
gpaul: Not so - `m.Answer[:]` when `m.Answer` is a slice has no effect. It certainly doesn't make a copy.
See https://play.golang.org/p/3ZhPxkovRr

resolver/resolver_test.go
gpaul: The point of `orig` and `tmp` is exactly to handle your previous concern regarding mutating the input. Notice that we dereference (`*truncate()`) where we use it.

For instance:

```
orig := *newMessage(maxSize)
// orig is a concrete value
tmp := orig
// tmp is also a concrete value, it is not orig, although tmp.Answer shares a
// backing array with orig.Answer which could cause a problem if this array
// got sorted later on, for instance.
msg := *truncate(&tmp, true, true)
// tmp is mutated, but never used again (hence it is called 'tmp')
// msg is a new concrete value.
// orig is left unchanged.
...
```

resolver/resolver.go
gpaul: Regarding the pointer - this is "c-style" code intended to be fast. It does not allocate new objects. If we were to return a new object instead of mutating the old one this function (and consequently mesos-dns as a whole) would be slower.

The reason we return the pointer to the input is to improve readability at the site where this call is made:

```
if err := w.WriteMsg(truncate(m, isUDP(w), setTruncateBit)); err != nil {
...
```
vs.
```
m = truncate(m, isUDP(w), setTruncateBit)
if err := w.WriteMsg(m); err != nil {
...
```

I'm easy either way - but this is the original author's reasoning behind this signature.

Regarding mutation and bugs in tests, see my comments lower down.

resolver/resolver_test.go
gpaul: As you read this code - keep in mind where you have values and where you have pointers.

If `a` is a value, then `b := a` is a new value. Obviously for pointers (and reference types like slices and maps) `b` would now be pointing at the same value.

resolver/resolver_test.go
```diff
@@ -417,27 +417,68 @@ func TestMultiError(t *testing.T) {
...
+	// truncate. This let's us test "truncate() of a large truncated message".
+	if opt := orig.IsEdns0(); opt != nil {
+		msg.SetEdns0(opt.UDPSize(), opt.Do())
+	}
+	msg2 := *truncate(&msg, true, true)
+	if !msg2.Truncated {
 		t.Fatal("Original truncation status was not preseved")
 	}
-	if tm2.Len() != tm.Len() {
+	if msg2.Len() != msg.Len() {
```
gpaul: All of the above notwithstanding - you're *correct* that this equality is a bug! :+1:

resolver/resolver_test.go
gpaul: And sorry if none of this is news to you :)

resolver/resolver.go
jdef: Thanks for pushing back on this. Reading the docs too quickly it seems. I first (wrongly) interpreted `limit` to be an index, not "some number of messages".

resolver/resolver.go
nlsun: I see, then you might want to use `copy`, I think this code's correctness depends on using a copy?

gpaul:
@nlsun I'm preparing changes that should address all your concerns. I've realized that having to explain the code in such detail obviously means the code isn't straight-forward, which means it isn't maintainable.

resolver/resolver.go
nlsun: Another way to test this, is to implement the naive version, and then iterate from 0 answers until the naive hits `(max * 5)`, so that binary search takes several tries to succeed.

resolver/resolver.go
nlsun: ah, i see, scratch that, `copy` is not needed.

resolver/resolver_test.go
nlsun: I don't think you should depend on dereferencing the struct to be a deep copy, instead you should just use `reflect.Deepcopy`. For example, if inside the struct, `dns.Msg` held another pointer, then the copy is invalid.

resolver/resolver_test.go
nlsun: Yeah, just use `reflect.DeepEquals` here

resolver/resolver.go
nlsun: I believe c-style code deserve the full treatment, including c-style function signatures. The author might really like one-liners, but I don't think it's worth sacrificing clarity. It's your call though.

resolver/resolver.go
```diff
@@ -467,29 +472,28 @@ func truncate(m *dns.Msg, udp bool) *dns.Msg {
 
 	furtherTruncation := m.Len() > max
 	m.Truncated = m.Truncated || furtherTruncation
-
+	if !setTruncateBit {
```
nlsun: This chunk of `m.Truncated` setting also struck me as kind of strange looking, but I couldn't think of a better way to express it off the top of my head.

resolver/resolver_test.go
gpaul: I considered that - would you mind having a look at my latest changes? They are targeted at addressing your original concerns.

resolver/resolver.go
gpaul: Agreed - my latest changes reflect the same reasoning.

resolver/resolver.go
gpaul: :+1:

gpaul:
This PR has been through the github washing machine - would you mind having a look at its replacement https://github.com/mesosphere/mesos-dns/pull/492 instead?

resolver/resolver_test.go
gpaul: Yep - I think my latest changes fix address this - there's no more address-of / dereferencing magic.

resolver/resolver.go
gpaul: Agreed - have a look at `truncateAnswers` and the accompanying tests in the latest commit: https://github.com/mesosphere/mesos-dns/pull/492/commits/be8772b23bd6748b5dccee965d21fc150ebda74e

PR #490, opened 2017-01-27
Reload before serving to avoid lookup failures on start.

alberts:
Avoids problems we've seen where Mesos-DNS will serve up NXDOMAINs shortly after a Mesos/Marathon/ZooKeeper restart.


alberts:
Still needs some fixing...

alberts:
Superceded by https://github.com/mesosphere/mesos-dns/pull/491

PR #492, opened 2017-01-31
resolver: make setting of the Truncate bit configurable

alberts:
@gpaul: merged https://github.com/mesosphere/mesos-dns/pull/488 a bit quickly. Please take a look at @nlsun's concerns there.

### DIFF
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -444,9 +444,7 @@ func (res *Resolver) handleEmpty(rs *records.RecordGenerator, name string, m, r 
 // compressing the message first and truncating it accordingly.
 func reply(w dns.ResponseWriter, m *dns.Msg, setTruncateBit bool) {
 	m.Compress = true // https://github.com/mesosphere/mesos-dns/issues/{170,173,174}
-	// Get the maximum allowed message size.
 	maxsize := maxMsgSize(isUDP(w), m.IsEdns0())
-	// Truncate the message by discarding records until it fits.
 	truncate(m, maxsize, setTruncateBit)
 	if err := w.WriteMsg(m); err != nil {
 		logging.Error.Println(err)
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -7,13 +7,14 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
-	"github.com/emicklei/go-restful"
+	restful "github.com/emicklei/go-restful"
 	_ "github.com/mesos/mesos-go/detector/zoo" // Registers the ZK detector
 	"github.com/mesosphere/mesos-dns/exchanger"
 	"github.com/mesosphere/mesos-dns/logging"
@@ -285,7 +286,7 @@ func (res *Resolver) HandleNonMesos(fwd exchanger.Forwarder) func(
 		} else if len(m.Answer) == 0 {
 			logging.CurLog.NonMesosNXDomain.Inc()
 		}
-		reply(w, m)
+		reply(w, m, res.config.SetTruncateBit)
 	}
 }
 
@@ -343,7 +344,7 @@ func (res *Resolver) HandleMesos(w dns.ResponseWriter, r *dns.Msg) {
 		logging.CurLog.MesosFailed.Inc()
 	}
 
-	reply(w, m)
+	reply(w, m, res.config.SetTruncateBit)
 }
 
 func (res *Resolver) handleSRV(rs *records.RecordGenerator, name string, m, r *dns.Msg) error {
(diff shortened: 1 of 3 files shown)


nlsun:
@mesosphere-ci Retest this please.

resolver/resolver.go
```diff
@@ -444,7 +444,11 @@ func (res *Resolver) handleEmpty(rs *records.RecordGenerator, name string, m, r
 // compressing the message first and truncating it accordingly.
 func reply(w dns.ResponseWriter, m *dns.Msg, setTruncateBit bool) {
 	m.Compress = true // https://github.com/mesosphere/mesos-dns/issues/{170,173,174}
-	if err := w.WriteMsg(truncate(m, isUDP(w), setTruncateBit)); err != nil {
+	// Get the maximum allowed message size.
```
nlsun: I don't think this comment the next are necessary, the comments that live with the function are more than enough.

resolver/resolver_test.go
```diff
@@ -433,52 +433,145 @@ func TestTruncateEdns0NoSetTruncateBit(t *testing.T) {
...
 	}
+	if msg.Truncated {
+		t.Fatal("Truncate bit set even though setTruncateBit was false")
+	}
+	// We set the Truncate bit on the message and confirm that the bit is
+	// cleared even though the message did not need to be truncated.
+	// This asserts that no matter the message or its size, truncate() will
+	// clear the Truncate bit if setTruncateBit=false is passed to truncate().
+	before := msg.Len()
+	msg.Truncated = true
```
nlsun: Since the comment above is meant for this line i think they should be touching.

resolver/resolver_test.go
```diff
@@ -433,52 +433,145 @@ func TestTruncateEdns0NoSetTruncateBit(t *testing.T) {
...
+	truncate(msg, max, false)
+	if msg.Truncated {
+		t.Fatal("Truncate bit not cleared")
+	}
+	if msg.Len() != before {
+		t.Fatal("message truncated further")
+	}
+	msg.Answer = append(msg.Answer, genA(1)...)
+	if l := msg.Len(); l < int(max) {
+		t.Fatalf("Message to small after adding answers: %d bytes", l)
```
nlsun: too

resolver/resolver_test.go
```diff
@@ -433,52 +433,145 @@ func TestTruncateEdns0NoSetTruncateBit(t *testing.T) {
...
+	// was large, however in this case since the maximum size is so small
+	// we test the space exhaustively.
+	max := uint16(4096)
+	for nn := 0; ; nn++ {
+		msg := Message(
+			Question("example.com.", dns.TypeA),
+			Header(false, dns.RcodeSuccess),
+			Answers(genA(nn)...))
+		before := msg.Len()
+		if before > int(max)*2 {
```
nlsun: I think we should go up to `int(max)*7)`, reason being, since we're using binary search, we're not really testing the binary search if we end at an even number. This is probably tested enough at `int(max)*7)`, but why not just stick in a few more iterations to be safe :).

resolver/resolver_test.go
```diff
@@ -433,52 +433,145 @@ func TestTruncateEdns0NoSetTruncateBit(t *testing.T) {
...
+		}
+	}
+}
+
+func TestMaxMsgSizeTCP(t *testing.T) {
+	msg := newTruncated()
+	msg.SetEdns0(4096, false)
+	edns0 := msg.IsEdns0()
+	if edns0 == nil {
+		panic("SetEdns0 did not set the EDNS0 (OPT) record")
```
nlsun: why panic instead of t.Fatal?

resolver/resolver_test.go
```diff
@@ -433,52 +433,145 @@ func TestTruncateEdns0NoSetTruncateBit(t *testing.T) {
...
+		t.Fatalf("Wrong max message size: %d", v)
+	}
+}
+
+func TestMaxMsgSizeUDP(t *testing.T) {
+	msg := newTruncated()
+	msg.SetEdns0(4096, false)
+	edns0 := msg.IsEdns0()
+	if edns0 == nil {
+		panic("SetEdns0 did not set the EDNS0 (OPT) record")
```
nlsun: same as above re: panic

resolver/resolver_test.go
gpaul: Ha! Love this comment!

resolver/resolver_test.go
gpaul: Yes - I spent some time deciding so here are my thoughts.

If `edns0` is nil, it means the `dns` library is broken. It does not mean the behaviour that is being tested has failed. It is closer to a runtime error than a failed assert.

resolver/resolver_test.go
nlsun: Actually a typo, i mean't `int(max)*3)` for that second one up there ;)

nlsun:
@mesosphere-ci Retest this please.

PR #493, opened 2017-02-01
Bump github.com/miekg/dns to 99f84ae56e75126dd77e5de4fae2ea034a468ca1

gpaul:
Bumped to the following commit to fix non-compliant compression of various fields. 

commit 99f84ae56e75126dd77e5de4fae2ea034a468ca1
Author: Matthijs Mekking <github@pletterpet.nl>
Date:   Mon Jan 30 11:53:01 2017 +0100

    return errTruncated after setting header bits (#442)

### DIFF
--- a/vendor/github.com/miekg/dns/.travis.yml
+++ b/vendor/github.com/miekg/dns/.travis.yml
@@ -1,7 +1,13 @@
 language: go
 sudo: false
 go:
-  - 1.5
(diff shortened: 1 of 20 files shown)


nlsun:
Looks like just updating a library, so it's probably fine.

alberts:
@gpaul please do a DC/OS bump too

PR #496, opened 2017-02-25
Upgrade to latest mesos-go

jdef:
* minor ZK detector interface changes (should have no impact on mesos-dns)
* mesos-go v0 API has been repackaged, update existing references to new package naming
* ignore `.bak` files


jdef:
mesos-dns was migrated to `dep`. If we want to update the mesos-go deps, we should do it in another PR.

PR #497, opened 2017-03-13
resolver: specify valid Accept content types

gpaul:
This PR fixes an issue where specifying `Accept: application/json` incorrectly resulted in a `406 Not Acceptable` response.

Fixes https://jira.mesosphere.com/browse/DCOS_OSS-611

### DIFF
--- a/resolver/resolver_test.go
+++ b/resolver/resolver_test.go
@@ -351,6 +351,37 @@ func TestHTTP(t *testing.T) {
 	}
 }
 
+// TestHTTPAcceptApplicationJson tests that valid requests that specify
+// 'Accept: application/json' succeed. This used to fail with
+// 406 Not Acceptable.
+// See https://jira.mesosphere.com/browse/DCOS_OSS-611
+func TestHTTPAcceptApplicationJson(t *testing.T) {
+	// setup DNS server (just http)
+	res, err := fakeDNS()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


gpaul:
@nlsun @jdef @alberts  PTAL

resolver/resolver_test.go
```diff
@@ -351,6 +351,38 @@ func TestHTTP(t *testing.T) {
...
+// 'Accept: application/json' succeed. This used to fail with
+// 406 Not Acceptable.
+// See https://jira.mesosphere.com/browse/DCOS_OSS-611
+func TestHTTPAcceptApplicationJson(t *testing.T) {
+	// setup DNS server (just http)
+	res, err := fakeDNS()
+	if err != nil {
+		t.Fatal(err)
+	}
+	res.version = "0.1.1"
```
jdef: why is this line needed?

gpaul: Good question - copy-pasted from `TestHTTP`. I suspect it is to make the tests proof against future API version updates.

gpaul: I had a look. The `res.version` is set in `TestHTTP` because we test the `/v1/version` endpoint in that test. In this new test it is unnecessary and I have removed it. I've squashed the commits and will merge this once the the checks all pass. Thanks for picking this up @jdef

PR #499, opened 2017-04-16
WIP: numbered ports

awdrius:
Hi all, 

I've decided to try DNS based service discovery recently and I had the same issues as noted in issue #61 (first part of the discussion): basically we have a lot of services that use service wide clustering and requires a dedicated tcp/udp port for it, meaning that deployed services need at last two ports assigned (e.g. first port for REST, second port for clustering). This makes current `_task._protocol.framework.domain` SRV record implementation cumbersome to use. Singularity is mesos scheduler we use and it does not have named ports support (yet?).

I've added a new configuration parameter `NumberPorts`, a quick implementation for it and documentation as well as tests. Once NumberPorts is set to true - SRV records will be created for each port per task, e.g. `_port0._task._protocol.framework.domain`  will return a list of fqdns and ports for `_task` for the very first port assigned to each task instance. `_port1` is for the second port and so on.

It's deployed on a cluster I'm trying DNS-SRV discovery and so far it looks OK but more testing/validation might be required.

P.S. the implementation considers that task information has port information in assigned port order so the actual ports (low-high) are left as is. I have only done test runs on mesos 1.0.1 and so far I did not get any issues with ports being out of order.

### DIFF
--- a/docs/docs/configuration-parameters.md
+++ b/docs/docs/configuration-parameters.md
@@ -4,7 +4,7 @@ title: Mesos-DNS Configuration Parameters
 
 ##  Mesos-DNS Configuration Parameters
 
-Mesos-DNS is configured through the parameters in a json file. You can point Mesos-DNS to a specific configuration file using the argument `-config=pathto/file.json`. If no configuration file is passed as an argument, Mesos-DNS will look for file `config.json` in the current directory. 
+Mesos-DNS is configured through the parameters in a json file. You can point Mesos-DNS to a specific configuration file using the argument `-config=pathto/file.json`. If no configuration file is passed as an argument, Mesos-DNS will look for file `config.json` in the current directory.
 
 The configuration file should include the following fields:
 
@@ -25,7 +25,7 @@ The configuration file should include the following fields:
   "zoneResolvers": {
     "weave": ["172.17.0.1"]
   },
-  "timeout": 5, 
+  "timeout": 5,
   "httpon": true,
   "dnson": true,
   "httpport": 8123,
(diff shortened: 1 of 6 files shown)


docs/docs/naming.md
```diff
@@ -4,7 +4,7 @@ title: Service Naming
 
 # Service Naming
 
-Mesos-DNS defines a DNS domain for Mesos tasks (default `.mesos`, see [instructions on configuration](configuration-parameters.html)). Running tasks can be discovered by looking up A and, optionally, SRV records within the Mesos domain. 
+Mesos-DNS defines a DNS domain for Mesos tasks (default `.mesos`, see [instructions on configuration](configuration-parameters.html)). Running tasks can be discovered by looking up A and, optionally, SRV records within the Mesos domain.
```
jdef: There are lots of whitespace changes here which seem unrelated to the goal of "numbered ports". Docs improvements are welcome, but such changes here that are unrelated to the overall goal end up making this harder to review.

docs/docs/configuration-parameters.md
```diff
@@ -102,13 +103,17 @@ The configured name will always be converted to a FQDN by ensuring it ends with
...
 
-`recurseon` controls if the DNS replies for names in the Mesos domain will indicate that recursion is available. The default value is `true`. 
+`recurseon` controls if the DNS replies for names in the Mesos domain will indicate that recursion is available. The default value is `true`.
 
 `enforceRFC952` will enforce an older, more strict set of rules for DNS labels. For details, see the [RFC-952](https://tools.ietf.org/html/rfc952). The default value is `false`.
 
 `IPSources` defines a fallback list of IP sources for task records,
 sorted by priority. If you use **Docker**, and enable the `netinfo` IPSource, it may cause tasks to become unreachable, because after Mesos 0.25, the Docker executor publishes the container's internal IP in NetworkInfo. The default value is: `["netinfo", "mesos", "host"]`
 
+`NumberPorts` enables generating service SRV records per service port.
```
jdef: better name please. `IndexedPortSRVs`? or something else?

jdef:
ok to test

nlsun:
I agree with @jdef here that getting named port support in your Mesos scheduler is the way to go. The last time I checked, the ordering of ports in Mesos is not guaranteed.

If you _really_ need a quick fix, then it sounds like you could just advertise a single port since it sounds like the 2 ports you use are tcp and udp, which may share the same port. But to reiterate, named ports are the way to go.

awdrius:
Hi guys,

apologies for the late reply.

I have raised named ports issue with Singularity team but so far there is no answer.

I really don'y mind if this patch gets ignored, just wanted to let others know that there is an implementation of it in case they are struggling with named ports and/or have other reasons that require numbered ports as a last resort/quick hack and are willing to risk it.

docs/docs/naming.md
awdrius: I saw that before commit but decided against changing it back as it looked weird to have a whitespace before new line. It should not have been my decision to make but I just went with it at the time of commit.

docs/docs/configuration-parameters.md
awdrius: Haven't thought about it at the time, good idea. Will change it in my repo in case other people are going to start using it.

PR #502, opened 2017-07-06
records: use SOA mname instead of rname

### LINKED ISSUE #501: Ns namaserver's A record is not returned
dig ns1.mesos.test.net

; <<>> DiG 9.9.4-RedHat-9.9.4-29.el7 <<>> ns1.mesos.test.net
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 6850
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;ns1.mesos.test.net.        IN      A                    <---- A record is not returned!!!

;; AUTHORITY SECTION:
ns1.mesos.test.net. 5 IN    SOA     ns1.mesos.test.net. root.ns1.mesos.test.net. 1498640947 60 600 86400 5

config.json:
  "refreshSeconds": 5,
  "ttl": 5,
  "domain": "mesos.test.net",
  "port": 53,
  "resolvers": ["1.1.1.1","2.2.2.2"],
  "timeout": 5, 
  "httpon": true,
  "dnson": true,
  "httpport": 8123,
  "externalon": true,
  "listener": "10.172.190.165",
  "SOAMname": "ns1.mesos.test.net",
  "SOARname": "root.ns1.mesos.test.net",
  "SOARefresh": 60,
  "SOARetry":   600,
  "SOAExpire":  86400,
  "SOAMinttl": 60,
  "IPSources": ["host", "mesos", "netinfo"]

jdef:
fixes #501


jdef:
@gpaul I did a small bit of refactoring in a separate commit from the unit test I added. The refactoring helped to keep the test small, and more cleanly separates state loading from record generation (as it should be). When this PR lands, I'd like to keep the commit history vs. squashing it.

jdef:
There's another, related issue that we *could* solve as part of this ticket: if the SOA `mname` isn't part of the mesos-dns domain, then mesos-dns SHALL NOT report an `A` RR. We don't perform any checks for this (https://github.com/mesosphere/mesos-dns/blob/9d21ea6de38bd371e34bde196ceaee94cdac8a2c/records/generator.go#L478).

See the examples here: http://www.zytrax.com/books/dns/ch8/soa.html

We should either:
(a) validate that the SOA `mname` is within the mesos-dns domain, or else;
(b) NOT generate `A` RR's for an `mname` that is outside the scope of the mesos-dns domain

Should I write up a separate ticket for this, or just fix as part of this one? @gpaul

jdef:
answered my own question re: the above: opened issue #504 to track additional SOA mname rules

PR #508, opened 2017-09-30
Support IPv6

drewkerrigan:
# Get DC/OS with IPv6

Start a new DC/OS cluster with this image:

```
wget https://downloads.dcos.io/dcos/testing/pull/1842/dcos_generate_config.sh
```

# Replace existing Mesos-DNS

On each master, update Mesos-DNS config:

In /opt/mesosphere/etc/mesos-dns.json:

Change `"IPSources": ["host", "netinfo"],` -> `"IPSources": ["netinfo", "host"],`

Then get the new mesos-dns with the following commands:

```
cd /opt/mesosphere/packages/mesos-dns--600da87080b7634f2380594499004a7ff0b34662/bin
mv mesos-dns mesos-dns.bak
curl -O https://959-26509179-gh.circle-artifacts.com/0/tmp/circle-artifacts.Dv9WDHl/mesos-dns--linux-amd64 --output mesos-dns--linux-amd64
mv mesos-dns--linux-amd64 mesos-dns
chmod 755 mesos-dns
systemctl restart dcos-mesos-dns.service
```

# Create a test app on `dcos6` network

```
{
    "id": "/ipv6-nginx",
    "cmd": "nginx -g 'daemon off;'",
    "instances": 1,
    "cpus": 0.1,
    "mem": 32,
    "container": {
        "type": "DOCKER",
        "docker": {
            "image": "nginx:stable"
        },
        "portMappings": [
            {
                "containerPort": 80,
                "protocol": "tcp",
                "name": "web"
            }
        ]
    },
    "requirePorts": false,
    "networks": [
        {
            "name": "dcos6",
            "mode": "container"
        }
    ],
    "healthChecks": [],
    "fetch": [],
    "constraints": []
}
```

# Test the changes

```
dig ipv6-nginx.marathon.mesos ANY

dig ipv6-nginx.marathon.mesos AAAA

dig _ipv6-nginx._tcp.marathon.mesos SRV

curl leader.mesos:8123/v1/hosts/ipv6-nginx.marathon.mesos

curl leader.mesos:8123/v1/services/_ipv6-nginx._tcp.marathon.mesos

curl leader.mesos:8123/v1/enumerate
```


records/generator.go
```diff
@@ -543,6 +553,66 @@ func (rg *RecordGenerator) insertRR(name, host string, kind rrsKind) (added bool
 	return
 }
 
+func rrsKindForIP(ip net.IP) rrsKind {
+	if ip.To4() != nil {
+		return A
+	} else if ip.To16() != nil {
+		return AAAA
+	}
+	return A
```
jdef: should this panic instead? it's unlikely that we'd ever end up here but I hate silent failures

records/generator.go
```diff
@@ -543,6 +553,66 @@ func (rg *RecordGenerator) insertRR(name, host string, kind rrsKind) (added bool
...
+}
+
+func rrsKindForIPStr(ip string) rrsKind {
+	if parsedIP := net.ParseIP(ip); parsedIP != nil {
+		return rrsKindForIP(parsedIP)
+	}
+	return A
```
jdef: panic? otherwise, what are the cases where returning A makes sense here?

records/generator.go
```diff
@@ -543,6 +553,66 @@ func (rg *RecordGenerator) insertRR(name, host string, kind rrsKind) (added bool
...
+}
+
+// hostToIPs attempts to parse hostname into an ip. If that doesn't work it will perform a
+// lookup and try to find one ipv4 and one ipv6 in the results.
+func hostToIPs(hostname string) (ipv4, ipv6 net.IP, ok bool) {
+	ok = false
```
jdef: `ok = false` in superfluous, please remove

records/generator.go
```diff
@@ -543,6 +553,66 @@ func (rg *RecordGenerator) insertRR(name, host string, kind rrsKind) (added bool
...
+			ok = true
+		} else if ipv6 = parsedIP.To16(); ipv6 != nil {
+			ok = true
+		}
+	} else {
+		if t4, err := net.ResolveIPAddr("ip4", hostname); err == nil {
+			ipv4 = t4.IP
+			ok = true
+		} else {
+			logging.Error.Printf("cannot translate hostname %q into an ip4 address", hostname)
```
jdef: why do we log this for ipv4 but not for ipv6? should this be verbose?

records/generator.go
```diff
@@ -543,6 +553,66 @@ func (rg *RecordGenerator) insertRR(name, host string, kind rrsKind) (added bool
...
+			if ipv4 == nil {
+				ipv4 = ip
+			}
+		} else if t6 := ip.To16(); t6 != nil {
+			if ipv6 == nil {
+				ipv6 = ip
+			}
+		}
+	}
+	return ipv4, ipv6
```
jdef: just `return` here w/o args is sufficient

records/generator.go
```diff
@@ -543,6 +553,66 @@ func (rg *RecordGenerator) insertRR(name, host string, kind rrsKind) (added bool
...
+			ok = true
+		} else {
+			logging.Error.Printf("cannot translate hostname %q into an ip4 address", hostname)
+		}
+		if t6, err := net.ResolveIPAddr("ip6", hostname); err == nil {
+			ipv6 = t6.IP
+			ok = true
+		}
+	}
+	return ipv4, ipv6, ok
```
jdef: just `return` here w/o args is sufficient

resolver/resolver.go
```diff
@@ -389,6 +415,14 @@ func (res *Resolver) handleSRV(rs *records.RecordGenerator, name string, m, r *d
 			}
 			m.Extra = append(m.Extra, aRR)
 			added[host] = struct{}{}
+		} else if aaaa, ok := rs.AAAAs.First(host); ok {
```
jdef: why only append these if ipv4 isn't found? why can't we return both? (or is that not kosher?)

resolver/resolver.go
```diff
@@ -718,6 +760,8 @@ func (res *Resolver) RestService(req *restful.Request, resp *restful.Response) {
 		var ip string
 		if r, ok := rs.As.First(host); ok {
 			ip = r
+		} else if r, ok := rs.AAAAs.First(host); ok {
```
jdef: why is ipv4 preferred?

records/generator.go
```diff
@@ -249,18 +245,28 @@ func (rg *RecordGenerator) frameworkRecords(sj state.State, domain string, spec
...
-			rg.insertRR(a, address, A)
+		slaveIP := ""
+		if ipv4, ipv6, ok := hostToIPs(slave.PID.Host); ok {
+			if ipv6 != nil {
+				rg.insertRR(a, ipv6.String(), AAAA)
+				slaveIP = ipv6.String()
+			}
+			if ipv4 != nil {
+				rg.insertRR(a, ipv4.String(), A)
+				// Prefer ipv4 for slaveIP
```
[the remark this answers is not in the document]
drewkerrigan: Valid point - I can instead make SlaveIPs a map of lists of strings so that we can later add records for both ipv4 and ipv6 for .slave.mesos recs

records/generator.go
drewkerrigan: I agree, my only concern is that there are a bunch of tests with master "ip addresses" like `6`, or `bob`, etc. I assume the tests are just flawed and I can make them into valid ipv4 addresses to pass the tests, but I'm not sure if there's a valid use case for masters with non-ip hostnames in master pids / addresses. Should we maybe just not create records and error log for those cases? I'm also fine with panic.

records/generator.go
drewkerrigan: See above comment.

resolver/resolver.go
drewkerrigan: You're correct, this was an oversight

resolver/resolver.go
drewkerrigan: Another oversight, should have both.

drewkerrigan:
@jdef PTAL.

re: "ipv4 is preferred in several places to ipv6 - why?"
- This was mainly due to some flawed assumptions I made. After taking another go at it, I realized we can and should use both when available.

re: "also, what's up with the flapping mesos-go deps?" 
- I initially thought that I would need an updated mesos-go (https://github.com/mesos/mesos-go/pull/328 has an updated upid package) but later decided to just copy a subset of upid into `records/state/upid` with the fixed `Parse` function. We should eventually get mesos-go updated though. I also noticed you already had another PR to update mesos-go to the new api/v0 (https://github.com/mesosphere/mesos-dns/pull/496).

re: "in the future it would be great to have a more well-structured commit stream to simplify the review process" - I can restructure commits before merge or sooner if you'd like - no excuses here :-)

records/generator.go
```diff
@@ -443,19 +434,25 @@ func (rg *RecordGenerator) taskContextRecord(ctx context, task state.Task, f sta
...
-	slaveIPKind := rrsKindForIPStr(ctx.slaveIP)
-	rg.insertTaskRR(arec+".slave"+tail, ctx.slaveIP, slaveIPKind, enumTask)
-	rg.insertTaskRR(canonical+".slave"+tail, ctx.slaveIP, slaveIPKind, enumTask)
+	// slaveIPs already only has at most one ipv4 and one ipv6
+	for _, sIPStr := range ctx.slaveIPs {
+		if sIP := net.ParseIP(sIPStr); sIP != nil {
+			rg.insertTaskRR(arec+".slave"+tail, sIP.String(), rrsKindForIP(sIP), enumTask)
+			rg.insertTaskRR(canonical+".slave"+tail, sIP.String(), rrsKindForIP(sIP), enumTask)
+		} else {
+			// ack: slave IP may not be an actual IP if labels.DomainFrag was used.
```
jdef: good catch. this is suspect. if not debugging/fixing this issue here, should probably create an issue to track

records/state/state_test.go
```diff
@@ -166,8 +166,8 @@ func statuses(st ...Status) taskOpt {
 	}
 }
 
-func slaveIP(ip string) taskOpt {
-	return func(t *Task) { t.SlaveIP = ip }
+func slaveIPs(ip string) taskOpt {
```
jdef: suggestion: `func slaveIPs(ip ...string) taskOpt { ... }`

resolver/resolver.go
```diff
@@ -372,23 +396,32 @@ func (res *Resolver) handleSRV(rs *records.RecordGenerator, name string, m, r *d
...
+		if len(rs.As[host]) == 0 && len(rs.AAAAs[host]) == 0 {
 			continue
 		}
-
-		if a, ok := rs.As.First(host); ok {
-			aRR, err := res.formatA(host, a)
-			if err != nil {
-				errs.Add(err)
-				continue
+		if _, aFound := aAdded[host]; !aFound {
```
jdef: is it worth using different priorities for ipv4 records vs ipv6 records? via http://ipv6friday.org/blog/2012/01/ipv6-and-dns/

> An IPv4 client should check the lowest priority first and find out that there’s no available host in that priority and switch to the second priority. An IPv6 client finds a host in the lowest priority. A dual stack client will also find the IPv6 host first and stick to that. DNS SRV records are for you to define how you want to be reached.

records/generator.go
drewkerrigan: Created https://github.com/mesosphere/mesos-dns/issues/509 and I'll reference from the comment.

resolver/resolver.go
drewkerrigan: Only the SRV records themselves have the `Priority` field, so this would only make sense if we were creating different SRV hostname records for ipv4 and ipv6 which we aren't currently. The code you commented on is only referring to the `additional section`. So in the example:

```
;; QUESTION SECTION:
;_ipv6-nginx._tcp.marathon.mesos. IN	SRV

;; ANSWER SECTION:
_ipv6-nginx._tcp.marathon.mesos. 60 IN	SRV	0 0 80 ipv6-nginx-6ijia-s1.marathon.mesos.

;; ADDITIONAL SECTION:
ipv6-nginx-6ijia-s1.marathon.mesos. 60 IN A	12.0.2.2
ipv6-nginx-6ijia-s1.marathon.mesos. 60 IN AAAA	fd01:b::2:8000:2
```

Where there are both IPv4 and IPv6 records for a single host (`ipv6-nginx-6ijia-s1.marathon.mesos.`), it would be up to the client to decide which they prefer. Priority would only apply if we were doing something like `ipv6-nginx-6ijia-s1.marathon.mesos.` for A and `ipv6-nginx-6ijia-s1.preferred.marathon.mesos.` for AAAA, and then we would have 2 separate SRV records for `_ipv6-nginx._tcp.marathon.mesos.`.

drewkerrigan:
@jdef PTAL, I made some minor updates to the docs and addressed the other comments.

resolver/resolver.go
jdef: You're right, thanks for clarifying.

drewkerrigan:
@jdef yes we'll add integration tests with the related DC/OS PR

PR #510, opened 2017-11-18
Updated logging regarding new masters.

ArmandGrillet:
The detection of new masters now results in a log line that is displayed even with a disabled verbose mode. 

The idea behind this change is that this event is important and not common thus it should be logged as such.

### DIFF
--- a/main.go
+++ b/main.go
@@ -84,7 +84,7 @@ func main() {
 			} else {
 				timeout.Stop()
 			}
-			logging.VeryVerbose.Printf("new masters detected: %v", masters)
+			logging.Verbose.Printf("new masters detected: %v", masters)
 			res.SetMasters(masters)
 			res.Reload()
 		case err := <-errch:


main.go
```diff
@@ -84,7 +84,7 @@ func main() {
 			} else {
 				timeout.Stop()
 			}
-			logging.VeryVerbose.Printf("new masters detected: %v", masters)
+			log.Printf("new masters detected: %v", masters)
```
rukletsov: Thanks for the PR! It looks like this project does not use `log.Printf` and instead uses `logging.Verbose`, according to [this](https://github.com/mesosphere/mesos-dns/blob/1810c09811dfeb8c9e72ee6c58102313f74e0e1c/logging/logging.go#L13-L24). This seems strange to me, also it looks like in DC/OS mesos-dns is not running in verbose mode by default. You should probably ask @drewkerrigan

ArmandGrillet: `logging.Verbose` also requires a flag and I think that having this log line displayed in all cases would be better for users. `log.Printf` is used in `tools/main.go` and `util/util.go`, as `log` was already imported in `main.go` it did not seem illogical to use it but I may be wrong.

jdef: `logging.Verbose` would be correct here. the philosophy of the `logging` package is that a log message carries an error signal or else it's some kind of an "info" (verbose) message. modules in mesos-dns should use `logging` to generate log messages.

using the stdlib `log` func in main should probably be rewritten. i haven't looked yet at the `util` package.

jdef:
I'm pretty sure that, at one point, very-verbose logging was enabled for mesos-dns on dc/os. I took a peek at the dcos config scripts and also observe that *any* "verbose" logging appears to be completely disabled. I wonder why? (Too noisy in larger clusters?)

ArmandGrillet:
@jdef I have created https://jira.mesosphere.com/browse/DCOS_OSS-1880 to report the verbose mode issue. This PR is the result of a customer issue where having verbose logging would have been quite useful. 

I do not think that this PR is dependent of DCOS_OSS-1880. If reviewers agree that an event such as new masters deserve more than a very verbose log line, it should be able to be merged.

jdef:
LGTM

Just rebased. Should be merged if CI passes

PR #511, opened 2017-11-20
Added error message when no masters given.

ArmandGrillet:
We now return an error message if no masters have been given when
generating a new Mesos master state loader.

### DIFF
--- a/records/state/client/cli.go
+++ b/records/state/client/cli.go
@@ -43,6 +43,8 @@ func LoadMasterStateTryAll(masters []string, stateLoader func(ip, port string) (
 
 	if len(masters) > 0 {
 		leader, masters = masters[0], masters[1:]
+	} else {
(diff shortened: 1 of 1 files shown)


records/state/client/cli.go
```diff
@@ -43,6 +43,10 @@ func LoadMasterStateTryAll(masters []string, stateLoader func(ip, port string) (
 
 	if len(masters) > 0 {
 		leader, masters = masters[0], masters[1:]
+	} else {
+		err = errors.New("No masters given to fetch state.json")
+		logging.Error.Println(err)
```
rukletsov: I don't think you should log here, it should be done by the caller based on the error you have created.

jdef:
I *think* this is OK. Unfortunate that CI is breaking because of linter failures - is the test suite being executed at all? If not, then I'd like to see that fixed before any other changes land.

PR #512, opened 2017-12-18
Set version string to 1.11 in order to reflect DC/OS version.

asridharan:
Created a `dcos/1.11` branch on `mesos-dns`. Have also updated the version string of `mesos-dns` to reflect the DC/OS version. 

Currently `mesos-dns --version` just returns the version string `dev`. This causes a lot of confusion in users. Since `mesos-dns` is explicitly tested against DC/OS and ships with DC/OS as a systemd unit, for every DC/OS branch we plan to set the `Version` string to the DC/OS version the branch is going to ship with.

https://jira.mesosphere.com/browse/DCOS_OSS-1990


jdef:
this is weird. you don't need to run dc/os to leverage mesos-dns on a mesos cluster. I disagree with this approach to versioning mesos-dns

asridharan:
@jdef the ask here is to highlight the version of `mesos-dns` runnins with a specific DC/OS version. My thought here is that we should have a separate `mesos-dns` branch for each DC/OS version, and so this goes into a `mesos-dns` branch which reflects the DC/OS version it goes with.

The other reasoning here is that the main testing that we do for `mesos-dns`, when testing against Meso, is in the DC/OS integration tests so wanted to peg the `mesos-dns` versioning in the given branch against the DC/OS version.

asridharan:
@drewkerrigan can you take a look as well.

asridharan:
@drewkerrigan the problem with something like `stable` is that it doesn't reflect the testing (exactly what was tested) that was done before a release. Since these branches are meant to go with DC/OS hence felt that reflecting the DC/OS version in these branches would make sense, giving a clear indication of the release against which the branch was tests? 

Does that make sense?

PR #515, opened 2018-01-26
add non-zero weight for returned SRV records

rubbish:
If this is zero, services like haproxy see that as a backend that's in maintenance. I couldn't see any problem with making it non-zero.


jdef:
ok to test

jdef:
thanks @rubbish can you provide a link to the relevant haproxy docs?

rubbish:
@jdef I couldn't find it the haproxy docs (https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#5.3.1), this is something I found in my own testing. 

But looking at the haproxy source, http://git.haproxy.org/?p=haproxy-1.8.git;a=blob;f=src/dns.c;h=a957710edb2f92e31069844a15338cc8834e8402;hb=HEAD#l527 it has: 
> Make sure weight is at least 1, so that the server will be used.

jdef:
OK, thanks. My gut says that the default weight should be configurable.
Maybe we make that default weight 1 so that SRVs play nice with haproxy by
default. The reason we make it configurable is so that people can roll back
to using 0 if the new default of 1 causes problems in their cluster.

How does that sound?

rubbish:
@jdef Sounds good. Added in 2f04f61

docs/docs/configuration-parameters.md
```diff
@@ -87,7 +87,9 @@ It is sufficient to specify just one of the `zk` or `masters` field. If both are
 
 `httpport` is the port number that Mesos-DNS monitors for incoming HTTP requests. The default value is `8123`.
 
-`externalon` is a boolean field that controls whether Mesos-DNS serves requests outside of the Mesos domain. The default value is `true`. 
+`externalon` is a boolean field that controls whether Mesos-DNS serves requests outside of the Mesos domain. The default value is `true`.
+
+`srvRecordDefaultWeight` is an int field field that defines what the default weight given to SRV records. The default value is `1`.
```
jdef: nit: s/what//

records/config.go
```diff
@@ -111,32 +113,33 @@ type Config struct {
...
-		HTTPPort:            8123,
-		DNSOn:               true,
-		HTTPOn:              true,
-		ExternalOn:          true,
-		SetTruncateBit:      true,
-		RecurseOn:           true,
-		IPSources:           []string{"netinfo", "mesos", "host"},
-		EnumerationOn:       true,
-		MesosAuthentication: httpcli.AuthNone,
+		ZkDetectionTimeout: 30,
```
jdef: nit: run this through go fmt please

rubbish:
should be gtg now

jdef:
retest this please

PR #517, opened 2018-02-26
Ignore inactive frameworks

drewkerrigan:
Current behavior of the record generator is to create A records for all frameworks even if `"active": false`. One specific scenario that this can be problematic is this:

1. Install marathon on marathon (MoM) DC/OS package with serviceID: `marathon`
2. `nslookup marathon.mesos` and note that there are now 2 A records for `marathon.mesos`
3. Uninstall MoM
4. `nslookup marathon.mesos` and note that there are still 2 A records for `marathon.mesos`, one of them invalid

This PR addresses this by ignoring inactive frameworks.

### DIFF
--- a/factories/fake.json
+++ b/factories/fake.json
@@ -213,6 +213,35 @@
              "user": "root"
         },
         {
+            "id":"1b4c4343-a8a1-48ca-a74c-4e24d7947303-0002",
+            "name":"marathon",
+            "pid":"scheduler-621043d2-ab95-448d-a5a9-970d475a8b4e@10.0.2.175:44369",
+            "capabilities":["TASK_KILLING_STATE","PARTITION_AWARE","REGION_AWARE"],
+            "hostname":"10.0.2.175",
+            "webui_url":"http:\/\/10.0.2.175:26074",
+            "active":false,
+            "connected":false,
+            "recovered":false,
+            "user":"root",
+            "failover_timeout":604800.0,
+            "checkpoint":true,
+            "registered_time":1519673766.85769,
+            "unregistered_time":0.0,
+            "principal":"marathon",
+            "resources":{
+              "disk":0.0,
+              "mem":0.0,
+              "gpus":0.0,
+              "cpus":0.0
+            },
+            "role":"marathon",
+            "tasks":[],
+            "unreachable_tasks":[],
+            "completed_tasks":[],
+            "offers":[],
+            "executors":[]
+        },
+        {
             "active": true,
             "checkpoint": true,
             "completed_tasks": [
--- a/records/generator.go
+++ b/records/generator.go
@@ -221,6 +221,9 @@ func (rg *RecordGenerator) InsertState(sj state.State, domain, ns, listener stri
 //     _framework._tcp.frameworkname.domain. // resolves to the driver port and IP of each framework
 func (rg *RecordGenerator) frameworkRecords(sj state.State, domain string, spec labels.Func) {
 	for _, f := range sj.Frameworks {
+		if !f.Active {
+			continue
+		}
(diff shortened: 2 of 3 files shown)


jdef:
I think this is a bit more nuanced than the solution proposed here. Consider a similar scenario, root Marathon + MoM. The root Marathon fails over, during which time mesos-dns takes a snapshot of Mesos state: the result of such a race might be that mesos-dns only reflects an entry for MoM, right? I don't think this is very desirable. I'm happy to be wrong. How does Spartan deal with something like this?

jdef:
"active" means something very specific to Mesos: https://github.com/apache/mesos/blob/49c642e98a7dac911a7d21aea1c429e979def0ab/src/master/master.hpp#L2167

urbanserj:
@jdef Spartan doesn't have records for frameworks. A proper fix would be to add a validation that framework names are unique, but it's not possible.

jdef:
if MoM is uninstalled properly (via Mesos teardown) then how is it even showing up in the `frameworks` list of the state.json?

drewkerrigan:
@jdef That is actually another problem altogether related to the DC/OS UI I believe. I noticed an error / notice in the UI that said something to the effect of `Could not complete teardown of framework marathon because multiple frameworkIDs exist for that name`

jdef:
it's not an orthogonal problem. the scenario you're solving for in the description here shouldn't happen if the MoM is being uninstalled properly.

drewkerrigan:
Valid point, I still wonder if we should be creating records for inactive frameworks though, but considering the default refresh period is 60 seconds, probably best to leave the records there.

jdef:
Well, like I said earlier, "active" means something very specific to Mesos. It does not distinguish between DC/OS "installed" vs. "uninstalled". What other use case(s) are you trying to solve for? If none, then I think we should cancel this PR

justinrlee:
I don't think there's a requirement at all that framework names be unique in Mesos.  FrameworkIDs, yes, but not names.

I think this calls for a larger discussion w.r.t DC/OS - if a framework is inactive but has active tasks, should we be routing to tasks on it (and/or providing DNS records for those tasks)?  How about inactive with no active tasks?  In my opinion, "yes" to the former, because the underlying task health indicates that stuff is "still going on", but "no" to the latter, because it indicates that a framework was not properly deregistered.

jdef:
Its a big problem for mesos-dns that it doesn't watch the master for
events, and instead it relies on periodic snapshots. But I don't think
there's any alternative given that events emitted by master don't actually
contain **all** the task metadata needed to generate appropriate records.
So we're stuck w/ state snapshots for now. Trying to make decisions based
on state that's possibly quite stale is error prone and we'll never satisfy
everyone's use cases. We (mesos-dns) need to pick a reasonable middle
ground and try to not over-fit a particular problematic scenario. Luckily
DC/OS ships w/ an additional service lookup/network load balancing solution
that's much more responsive to cluster dynamics than is mesos-dns.

Framework names ARE NOT guaranteed to be unique and this has actually
proved to be a problem in practice for many clusters. I'm pretty sure that
Mesos is going to ship w/ a flag, at some point, that allows someone to
enforce a name uniqueness constraint - and that we'll end up incorporating
that into DC/OS. The possibility of duplicate names presents several
problems for DC/OS clusters.
https://issues.apache.org/jira/browse/MESOS-1719?focusedCommentId=16126633&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16126633

Also, your latter case doesn't guarantee that the framework was not
properly deregistered - a framework could legitimately have zero tasks and
being in the process of failing over.

PR #519, opened 2018-05-08
use container ports in tasks SRV records

vixns:


records/state/state.go
```diff
@@ -200,6 +208,23 @@ func statusIPs(st []Status, src func(*Status) []string) []string {
 	return nil
 }
 
+// MapPort returns the mapped port if available, or the host port
+// listening on.
+// TODO: error checking, is Statuses[0] always the current status ?
+func MapPort(t Task, hostport int) (int, string) {
+	ni := t.Statuses[0].ContainerStatus.NetworkInfos
+	for n := range ni {
+		if ni[n].PortMappings != nil {
```
jdef: why is this `if` check needed?

vixns: it isn't, I will remove it

vixns:
`UseContainerPorts` is not a good name, `SRVTaskRecordsUseContainerPortsWithNetinfo` has a better semantic, but may be too long, any idea for a better option name ?

jdef:
SRVPreferContainerPorts ?

records/generator.go
```diff
@@ -202,7 +202,7 @@ func hashString(s string) string {
 }
 
 // InsertState transforms a StateJSON into RecordGenerator RRs
-func (rg *RecordGenerator) InsertState(sj state.State, domain, ns, listener string, masters, ipSources []string, spec labels.Func) error {
+func (rg *RecordGenerator) InsertState(sj state.State, domain, ns, listener string, masters, ipSources []string, SRVPreferContainerPorts bool, spec labels.Func) error {
```
jdef: s/SRV/srv/

records/generator.go
```diff
@@ -368,7 +368,7 @@ func (rg *RecordGenerator) listenerRecord(listener string, ns string) {
 	}
 }
 
-func (rg *RecordGenerator) taskRecords(sj state.State, domain string, spec labels.Func, ipSources []string) {
+func (rg *RecordGenerator) taskRecords(sj state.State, domain string, spec labels.Func, ipSources []string, SRVPreferContainerPorts bool) {
```
jdef: s/SRV/srv/

records/generator.go
```diff
@@ -382,21 +382,22 @@ func (rg *RecordGenerator) taskRecords(sj state.State, domain string, spec label
...
-	taskID   string
-	slaveID  string
-	taskIPs  []net.IP
-	slaveIPs []string
+	taskName                string
+	taskID                  string
+	slaveID                 string
+	taskIPs                 []net.IP
+	slaveIPs                []string
+	SRVPreferContainerPorts bool
```
jdef: s/SRV/srv/

records/generator.go
```diff
@@ -382,21 +382,22 @@ func (rg *RecordGenerator) taskRecords(sj state.State, domain string, spec label
...
 }
 
-func (rg *RecordGenerator) taskRecord(task state.Task, f state.Framework, domain string, spec labels.Func, ipSources []string, enumFW *EnumerableFramework) {
+func (rg *RecordGenerator) taskRecord(task state.Task, f state.Framework, domain string, spec labels.Func, ipSources []string, SRVPreferContainerPorts bool, enumFW *EnumerableFramework) {
```
jdef: s/SRV/srv/

records/generator.go
```diff
@@ -488,7 +490,11 @@ func (rg *RecordGenerator) taskContextRecord(ctx context, task state.Task, f sta
 	}
 
 	for _, port := range task.DiscoveryInfo.Ports.DiscoveryPorts {
-		target := canonical + tail + ":" + strconv.Itoa(port.Number)
+		p := port.Number
+		if ctx.SRVPreferContainerPorts {
+			p = state.MapPort(task, port.Number)
```
jdef: would prefer: `p = state.MapPort(task, p)`

records/generator_perf_test.go
```diff
@@ -46,13 +46,14 @@ func BenchmarkTaskRecord_withoutDiscoveryInfo(b *testing.B) {
...
-		enumFW    EnumerableFramework
-		rg        RecordGenerator
+		task                    state.Task
+		f                       state.Framework
+		domain                  string
+		spec                    labels.Func
+		ipSources               []string
+		enumFW                  EnumerableFramework
+		rg                      RecordGenerator
+		SRVPreferContainerPorts bool
```
jdef: s/SRV/srv/

records/state/state.go
```diff
@@ -200,6 +208,20 @@ func statusIPs(st []Status, src func(*Status) []string) []string {
...
 }
 
+// MapPort returns the mapped port if available, or the host port
+// listening on.
+func MapPort(t Task, hostport int) int {
+	ni := t.Statuses[0].ContainerStatus.NetworkInfos
+	for n := range ni {
+		for m := range ni[n].PortMappings {
+			if ni[n].PortMappings[m].HostPort == hostport {
+				return ni[n].PortMappings[m].ContainerPort
```
jdef: possible edge case: if container port == 0 do we really want to use it? or should we fall back to host port? it's probably a framework bug to indicate a container port of zero. maybe we should log such cases and return the host port instead?

vixns: For container port == 0, using the hostport seems fine.

vixns:
oops rebased on wrong branch, sorry. Re-rebased and force pushed.

PR #523, opened 2018-07-19
Feature/psod 584

aramhakobyan:
adding rpmbuild spec files


jdef:
What's the motivation for this? Why was the package namespace changed? Looks like a PR pushed to the wrong GH repo...

aramhakobyan:
accidentally created PR on parent repo

PR #525, opened 2018-08-20
Moves mesos state endpoint from /state.json to /state

GoelDeepak:
Mesos has deprecated/removed the `/state.json` endpoint and has
replaced it with `/state`. This patch updates the url path from
`/state.json` to `/state`

jira: DCOS_OSS-3941

### DIFF
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -783,7 +783,9 @@ func panicRecover(f func(w dns.ResponseWriter, r *dns.Msg)) func(w dns.ResponseW
 			if rec := recover(); rec != nil {
 				m := new(dns.Msg)
 				m.SetRcode(r, 2)
-				_ = w.WriteMsg(m)
+				if err := w.WriteMsg(m); err != nil {
+					logging.Error.Println(err)
+				}
 				logging.Error.Println(rec)
 			}
 		}()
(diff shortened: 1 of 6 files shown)


resolver/resolver.go
```diff
@@ -783,7 +783,9 @@ func panicRecover(f func(w dns.ResponseWriter, r *dns.Msg)) func(w dns.ResponseW
 			if rec := recover(); rec != nil {
 				m := new(dns.Msg)
 				m.SetRcode(r, 2)
-				_ = w.WriteMsg(m)
+				if err := w.WriteMsg(m); err != nil {
```
jdef: nit: formatting is hosed here. please fix

vespian:
I have gone through the CircleCi failures:
```
httpcli/iam/file.go:15::warning: Potential file inclusion via variable,MEDIUM,HIGH (gosec)
records/config.go:316::warning: Potential file inclusion via variable,MEDIUM,HIGH (gosec)
records/generator.go:6::warning: Blacklisted import crypto/sha1: weak cryptographic primitive,MEDIUM,HIGH (gosec)
records/generator.go:200::warning: Use of weak cryptographic primitive,MEDIUM,HIGH (gosec)
```
I believe that the problem is that linters are a moving target and that the newest version of gosec points out the things that were OK in previous versions.

First two items seem bogus - files are included via variable but for a legitimate reason. Remaining two - `sha1` is used only for calculating unique taskID in the context. Only first five chars are used. I am pretty confident that using sha1 here is not a security concern.

I am going to whitelist all 4 cases and push the commit on top of Deepak's.

PR #537, opened 2019-04-10
[DO-NOT-MERGE] hack to add etcd SRV record for some experimentation

GoelDeepak:


jdef:
The (older) etcd-mesos framework used a different "protocol" to get nice SRV records (avoids collision) for discovery. https://github.com/mesosphere/etcd-mesos/blob/ed463c2c4343f5714e180f9df5a1ba6d7302c164/scheduler/scheduler.go#L969

Issue #541, opened 2019-08-02
only showing the 2nd network


f1-outsourcing:
Have launched a task like this

```
"networks": [ 
    { "mode": "container", "name": "cni-apps" },
    { "mode": "container", "name": "cni-apps-public", "labels": {"CNI_ARGS": "IP=x.x.x.x"} }
    ],

```

dig +short only shows me the ip address of the last network.

using v0.7.0-rc2

f1-outsourcing:
When is this finally being fixed?

f1-outsourcing:
any one?

PR #553, opened 2020-07-29
Remove limit of one IP address per A record, modify Task.IPs behavior…

drewkerrigan:
… to return only IPs from first populated source

Addresses https://github.com/mesosphere/mesos-dns/issues/541

* Added second network_infos to a task in `factories/fake.json`
* Added corresponding test in `records/generator_test.go`
* Removed `ipsTo4And6` in `records/generator.go`, it was unnecessarily limiting IP lists to a single ipv4 and ipv6 address for no good reason other than to perhaps avoid creating A records for more than one source
* Modified behavior of `Task.IPs` in `records/state/state.go` such that it returns all ip addresses from the first populated configured source
    * The previous behavior was to return all available IP addresses from all configured sources. These additional IPs were never used, because the list was always limited to a single ipv4 and ipv6 address.
* Updated a test in `records/state/state_test.go` to pass with the new `Task.IPs` behavior

### DIFF
--- a/factories/fake.json
+++ b/factories/fake.json
@@ -83,6 +83,15 @@
                                         "ip_addresses": [
                                             {
                                                 "protocol": "IPv4",
+                                                "ip_address": "12.0.1.3"
+                                            }
+                                        ],
+                                        "name": "dcos6"
+                                    },
+                                    {
+                                        "ip_addresses": [
+                                            {
+                                                "protocol": "IPv4",
                                                 "ip_address": "12.0.1.2"
                                             },
                                             {
--- a/records/generator.go
+++ b/records/generator.go
@@ -567,7 +567,7 @@ func rrsKindForIPStr(ip string) rrsKind {
 
 // hostToIPs attempts to parse a hostname into an ip.
 // If that doesn't work it will perform a lookup and try to
-// find all ipv4 and ipv6 addresses for teh hostname.
+// find all ipv4 and ipv6 addresses for the hostname.
 func hostToIPs(hostname string) (ips []net.IP) {
 	if ip := net.ParseIP(hostname); ip != nil {
 		ips = []net.IP{ip}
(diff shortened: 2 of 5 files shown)


drewkerrigan:
To test the changes, I've uploaded binaries here: https://github.com/drewkerrigan/mesos-dns/releases/tag/v0.7.1

f1-outsourcing:
Drew I will ask support to have a look at this.

factories/fake.json
```diff
@@ -79,6 +79,15 @@
...
                                 },
                                 "network_infos": [
+                                    {
+                                        "ip_addresses": [
+                                            {
+                                                "protocol": "IPv4",
+                                                "ip_address": "12.0.1.3"
+                                            }
+                                        ],
+                                        "name": "dcos6"
```
jkoelker, line 89: Technically not a problem, but lulz. ;)

jkoelker:
Howdy! Thanks for taking care of this. It LGTM, but as I'm not 100% on the ramifications on other parts of DC/OS I asked internally for peeps to let me know by 2300 UTC on Wednesday if they see an issue with multiple addresses. From a DNS/Networking POV, it totes makes sense, but just want to make sure there aren't issues that we'll need to shake out elsewhere before incorporating it. If I dont' hear anything back I'll merge it on Wednesday. Happy Hacking!

f1-outsourcing:
> 
> 
> Howdy! Thanks for taking care of this. It LGTM, but as I'm not 100% on the ramifications on other parts of DC/OS I asked internally for peeps to let me know by 2300 UTC on Wednesday if they see an issue with multiple addresses. From a DNS/Networking POV, it totes makes sense, but just want to make sure there aren't issues that we'll need to shake out elsewhere before incorporating it. If I dont' hear anything back I'll merge it on Wednesday. Happy Hacking!

This is an issue[1] I have noticed, but since (I guess) most dcos users do not seem to use multiple ip's, this should not be much of an issue.

[1]
https://github.com/mesosphere/mesos-dns/issues/554

factories/fake.json
drewkerrigan: The network name? That's actually the real default ipv6 network name in dcos :-) https://github.com/dcos/dcos/blob/270cb17ec173dc52e69892c1c3ba796414f46936/gen/calc.py#L1140

jkoelker: Right, but that's not an ipv6 address ;)

drewkerrigan: Whoops, missed that!

jkoelker (approved):

